PR #3, opened 2021-05-31
Change operator structure

lujiajing1126:
## Motivation

As we've discussed this morning, in order to eliminate ambiguity of relationship between key and values in `PairQuery`, I would like to change the structure of `PairQuery` to only allow a single operator in `PairQuery`.

## Design

Rather than put multiple operators and values in one `PairQuery`, we may use the following methodology to compose conditional query,

- Range operation can be implemented by submitting two conditions, i.e. `[{key: duration, op: GT, values: [minValue]}, {key: duration, op: LE, values: [maxValue]}]`
- Having: a single condition `{key: peer.services, op: HAVING, values: [1,2,3]}`
- Composite indexes: For example, in order to make composite index “colA-colB-colC-colD” work, all four columns should be required, i.e. colA, colB and colC must not be skipped and only the condition for the rightmost column may not be EQUAL. `[{key: colA, op: EQ, values: [valA]}, {key: colB, op: EQ, values: [valB]}], {key: colC, op: EQ, values: [valC], {key: colD, op: EQ, values: [valD]}}`

Following the idea, the Query module should be able to 
1. merge operations for the same key, 
2. decide whether any composite index can be made use of and then choose the most suitable and restrictive one to improve performance.

In the future, we may also support HINT to force the query module to use some indexes.

## Alternatives

Use `Union of Enums`, but it is not supported by FlatBuffers. See issue https://github.com/google/flatbuffers/issues/5791

### DIFF
--- a/api/fbs/v1/query.fbs
+++ b/api/fbs/v1/query.fbs
@@ -19,8 +19,11 @@ include "database.fbs";
 
 namespace banyandb.v1;
 
-// BinaryOp specify the operation imposed to the given query condition
-enum BinaryOp: byte {
+// BinaryOp specifies the operation imposed to the given query condition
+// For EQ, NE, LT, GT, LE and GE, only one operand should be given, i.e. one-to-one relationship.
+// HAVING and NOT_HAVING allow multi-value to be the operand such as array/vector, i.e. one-to-many relationship.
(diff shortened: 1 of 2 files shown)


wu-sheng (approved):

hanahmily (approved):
LGTM

PR #4, opened 2021-05-31
Add KV module and refactor storage module

hanahmily:
These changes intend to introduce the KV storage and refactor storage module to wrap the KV storage.

The storage module only implements minimal functionality. @ascrutae Would you pls finish the rest and major part of this module?

Cc @lujiajing1126 @Fine0830


wu-sheng:
> Would you pls finish the rest and majority items?

Does he have a timeline? If not, we need to move forward.

wu-sheng (approved):
We could merge this, but I notice our binary license and notice have not been taken care of.

PR #5, opened 2021-06-02
Feat: query module

lujiajing1126:
## Design

> This is a very perlimenary PR for the query module so far. Much things have to be considered further.

Since it would be a principle module, I just want to discuss the current design/implementation ASAP to avoid improper design  and find some better ideas to proceed.

### Logical Plan

A Logical Plan is a DAG (Directed Acyclic Graph) of Params.
The `Param` defines necessary parameters for query execution. The parameters are well prepared during the logical plan composing in order to reduce extra cost while executing physical plan.

### Plot

the logical plan can be plotted as Dot graph. For example, 

```dot
digraph  {

	n2[label="ChunkIDsFetch{metadata={group=skywalking,name=trace},projection=[TraceID startTime]}"];
	n1[label="ChunkIDsMerge{}"];
	n7[label="IndexScan{begin=1623203253604099000,end=1623214053604099000,KeyName=duration,conditions=[<=1000],metadata={group=skywalking,name=trace}}"];
	n4[label="Pagination{Offset=0,Limit=0}"];
	n5[label="Root{}"];
	n3[label="SortMerge{fieldName=startTime,sort=DESC}"];
	n6[label="TraceIDFetch{TraceID=aaaaaaaa,metadata={group=skywalking,name=trace},projection=[TraceID startTime]}"];
	n2->n3;
	n1->n2;
	n7->n1;
	n5->n6;
	n5->n7;
	n3->n4;
	n6->n3;
}
```

We can leverage [the online toolkit](https://dreampuf.github.io/GraphvizOnline/) to visualize the logical plan,

graphviz

### Physical Plan

A Physical Plan contains the logical plan and the `Transform`(s) corresponding to each `logical.Op`.

While the plan is triggered to run, a reversed topology-sorted slices (`Future` as items) will be generated from the logical plan.

## Tasks

- [x] client utils for building `EntityCriteria` 
- [x] logical plan: Ops such as Sort, OffsetAndLimit, ChunkIDMerge, TableScan and IndexScan
- [x] physical plan: topology sort, `Transform`
- [ ] complete API to connect with `Liaison` (Add handlers) (Maybe next PR)

### Index selection and optimization stage

For now, I only use single-value indexes. But in the current implementation, we may be able to improve index selection during the process of generating Logical Plan.

Any better idea? Since for the traditional databases, normally they have optimization stages (usually after generating hierarchical logical plan?) for indexes selection. How can we fit this optimization stage in our implementations?

### Sort and field orderliness

I believe we have to impose stronger preconditions to sort-field since it is not possible to sort on a sparse field.

And Sort requires the Field to be arranged in a strict order, i.e. we have to use `fieldIndex` number to access the field that is needed to be sorted quickly. Otherwise, it may cost much resources to find the specific field every time.


hanahmily:
> Index selection and optimization stage
> For now, I only use single-value indexes. But in the current implementation, we may be able to improve index selection during the process of generating Logical Plan.
> 
> Any better idea? Since for the traditional databases, normally they have optimization stages (usually after generating hierarchical logical plan?) for indexes selection. How can we fit this optimization stage in our implementations?

From my perspective, the logical plan can't determine which indices should be used, which causes we can't optimize the logical plan before transferring it to physical plans. The reason is we don't support the composited index. A field doesn't exist in more than one index rule, which causes the logical plan doesn't have to select a better rule in a collection. In another word, field criteria separate from each other. I didn't find any paths to do optimization in current step. 

> Sort and field orderliness
> I believe we have to impose stronger preconditions to sort-field since it is not possible to sort on a sparse field.
> 
> And Sort requires the Field to be arranged in a strict order, i.e. we have to use fieldIndex number to access the field that is needed to be sorted quickly. Otherwise, it may cost much resources to find the specific field every time.

Make sense, let's twist index rule to indicate which field can be sorted, and how to sort(asc/desc).

lujiajing1126:
> From my perspective, the logical plan can't determine which indices should be used, which causes we can't optimize the logical plan before transferring it to physical plans. The reason is we don't support the composited index. A field doesn't exist in more than one index rule, which causes the logical plan doesn't have to select a better rule in a collection. In another word, field criteria separate from each other. I didn't find any paths to do optimization in current step.

So as I understood, in our scenario, the indexed to be used is determinant which does not need further optimization, at least for now, right?

A DAG is hard to be optimized. If we need, we might be better to have a hierarchical structure as a logical plann.

hanahmily:
> the indexed to be used is determinant which does not need further optimization
No. Once composite indices are introduced, we should choose which index or indices should be used, and ignore some particular index. 

> A DAG is hard to be optimized. 
Isn't the DAG composited by request/query? The DAGs should be varied depending on the criteria.

lujiajing1126:
> > the indexed to be used is determinant which does not need further optimization
> 
> No. Once composite indices are introduced, we should choose which index or indices should be used, and ignore some particular index.

I see.

> > A DAG is hard to be optimized.
> 
> Isn't the DAG composited by request/query? The DAGs should be varied depending on the criteria.

No. I mean as long as a DAG is generated, it is hard to optimize it since we can only know its parents and children during the walk.

So I suppose, in our context, we can generate/choose an already optimized "logical" plan according to the criteria instead of optimizing an existing set of logical plans like what a general databse does.

hanahmily:
> So I suppose, in our context, we can generate/choose an already optimized "logical" plan according to the criteria instead of optimizing an existing set of logical plans like what a general databse does.

got your point. banyandb will implement rule-based and cost-based optimization. The first is the "logical" plan as you mentioned, and the latter is the physical one. Based on that, if the "DAG" path is too rigid to adapt to cost-based optimization, we might choose another solution for the architecture of the execution plan.

lujiajing1126:
> > So I suppose, in our context, we can generate/choose an already optimized "logical" plan according to the criteria instead of optimizing an existing set of logical plans like what a general databse does.
> 
> got your point. banyandb will implement rule-based and cost-based optimization. The first is the "logical" plan as you mentioned, and the latter is the physical one. Based on that, if the "DAG" path is too rigid to adapt to cost-based optimization, we might choose another solution for the architecture of the execution plan.

Yes. After some investigation, I found tree-structured logical plan may be more suitable for our cases,

- use rule-based optimization for logical plan
- generate multiple physical plans and choose the best one based on cost-based optimizer

while DAG is still feasible for a physical plan. I will continue working on this PR.

PR #6, opened 2021-06-02
Implement the writing protocol for trace

Fine0830:
Implement the writing protocol for trace. Refer to https://github.com/apache/skywalking/issues/6922


api/fbs/v1/rpc.fbs
```diff
@@ -16,10 +16,12 @@
 // under the License.
 
 include "query.fbs";
+include "write.fbs";
 
 namespace banyandb.v1;
 
 // Query is the RPC service exposed
 rpc_service Query {
```
hanahmily, line 24: Since you put write rpc here, the service name should be `Trace`

api/fbs/v1/write.fbs
```diff
@@ -0,0 +1,60 @@
...
+    // entity_id could be span_id of a Span or segment_id of a Segment in the context of Trace
+    entity_id: string;
+    // timestamp_nanoseconds is in the timeunit of nanoseconds. It represents
+    // 1) either the start time of a Span/Segment,
+    // 2) or the timestamp of a log
+    timestamp_nanoseconds: uint64;
+    // binary representation of segments, including tags, spans...
+    data_binary: [ubyte];
+    // support all of indexed fields in the fields.
+    fields: [Field];
```
hanahmily, line 50: Could you elaborate on why you only write values instead of key-value paires? Pls, comment on `fileds` to show more details.

api/fbs/v1/common.fbs
```diff
@@ -0,0 +1,41 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+namespace banyandb.v1;
+
+// Sort is either descending or ascending
```
hanahmily, line 20: Why move them from query to common even write schema doesn't refer to them?

banyand/liaison/trace/trace.go
```diff
@@ -18,5 +18,5 @@
 namespace v1;
 
 table Trace {
-    
-}
\ No newline at end of file
+
```
hanahmily, line 21: Why change a `fbs` file to a `go` one?

hanahmily (changes requested):
Ci failed, pls fix them.

api/fbs/v1/common.fbs
Fine0830: As they are used in `query` and `write`, I move they to common and refer them to `query` and `write`.

banyand/liaison/trace/trace.go
Fine0830: Oh, my bad...

api/fbs/v1/rpc.fbs
```diff
@@ -16,10 +16,12 @@
...
 
-// Query is the RPC service exposed
-rpc_service Query {
-    QueryTraces(EntityCriteria):TracesResponse;
-}
\ No newline at end of file
+// Trace is the RPC service exposed
+rpc_service Trace {
+    QueryTraces(EntityCriteria): TracesResponse;
+    WriteTraces(WriteEntity): WriteResponse(streaming: "bidi");
```
wu-sheng, line 26: @hanahmily Do you need `bidi`? It is typically not useful unless OAP and BanyanDB should communicate during writing data(trace).

api/fbs/v1/write.fbs
```diff
@@ -0,0 +1,66 @@
...
+    data_binary: [ubyte];
+    // support all of indexed fields in the fields.
+    // Field only has value, as the value of FieldValue match with the key
+    // by the index rules and index rule bindings of Metadata group.
+    // indexed fields of multiple entities are compression in the fields.
+    fields: [Field];
+}
+
+table WriteEntity {
+    // the mate_data is only required in the first query.
```
wu-sheng, line 60: `first query`?

api/fbs/v1/write.fbs
wu-sheng: Consider using `Pair`?

wu-sheng: @hanahmily Should we consider setting field name once in whole streaming?

hanahmily: The keys of fields(indexed ones) are defined in the series's metadata component. We don't have to do metadata exchange here.

The old metadata exchange path is based on tags that are removed from the query execution plan, being merged into the `data_binary`

api/fbs/v1/write.fbs
hanahmily: should be `frist write`?

api/fbs/v1/write.fbs
wu-sheng: > The keys of fields(indexed ones) are defined in the series's metadata component. We don't have to do metadata exchange here.

But how do you know the field name as this is basically an array only. Such as which element is `traceid`, or `endpoint name`.

api/fbs/v1/rpc.fbs
hanahmily: I would like to introduce the `reactive` pattern to provides some details about writing status. 
 1. failure counts.
 2. lastest write timestamp.
The above metrics can be exported to OAP self-ob component.

api/fbs/v1/write.fbs
hanahmily: For example.
The metadata is `[traceid, endpoint_name]`, the index **0** is `traceid`, and index **1** is `endpoint_name`. Once [`xxx-aa`, `/query`] gets written, based on the order of fields in metadata, banyandb will know `xxx-aa` is the `traceid`, because their index is the No.0.

api/fbs/v1/write.fbs
Fine0830: `first write` is better.

api/fbs/v1/rpc.fbs
wu-sheng: OK, make sense.

api/fbs/v1/write.fbs
wu-sheng: This concept works, but I can't see this in `Metadata`.

api/fbs/v1/write.fbs
```diff
@@ -0,0 +1,66 @@
...
+    meta_data: Metadata;
+    // the entity is required except the first query.
+    entity: entityValue;
```
wu-sheng, line 63: Why first write can't carry the entity?

api/fbs/v1/write.fbs
hanahmily: The metadata I mentioned is actually the combination of the IndexRule and IndexRuleBinding. I will avoid to use "metadata" to mention it.

api/fbs/v1/write.fbs
hanahmily: The first request also needs this. Pls polish your commnets

api/fbs/v1/write.fbs
wu-sheng: OK, you mean this is defined in the schema? Back to creating the table?

banyand/liaison/trace/trace.fbs
```diff
@@ -18,5 +18,5 @@
 namespace v1;
 
 table Trace {
-    
-}
\ No newline at end of file
+
```
hanahmily, line 21: Why do you add `trace.fbs` into the liaison module?

hanahmily:
@Fine0830 There's still some nits that should be fixed

api/fbs/v1/write.fbs
hanahmily: Exactly. Based on them, we don't define the key in write protocol

Fine0830: > Consider using `Pair`?

If we use `Pair` instead `Field`, it seems to conflict with `Pair` in the `query`.

wu-sheng: > > Consider using `Pair`?
> 
> If we use `Pair` instead `Field`, it seems to conflict with `Pair` in the `query`.

2 things. One we don't need `Pair` here according to Gao's response, we have the metadata in place. Two, I used to mean using the `Pair` if the field name is required

api/fbs/v1/rpc.fbs
```diff
@@ -16,10 +16,12 @@
...
 namespace banyandb.v1;
 
-// Query is the RPC service exposed
-rpc_service Query {
-    QueryTraces(EntityCriteria):TracesResponse;
-}
\ No newline at end of file
+// Trace is the RPC service exposed
+rpc_service Trace {
+    QueryTraces(EntityCriteria): TracesResponse;
```
lujiajing1126, line 25: Since the service already has the name `Trace`, do you still need to duplicate `Traces` in the method?

api/fbs/v1/write.fbs
```diff
@@ -0,0 +1,66 @@
...
+
+table String {
+    value: string;
+}
+
+table Int {
+    value: int64;
+}
+
+union ValueType {
```
lujiajing1126, line 30: Is it possible/necessary to distinguish single-value and multi-value data types? For example: `String`, `StringArray`? 

For now, it seems every value is a vector which may not be efficient in the storage layer.

api/fbs/v1/rpc.fbs
wu-sheng: Make sense, should be query and write.

hanahmily: +1

api/fbs/v1/write.fbs
hanahmily: Your path might reduce the data size theoretically, but I don't know how much it will help get a more compacted data format.

Since banyandb's core value is to find a more effective way to store trace data, this approach makes sense to me. Let's introduce `StringArray` and `IntArray` here

api/fbs/v1/write.fbs
```diff
@@ -0,0 +1,76 @@
...
+    Int,
+    IntArray
+}
+
+table FieldValue {
+    value: ValueType;
+}
+
+table Field {
+    value: [FieldValue];
```
lujiajing1126, line 50: Now it should be a scalar instead of a vector

lujiajing1126 (changes requested):

hanahmily (approved):
LGTM

PR #9, opened 2021-06-07
Add API for trace series, index rule and index rule binding

hanahmily:
This PR intends:

 1. Introduce schema definition: trace series, index rule, and index rule binding.
 2. Add pre-defined skywalking schema objects.
 3. Add interfaces to the Series module for retrieving schema objects.

You might notice that I added `service_name`, `service_instance_name` and  `endpoint_name` to skywalking TraceSeries(https://github.com/apache/skywalking-banyandb/compare/api-schema?expand=1#diff-9d5ac9ffaa5d323e02fdf127c4977aecc0e9c31c6de5ecff792f7e81768b31dfR23-R34), but didn't add them to the IndexRule correspondingly. They will be processed by the `MetaSeries` which will be introduced in the following up PRs. The  `MetaSereis` index service/instance/endpoint name to id. 

@lujiajing1126


banyand/series/schema/sw/trace_series.json
```diff
@@ -0,0 +1,85 @@
+{
```
hanahmily, line 1: @wu-sheng @Fine0830  This's the pre-defined schema for writing and querying as we discussed in the write API PR.

api/fbs/v1/schema.fbs
```diff
@@ -27,21 +27,50 @@ struct Duration {
...
     metadata:Metadata;
+    // fields defines accetped fields
+    fields:[FieldSpec];
+    // reserved_fields_map indicates how to index reserved fields to ingested fields
+    reserved_fields_map:TraceFieldMap;
     // duration determines how long a TraceSeries keeps its data
     duration:Duration;
-    // updated_at indicates when the TraceSeries is updated
-    updated_at:uint64;
+    // updated_nanoseconds_at indicates when the TraceSeries is updated
```
lujiajing1126, line 64: `updated_at_nanoseconds`?

banyand/series/series.go
```diff
@@ -22,29 +22,66 @@ import (
...
+}
+
+//TraceRepo contains trace and entity data
+type TraceRepo interface {
+	//FetchTrace returns data.Trace by traceID
 	FetchTrace(traceID string) (data.Trace, error)
-	FetchEntity(chunkIDs []string, fields []string) ([]data.Entity, error)
-	ScanEntity(startTime, endTime uint64, fields []string) ([]data.Entity, error)
+	//FetchEntity returns data.Entity by ChunkID
+	FetchEntity(chunkIDs []string, opt ScanOptions) ([]data.Entity, error)
```
lujiajing1126, line 52: `chunkIDs` should be `string` used here or `uint64` defined in `api/common/id.go`?

banyand/series/series.go
```diff
@@ -22,29 +22,66 @@ import (
...
+//ScanOptions contain options
+//nolint
+type ScanOptions struct {
+	projection []string
+	state      TraceState
+	limit      uint32
+}
+
+//TraceRepo contains trace and entity data
+type TraceRepo interface {
```
lujiajing1126, line 48: Do we need `metadata` for methods in `TraceRepo`?

api/fbs/v1/schema.fbs
```diff
@@ -27,21 +27,50 @@ struct Duration {
...
 // Catalog refers to a placement contains objects belonged to a particular data type
 enum Catalog:byte { Trace = 0, Log = 1, Metric = 2 }
 
 // IndexType determine the index structure under the hood
-enum IndexType:byte { Text = 0, Numerical = 1, ID = 2}
+// Fields with SeriesInternal type is reserved by Series module, would not be indexed by Index module.
+enum IndexType:byte { Text = 0, Numerical = 1, ID = 2, MultiText = 3, MultiNumberical = 4, SeriesInternal = 5 }
 
 // IndexObject defines who should be indexed.
 table IndexObject {
```
lujiajing1126, line 76: An indexName may be necessary for the object of `IndexRule`? Otherwise how can we determine the index while searching with API in `index.go`?

```go
type Repo interface {
	Search(indexName string, startTime, endTime uint64, conditions []Condition) ([]common.ChunkID, error)
}
```

banyand/series/series.go
```diff
@@ -22,29 +22,66 @@ import (
...
+type SchemaRepo interface {
+	TraceSeries() schema.TraceSeries
+	IndexRule() schema.IndexRule
+	IndexRuleBinding() schema.IndexRuleBinding
+}
+
+//IndexFilter provides methods to find a specific index related objects
+type IndexFilter interface {
+	//RulesBySubject fetches IndexRule by Series defined in IndexRuleBinding
+	RulesBySubject(ctx context.Context, subject v1.Series) ([]v1.IndexRule, error)
```
lujiajing1126, line 72: Add an API `FindRuleBySubjectAndField(s)` which returns a single IndexRule for the given field(s)

api/fbs/v1/schema.fbs
hanahmily: You can leverage methods in `IndexFilter` to get desired index rule.

banyand/series/series.go
hanahmily: `uint64`

banyand/series/series.go
hanahmily: A single field might belong to several index rules, for example, a field is indexed separately or belongs to a combinate index. so the returned rule should be a vector instead of a single item.

api/fbs/v1/schema.fbs
hanahmily: make sense

banyand/series/series.go
hanahmily: yep, we need it.

api/fbs/v1/schema.fbs
lujiajing1126: > You can leverage methods in `IndexFilter` to get desired index rule.

Oh, I see the metadata in the `IndexRule`. But do we need a `group` here? Or indexRule's `name` is globally unique?

hanahmily: IndexObject is a portion of IndexRule instead of an individual resource. We don't set a group to it. 

IndexRule has metadata that has a group.

lujiajing1126: > IndexObject is a portion of IndexRule instead of an individual resource. We don't set a group to it.
> 
> IndexRule has metadata that has a group.

Yes, I understood. But according to the signature of `Repo.Search`, we have to pass a `indexName` (string) as the first argument, specifically, which name should be given here?

hanahmily: user will input the trace series, in the skywalking scenario, the name is `sw` and the group is `default`. The combination of them is the metadata of trace series. Now you have two paths to get the index name:

1. Using methods in `SchemaRepo`. Finding IndexRuleBinding by subject(above trace series) is the first step. Then you can get the desired IndexRule by the `rule_ref` in an IndexRuleBinding.

2. A sugar method in `IndexFilter`. `RulesBySubject` provides a similar querying function as indicated in the first option. You can use it to find the IndexRule by a TraceSeries's metadata.

The indexName of `Repo.Search` is the metadata of IndexRule. But I should twist the signature of this method to support a global index name.

banyand/series/series.go
hanahmily: I add a new parameter to filter a specific field(s), pls take a look

banyand/index/index.go
```diff
@@ -31,7 +31,7 @@ type Condition struct {
 }
 
 type Repo interface {
-	Search(indexName string, startTime, endTime uint64, conditions []Condition) ([]common.ChunkID, error)
+	Search(index common.Metadata, startTime, endTime uint64, conditions []Condition) ([]common.ChunkID, error)
```
lujiajing1126, line 34: Would you complete this `Condition` struct in this PR or maybe the following ones?

lujiajing1126:
Shall we be worry about the performance hits caused by dereferencing the pointers to structs such as `Spec`, `Metadata`, etc.

banyand/index/index.go
hanahmily: I would like to left it to you to complete it .

hanahmily:
> Shall we be worry about the performance hits caused by dereferencing the pointers to structs such as `Metadata`, `Metadata.Spec` etc.
> 
> I found that almost all APIs are using values instead of pointers for structs.

I prefer to pass by value, which is much cheaper than C. furthermore, immutable parameters are safer from bugs.

PR #12, opened 2021-06-30
Feat: switch to zerolog

lujiajing1126:
The `rs/zerolog` is much faster (at least ~2x) than `uber/zap` as I have done the benchmark.

The benchmarks (run for 10 seconds) focus on a three-field(`String`, `Int`, `Float32`) logging for both `zerolog` and `zap` with either JSON encoder and CONSOLE encoder.

- Positive: Log Level >= Logger Level, i.e. will have side-effect
- Negative: Log Level < Logger Level, i.e. no output

For zap,

```
goos: darwin
goarch: amd64
pkg: ...
cpu: Intel(R) Core(TM) i5-8257U CPU @ 1.40GHz
BenchmarkVanillaZapTextPositive-8   	40449751	       341.4 ns/op	      64 B/op	       3 allocs/op
BenchmarkVanillaZapTextNegative-8   	1000000000	         2.782 ns/op	       0 B/op	       0 allocs/op
BenchmarkVanillaZapJSONNegative-8   	247553424	        52.61 ns/op	     192 B/op	       1 allocs/op
BenchmarkVanillaZapJSONPositive-8   	22189645	       547.3 ns/op	     192 B/op	       1 allocs/op
PASS
```

For zerolog,

```
goos: darwin
goarch: amd64
pkg: ...
cpu: Intel(R) Core(TM) i5-8257U CPU @ 1.40GHz
BenchmarkVanillaZerologTextPositive-8   	76793395	       158.5 ns/op	       0 B/op	       0 allocs/op
BenchmarkVanillaZerologTextNegative-8   	1000000000	         1.199 ns/op	       0 B/op	       0 allocs/op
BenchmarkVanillaZerologJSONNegative-8   	1000000000	         3.178 ns/op	       0 B/op	       0 allocs/op
BenchmarkVanillaZerologJSONPositive-8   	59657577	       215.7 ns/op	       0 B/op	       0 allocs/op
PASS
```

It is obvious that in JSON encoder + Negative scenarios, due to the better designed API, zerolog is much faster with real zero allocation.

It would be useful if we want to put more debug logs in the program without any performance issue in the production.


pkg/logger/setting.go
```diff
@@ -66,24 +68,16 @@ func Init(cfg Logging) (err error) {
...
 		return nil, err
 	}
-	var encoderConfig zap.Config
+	w := io.Writer(os.Stderr)
 	switch cfg.Env {
 	case "dev":
-		encoderConfig = zap.NewDevelopmentConfig()
-		encoderConfig.EncoderConfig.EncodeLevel = zapcore.CapitalColorLevelEncoder
-		break
+		w = zerolog.ConsoleWriter{Out: os.Stderr}
```
hanahmily, line 78: Would you use https://github.com/rs/zerolog#pretty-logging to format dev log like: 
`// Output: 2006-01-02T15:04:05Z07:00 | INFO  | ***Hello World**** foo:BAR`

hanahmily (approved):
LGTM except for some nits.
It really improved Zap's API and gets better performance. Cool!

pkg/logger/setting.go
lujiajing1126: Fixed.

```
2021-07-01T16:32:43+08:00 | DEBUG | ***register flags**** module:STANDALONE name:DATABASE registered:1 total:3
```

PR #13, opened 2021-07-02
feat: serialize the entity for Write

Fine0830:
Serialize the entity for Write.


pkg/fb/api.go
```diff
@@ -0,0 +1,371 @@
...
+	"github.com/apache/skywalking-banyandb/pkg/convert"
+)
+
+type ComponentBuilderFunc func(*flatbuffers.Builder)
+
+type WriteEntityBuilder struct {
+	*flatbuffers.Builder
+}
+
+func NewEntityBuilder() *WriteEntityBuilder {
```
hanahmily, line 35: Please insert `Write` to the func name since both `write` and `query` builders are placed in this file.

pkg/fb/api.go
```diff
@@ -0,0 +1,371 @@
...
+	v1.EntityStartDataBinaryVector(b.Builder, dataBinaryLength)
+	for i := dataBinaryLength; i >= 0; i-- {
+		b.Builder.PrependByte(byte(i))
+	}
+	dataBinaryOffset := b.Builder.EndVector(dataBinaryLength)
+
+	return dataBinaryOffset
+}
+
+func (b *WriteEntityBuilder) Build(funcs ...ComponentBuilderFunc) (*flatbuffers.Builder, error) {
```
hanahmily: Could you rename it to `WriteBuild`?

hanahmily (commented):
LGTM except for some nits. @lujiajing1126 Could you review it?

Fine0830:
Done

lujiajing1126:
LGTM.

It would be better if you can separate two builders in two files?

hanahmily:
> It would be better if you can separate two builders in two files?

The current file size is sane and doesn't need to split into two items.

PR #14, opened 2021-07-04
Query Module: Executor 

lujiajing1126:
This PR adds Executor for logical plan.

### The API of `index.Repo`

```go
type Repo interface {
        // 1) which Metadata should be used here, since metadata of IndexRule cannot be used to uniquely determine the traceSeries
        // So for now, I use metadata of `TraceSeries`
        // 2) Condition
	Search(index common.Metadata, startTime, endTime uint64, conditions []Condition) ([]common.ChunkID, error)
}
```

### Schema

The entity to be built should fully comply with the schema defined, in particular for the sparse fields.

It is not constraint currently.

### Test Dataset

We need to add a dataset for testing and performance benchmark in the near future. 

So far, I only use mock data.


hanahmily:
Cool! I will review the details soon.

> We need to add a dataset for testing and performance benchmark in the near future.

After integrating with skywalking OAP, we can record the online/real dataset which is the best choice for the testing purpose.

api/common/id.go
```diff
@@ -18,3 +18,40 @@
...
 
 type ChunkID uint64
+
+type ChunkIDs []ChunkID
+
+// HashIntersect returns an intersection of two ChunkID arrays
+// without any assumptions on the order. It uses a HashMap to mark
+// the existence of a item.
+func (c ChunkIDs) HashIntersect(other ChunkIDs) ChunkIDs {
+	if c == nil && other == nil {
```
hanahmily, line 28: The receiver `c` should not be nil from what I know. Can we remove this precondition?

pkg/logical/expr_literal.go
```diff
@@ -52,12 +57,20 @@ func (i *int64Literal) String() string {
...
 
-var _ Expr = (*int64ArrLiteral)(nil)
+var _ LiteralExpr = (*int64ArrLiteral)(nil)
 
 type int64ArrLiteral struct {
 	arr []int64
 }
 
+func (i *int64ArrLiteral) Bytes() [][]byte {
+	var b [][]byte
```
hanahmily, line 67: how about `make` a slice with the capacity as `i.arr`

hanahmily (commented):
How about move `pkg/logical` and `pkg/execution` to `pkg/query/logical` and `pkg/query/execution` since they should be classified to `query` module.

api/common/id.go
lujiajing1126: Fixed

pkg/logical/expr_literal.go
lujiajing1126: I've created a slice with capacity, for both `int64ArrLiteral` and `strArrLiteral`.

Later, we may add a general byte buffer module to optimize memory allocation.

lujiajing1126:
> How about move `pkg/logical` and `pkg/execution` to `pkg/query/logical` and `pkg/query/execution` since they should be classified to `query` module.

Moved to `pkg/query/{logical,executor}`

PR #15, opened 2021-07-12
Trace series module 

hanahmily:
These changes implement the trace series module:

 * Add scan method to KV layer.
 * Implement rough functions of trace series. @lujiajing1126 Pls review its interface,  whether it matches the query module's requirement. 
 * Implement local shard event transferring between trace series and liaison. @Fine0830 you might interest in it


Fine0830 (approved):
LGTM

banyand/series/trace/query.go
```diff
@@ -0,0 +1,253 @@
...
+	if total < 1 {
+		total = 10
+	}
+	states := make([]byte, 0, 2)
+	switch opt.State {
+	case series.TraceStateSuccess:
+		states = append(states, StateSuccess)
+	case series.TraceStateError:
+		states = append(states, StateError)
+	default:
```
lujiajing1126, line 80: I guess I've added a default `TraceState`. Do I need to remove this default field?

banyand/liaison/grpc/grpc.go
```diff
@@ -35,10 +39,33 @@ type Server struct {
...
 
-func NewServer(ctx context.Context, pipeline queue.Queue) *Server {
-	return &Server{pipeline: pipeline}
+func (s *Server) Rev(message bus.Message) (resp bus.Message) {
+	data, ok := message.Data().([]byte)
+	if !ok {
+		s.log.Warn().Msg("invalid event data type")
+		return
+	}
+	shardEvent := v1.GetRootAsShardEvent(data, 0)
```
lujiajing1126, line 51: I am curious how can we ensure that the data received is `ShardEvent`?

banyand/series/trace/common_test.go
```diff
@@ -0,0 +1,270 @@
...
+
+func getEntity(id string, binary []byte, items ...interface{}) entity {
+	return entity{
+		id:     id,
+		binary: binary,
+		items:  items,
+	}
+}
+
+func getEntityWithTS(id string, binary []byte, ts uint64, items ...interface{}) entity {
```
lujiajing1126, line 235: As I understood, we have to pass items in strict order since we have to keep the same order as defined in the schema, right?

lujiajing1126:
I only have question how can we ensure the orderness of the fields while writing. From the tests, it seems we have to manually  make alignments.

Otherwise, LGTM.

banyand/series/schema/sw/trace_series.json
```diff
@@ -33,7 +37,11 @@
       "type": "String"
     },
     {
-      "name": "state",
+      "name": "duration",
```
lujiajing1126, line 40: Finally we have `duration`!

banyand/queue/local.go
```diff
@@ -21,7 +21,7 @@ import (
 	"go.uber.org/multierr"
 
 	"github.com/apache/skywalking-banyandb/banyand/discovery"
-	"github.com/apache/skywalking-banyandb/banyand/internal/bus"
+	bus2 "github.com/apache/skywalking-banyandb/pkg/bus"
```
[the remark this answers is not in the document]
hanahmily: typo by IDE auto-generation 😅

banyand/liaison/grpc/grpc.go
hanahmily: The `event.TopicShardEvent` indicates the underlying binaries are marshaled from `ShardEvent`. That's why we introduce `common.KindVersion` and `bus.Topic` into the event bus system. 

As you know, GO lacks the generic type. We choose the interface for generics instead of reflection or code-generation because GO will support generic in the next generation, GO2. The interface style will help banyandb switch to such a generic type conveniently.

banyand/series/trace/common_test.go
hanahmily: Exactly. This convention would reduce the overhead of write data.

lujiajing1126: > Exactly. This convention would reduce the overhead of write data.

Make sense to me. But we probably need a write method per schema.

api/common/id.go
```diff
@@ -17,4 +17,9 @@
 
 package common
 
-type ChunkID uint64
+type ChunkID []byte
+type SeriesID uint64
```
lujiajing1126, line 21: What's the purpose of the `SeriesID`?

banyand/series/trace/write.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/apache/skywalking-banyandb/api/common"
+	"github.com/apache/skywalking-banyandb/api/data"
+	bydb_bytes "github.com/apache/skywalking-banyandb/pkg/bytes"
+	"github.com/apache/skywalking-banyandb/pkg/convert"
+	"github.com/apache/skywalking-banyandb/pkg/fb"
+	"github.com/apache/skywalking-banyandb/pkg/partition"
+)
+
+//TODO: To add WAL to handle crashing
```
lujiajing1126, line 34: I suppose there is already a WAL in badger

https://github.com/dgraph-io/badger/blob/master/memtable.go#L44-L54

banyand/liaison/grpc/grpc.go
lujiajing1126: I see. I found this can be guaranteed by the `bus`. Make sense to me

banyand/series/trace/write.go
hanahmily: Every badger database has a WAL which can't guarantee consistency across multiple KV databases. In a single block, trace series will create 7 badger databases, we will introduce a global WAL to make them consistent.

api/common/id.go
hanahmily: It's the routing key that is defined in `sw/trace_series.json`. Liaison leverages it to determine which shard a segment should be forward.

banyand/series/trace/query.go
hanahmily: done

PR #17, opened 2021-07-14
Add series event

hanahmily:
Introduce the series event in order to help liaison to composite `seriesID`. This event contains filed names that indicates which fields should be included for building the `seriesID`

### DIFF
--- a/api/event/discovery.go
+++ b/api/event/discovery.go
@@ -28,18 +28,12 @@ var (
 		Version: "v1",
 		Kind:    "event-shard",
 	}
-	TopicShardEvent = bus.Topic{
-		ID:   ShardEventKindVersion.String(),
-		Type: bus.ChTypeUnidirectional,
-	}
+	TopicShardEvent        = bus.UniTopic(ShardEventKindVersion.String())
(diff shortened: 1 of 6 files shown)


Fine0830 (approved):

api/event/discovery.go
```diff
@@ -32,13 +32,22 @@ var (
 		ID:   ShardEventKindVersion.String(),
 		Type: bus.ChTypeUnidirectional,
 	}
+	SeriesEventKindVersion = common.KindVersion{
+		Version: "v1",
+		Kind:    "event-series",
+	}
+	TopicSeriesEvent = bus.Topic{
```
lujiajing1126, line 39: Since we have `bus.Uni` method, we should use them

lujiajing1126:
LGTM.

As I understand, these events are defined for write, not query/read, right?

api/event/discovery.go
hanahmily: make sense

hanahmily:
> LGTM.
> 
> As I understand, these events are defined for write, not query/read, right?

yep, for the query scenario, the liaison will choose an arbitrary shard to send the request.

PR #19, opened 2021-07-18
Introduce bytebuffer pool

lujiajing1126:
In this PR, I've introduced a very simple bytebuffer to optimize byte manipulation.

The benchmark of query path has been added, the result shows approx. ~10% less allocation.

### DIFF
--- a/banyand/series/trace/common_test.go
+++ b/banyand/series/trace/common_test.go
@@ -53,8 +53,11 @@ func (b ByEntityID) Swap(i, j int) {
 	b[i], b[j] = b[j], b[i]
 }
 
-func setup(t *testing.T) (*traceSeries, func()) {
-	_ = logger.Bootstrap()
+func setup(t testing.TB) (*traceSeries, func()) {
+	_ = logger.Init(logger.Logging{
+		Env:   "test",
+		Level: "warn",
+	})
 	db, err := storage.NewDB(context.TODO(), nil)
 	assert.NoError(t, err)
 	assert.NoError(t, db.FlagSet().Parse(nil))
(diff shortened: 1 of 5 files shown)


wu-sheng:
Ha, I am thinking about this recently for OAP :)

lujiajing1126:
> Ha, I am thinking about this recently for OAP :)

You mean optimizing the memory usage for reading and writing entities with buffer?

wu-sheng:
> > Ha, I am thinking about this recently for OAP :)
> 
> You mean optimizing the memory usage for reading and writing entities with buffer?

Not related to this feature, it is an OAP internal optimization(https://github.com/apache/skywalking/issues/7322). We are improving a lot in this iteration.

lujiajing1126:
> > > Ha, I am thinking about this recently for OAP :)
> > 
> > 
> > You mean optimizing the memory usage for reading and writing entities with buffer?
> 
> Not related to this feature, it is an OAP internal optimization([apache/skywalking#7322](https://github.com/apache/skywalking/issues/7322)). We are improving a lot in this iteration.

I see. I've also tried object pooling for this PR. But it seems that it would cause some sort of performance deterioration. I guess it is related to our test fixtures which are definitely too simple.

We will try later with more realistic load after we integrate it with OAP server.

wu-sheng:
Like I said in that PR, object pooling is not always an optimization. It could have bad side effects easily in many many cases. 

That is why the OAP is considering this after 5 years when all features and architecture are ready and stable already for years

hanahmily (commented):
The pool looks like over-engineering even it reduces some allocations:

1. The benefits of this pool are not enough to introduce such a heavy component. At the time, this pool is too rough to use, we should take a long time to polish it. 
2. Every module keeps changing from time to time, As Sheng's mentioned, after 5 years, the OAP server starts to tune allocation issues. Banyandb is not ready to tune performance in such a way for now. 

Of course, some well-known performance issues should be fixed at this phase, but massive performance tuning should be avoided.

PR #21, opened 2021-07-21
Change String to Str to avoid class name conflicts

lujiajing1126:

Since we use fbs in Java, I am going to run a benchmark for fb/pb in Java.

If we still use `String`, there will be conflicts between `banyandb.v1.String` and `java.lang.String` since `flatc` cannot generate fully-qualified names.

For example, the following `String` will be `banyandb.v1.String` in the same package rather than the native `java.lang.String`.

```java
// automatically generated by the FlatBuffers compiler, do not modify

package banyandb.v1;

public final class Action {
  private Action() { }
  public static final byte Put = 0;
  public static final byte Delete = 1;

  public static final String[] names = { "Put", "Delete", };

  public static String name(int e) { return names[e]; }
}
```

### DIFF
--- a/api/fbs/Makefile
+++ b/api/fbs/Makefile
@@ -18,7 +18,16 @@
 
 VERSION ?= v1
 fbs := $(wildcard $(VERSION)/*.fbs)
+GO_FLAGS := --go-namespace $(VERSION) --gen-onefile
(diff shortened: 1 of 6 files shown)


hanahmily (approved):
LGTM cc @Fine0830

PR #22, opened 2021-07-22
Benchmark flatbuffers and protobuf

lujiajing1126:
## Benchmark env

- CPU: Intel(R) Core(TM) i5-8257U CPU @ 1.40GHz
- Memory: 8 GB 2133 MHz LPDDR3
- Java: JDK8u292b10
- protoc: 3.17.3
- protobuf-java: 3.17.2
- flatc: 2.0.0
- flatbuffers-java: 2.0.2

## Performance: Serialization+Java

```
/**
 * # JMH version: 1.32
 * # VM version: JDK 1.8.0_292, OpenJDK 64-Bit Server VM, 25.292-b10
 * # VM invoker: /Library/Java/JavaVirtualMachines/adoptopenjdk-8.jdk/Contents/Home/jre/bin/java
 * # VM options: -javaagent:/Applications/IntelliJ IDEA.app/Contents/lib/idea_rt.jar=55698:/Applications/IntelliJ IDEA.app/Contents/bin -Dfile.encoding=UTF-8
 * # Blackhole mode: full + dont-inline hint
 * # Warmup: 5 iterations, 10 s each
 * # Measurement: 5 iterations, 10 s each
 * # Timeout: 10 min per iteration
 * # Threads: 1 thread, will synchronize iterations
 * # Benchmark mode: Average time, time/op
 * <p>
 * Benchmark                                                                  Mode  Cnt     Score    Error   Units
 * WriteEntitySerializationTest.flatbuffers                                   avgt   25  3044.054 ± 34.837   ns/op
 * WriteEntitySerializationTest.flatbuffers:·gc.alloc.rate                    avgt   25   911.872 ± 10.301  MB/sec
 * WriteEntitySerializationTest.flatbuffers:·gc.alloc.rate.norm               avgt   25  3056.000 ±  0.001    B/op
 * WriteEntitySerializationTest.flatbuffers:·gc.churn.PS_Eden_Space           avgt   25   912.394 ± 10.261  MB/sec
 * WriteEntitySerializationTest.flatbuffers:·gc.churn.PS_Eden_Space.norm      avgt   25  3057.783 ± 10.009    B/op
 * WriteEntitySerializationTest.flatbuffers:·gc.churn.PS_Survivor_Space       avgt   25     0.190 ±  0.018  MB/sec
 * WriteEntitySerializationTest.flatbuffers:·gc.churn.PS_Survivor_Space.norm  avgt   25     0.637 ±  0.059    B/op
 * WriteEntitySerializationTest.flatbuffers:·gc.count                         avgt   25  3878.000           counts
 * WriteEntitySerializationTest.flatbuffers:·gc.time                          avgt   25  2168.000               ms
 * WriteEntitySerializationTest.protobuf                                      avgt   25   514.010 ± 12.638   ns/op
 * WriteEntitySerializationTest.protobuf:·gc.alloc.rate                       avgt   25  3833.648 ± 90.162  MB/sec
 * WriteEntitySerializationTest.protobuf:·gc.alloc.rate.norm                  avgt   25  2168.000 ±  0.001    B/op
 * WriteEntitySerializationTest.protobuf:·gc.churn.PS_Eden_Space              avgt   25  3835.530 ± 94.020  MB/sec
 * WriteEntitySerializationTest.protobuf:·gc.churn.PS_Eden_Space.norm         avgt   25  2168.989 ±  6.134    B/op
 * WriteEntitySerializationTest.protobuf:·gc.churn.PS_Survivor_Space          avgt   25     0.195 ±  0.020  MB/sec
 * WriteEntitySerializationTest.protobuf:·gc.churn.PS_Survivor_Space.norm     avgt   25     0.110 ±  0.011    B/op
 * WriteEntitySerializationTest.protobuf:·gc.count                            avgt   25  3629.000           counts
 * WriteEntitySerializationTest.protobuf:·gc.time                             avgt   25  2227.000               ms
 */
```

## Performance: Deserialization+Go

```
goos: darwin
goarch: amd64
pkg: github.com/apache/skywalking-banyandb/benchmark/go-bench
cpu: Intel(R) Core(TM) i5-8257U CPU @ 1.40GHz
Benchmark_Deser_Flatbuffers-8   	100000000	       730.5 ns/op	      64 B/op	       2 allocs/op
Benchmark_Deser_Protobuf-8      	14826262	      5044 ns/op	    1944 B/op	      49 allocs/op
PASS
ok  	github.com/apache/skywalking-banyandb/benchmark/go-bench	153.927s
```

## Size

For the same entity as illustrated in `WriteEntitySerializationTest.EntityModel`, (Unit in bytes)

- Flatbuffers: 512
- Protobuf: 169

which means `Protobuf` is much more compact.

## Conclusion

From the perspective of bandwidth and write performance, protobuf is definitely a better choice.

However, `flatbuffer` has better *deserialization* performance, in particular for partially read. The process of deserialization 
 (i.e. `GetRootAs***`) actually does nothing. The real deserialization happens when users try to read from the byte buffer.

## References

https://www.ida.liu.se/~nikca89/papers/networking20c.pdf

Our conclusion is similar to what has been described in the above paper. See Fig.3,4,5.


hanahmily:
Thanks, @lujiajing1126. From the result, we might drop off flatbuffer due to its performance downside. Pls submitting the golang comparison result, we want to know this is particular to Java or a common problem.

wu-sheng:
OK, that deserializaion doesn't matter, as the client side actually has to read all fields.

I am good to replace fb back to protobuf like SkyWalking does.

wu-sheng:
FYI @kezhenxu94 Like we noticed is SkyWalking perf testing. We could reject any fb at the core unless we realy are going to read 1/10 fields. But in the SkyWalking, it seems wouldn't be like this.

hanahmily:
> FYI @kezhenxu94 Like we noticed is SkyWalking perf testing. We could reject any fb at the core unless we realy are going to read 1/10 fields. But in the SkyWalking, it seems wouldn't be like this.

From the go deserialization benchmark, fb is better than pb, about 7x. Both of them read all fields. So the performance varies between client and server. 

But pb's API is much easier than fb's, which plays a critical role to replace fb. Since they have similar performance, fb is better at deserialization and pb is better at serialization, a more easy API for interoperation is the main factor needed to be considered. 

Furthermore, I use gogoproto to tune unmarshaling, it's better than before, about 40%. That proves pb has more potential for tuning, that's another advantage.

hanahmily:
@lujiajing1126 Would you pls transfer our API from flatbuffer to protobuffer. [gogoproto](https://github.com/gogo/protobuf) might be a better choice for us.

lujiajing1126:
> @lujiajing1126 Would you pls transfer our API from flatbuffer to protobuffer. [gogoproto](https://github.com/gogo/protobuf) might be a better choice for us.

Sure. I will do that after we make a final decision.

PR #23, opened 2021-07-25
Switch to protobuf

lujiajing1126:
**WARNING: Huge amount of change has been made in this PR.**

- [x] `Flatbuffers` fully removed. (I guess there is still some references in `go.sum`)
- [x] `Protbuf` introduced with adaptions made to fit with pb conventions.
- [x] (Local) Tests Pass for `pkg`, `banyand` directories.
- [x] License Check. (`proto` files are not supported by Skywalking-eyes, manually added. Skip `*.pb.go` and `*.textproto`) 
- [x] Use `buf` to lint and generate `proto`s.

### Relative timestamp in ChunkID

The timestamp contained in `chunkID` is not the same as the `timestamp` field in the entity since we are using a relative time shift. This shift will cause inaccurate Scan results.


api/proto/Makefile
```diff
@@ -15,19 +15,14 @@
...
 
 .PHONY: generate-go
-generate-go: $(fbs)
-	flatc --go --grpc $(GO_FLAGS) -o $(VERSION) $^
-
-.PHONY: generate-java
-generate-java: $(fbs)
-	flatc --java --grpc $(JAVA_FLAGS) -o $(VERSION) $^
\ No newline at end of file
+generate-go:
```
hanahmily, line 24: Would we leverage https://github.com/bufbuild/buf to check proto files and generate stubs?

banyand/series/trace/write.go
```diff
@@ -55,13 +54,16 @@ func (t *traceSeries) Write(seriesID common.SeriesID, shardID uint, entity data.
...
+	err = wp.TimeSeriesWriter(shardID, dataStoreName).Put(seriesIDBytes, entityVal.GetDataBinary(), ts)
 	if err != nil {
-		return 0, errors.Wrap(err, "failed to write traceSeries data")
+		return 0, errors.Wrap(err, "fail to write traceSeries data")
 	}
-	builder := flatbuffers.NewBuilder(0)
-	builder.Finish(fb.CopyFields(entityVal, builder))
-	err = wp.TimeSeriesWriter(shardID, fieldsStoreName).Put(seriesIDBytes, builder.FinishedBytes(), ts)
+
+	byteVal, err := proto.Marshal(entityVal)
```
hanahmily, line 62: entityVal should write into fieldsStore without databinary. Could you extract fields from entityVal,  then save them into the below kv store?

hanahmily (commented):
it looks good for now. We should improve the codes to pass the CI check. And I recommend leveraging `buf` to manage proto files.

lujiajing1126:
@hanahmily I need some help from you to setup CI workflow. It seems there is restriction of Actions usage in this repo.

Could you help setup the workflow?

hanahmily:
> @hanahmily I need some help from you to setup CI workflow. It seems there is restriction of Actions usage in this repo.
> 
> Could you help setup the workflow?

How about leave the workflow alone, let's merge it with some CI failures, I will submit  another PR to fix them

banyand/series/trace/write.go
lujiajing1126: Fixed

api/proto/Makefile
lujiajing1126: I've added `buf.yaml` and `buf.gen.yaml` to leverage `buf`.

I exclude two rules `RPC_REQUEST_STANDARD_NAME` and `RPC_RESPONSE_STANDARD_NAME` since they require `*Request` and `Response` for request and response types. Do we need to impose these two rules?

lujiajing1126:
> > @hanahmily I need some help from you to setup CI workflow. It seems there is restriction of Actions usage in this repo.
> > Could you help setup the workflow?
> 
> How about leave the workflow alone, let's merge it with some CI failures, I will submit another PR to fix them

Sure.

PR #24, opened 2021-07-26
Update CI

hanahmily:


go.mod
```diff
@@ -6,22 +6,23 @@ require (
...
+	github.com/spf13/viper v1.8.1
 	github.com/stretchr/testify v1.7.0
 	go.uber.org/atomic v1.7.0
-	go.uber.org/multierr v1.6.0
+	go.uber.org/multierr v1.7.0
 	golang.org/x/net v0.0.0-20210716203947-853a461950ff // indirect
 	golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c // indirect
 	google.golang.org/genproto v0.0.0-20210722135532-667f2b7c528f // indirect
-	google.golang.org/grpc v1.39.0
+	google.golang.org/grpc v1.40.0-dev.0.20210623211556-d9eb12feed7a
```
lujiajing1126, line 25: Why do we need to update to an unstable version `1.40.0-dev`?

banyand/index/gomock_reflect_786311153/prog.go
```diff
@@ -0,0 +1,66 @@
...
+	"reflect"
+
+	"github.com/golang/mock/mockgen/model"
+
+	pkg_ "github.com/apache/skywalking-banyandb/banyand/index"
+)
+
+var output = flag.String("output", "", "The output file name, or empty to use stdout.")
+
+func main() {
```
lujiajing1126, line 19: what's the purpose of this file?

.github/workflows/go.yml
```diff
@@ -39,8 +39,6 @@ jobs:
         uses: apache/skywalking-eyes@5c90c9698806040f50671a172a7cac17974ecebd
       - name: Update dependencies
         run: GOPROXY=https://proxy.golang.org go mod download
-      - name: Install Protoc
```
lujiajing1126, line 42: If we remove this, we should install `protoc` in another way

lujiajing1126 (commented):
`flatc` installation can be also removed.

go.mod
hanahmily: by mistake, has reverted it

banyand/index/gomock_reflect_786311153/prog.go
hanahmily: has removed it

.github/workflows/go.yml
hanahmily: done

pkg/pb/database.go
```diff
@@ -25,141 +25,141 @@ import (
 	v1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/v1"
 )
 
-type shardEventBuilder struct {
+type ShardEventBuilder struct {
```
hanahmily, line 28: The exported func in this file return unexported types, which causes the lint check to fail. @lujiajing1126 Would you pls check whether the way I fixed is correct?

lujiajing1126: > The exported func in this file return unexported types, which causes the lint check to fail. @lujiajing1126 Would you pls check whether the way I fixed is correct?

Yes. I suppose the change is correct. I intended not to expose/export the structs and only allow users to create `builder`(s) by the given functions.

I don't know why this pattern is not recommended by the linter. Ref: https://stackoverflow.com/a/21470517

api/buf.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+version: v1beta1
+build:
+  roots:
+    - ./proto
+lint:
+  use:
+    - DEFAULT
+  except:
+    - RPC_REQUEST_STANDARD_NAME
```
lujiajing1126, line 28: As I asked in the last PR #23 , do we need to impose these two rules?

scripts/build/base.mk
```diff
@@ -0,0 +1,87 @@
...
+PROTOC := $(tool_bin)/protoc
+$(PROTOC):
+	@echo "Install protoc..."
+	@if ! command -v curl >/dev/null 2>/dev/null; then echo "error: curl must be installed"  >&2; exit 1; fi
+	@if ! command -v unzip >/dev/null 2>/dev/null; then echo "error: unzip must be installed"  >&2; exit 1; fi
+	@rm -f $(tool_bin)/protoc
+	@rm -rf $(tool_include)/google
+	@mkdir -p $(tool_bin) $(tool_include)
+	$(eval protoc_tmp := $(shell mktemp -d))
+	cd $(protoc_tmp); curl -sSL https://github.com/protocolbuffers/protobuf/releases/download/v$(protoc_version)/protoc-$(protoc_version)-$(protoc_os)-$(protoc_arch).zip -o protoc.zip
```
lujiajing1126, line 52: Just a comment: this cannot work on osx-arm64 platform, i.e. Apple Chip.

(Unfortunately, I changed my working laptop last week....🤣)

lujiajing1126:
LGTM

pkg/pb/database.go
hanahmily: It's not a common pattern in go. If you want to force users to use the factory method, an interface should be better, which is a canonical pattern to gophers.

lujiajing1126: > It's not a common pattern in go. If you want to force users to use the factory method, an interface should be better, which is a canonical pattern to gophers.

Sure. Make sense to me.

api/buf.yaml
hanahmily: It's the standard/common pattern to name the request. skywalking keeps using this pattern to build grpc service.

pkg/pb/database.go
lujiajing1126: > It's not a common pattern in go. If you want to force users to use the factory method, an interface should be better, which is a canonical pattern to gophers.

I did some investigation. I seems this pattern does commonly used by developers.
See https://github.com/golang/lint/issues/210

scripts/build/base.mk
hanahmily: you can tweak it to build protoc from sources, or copy protoc into bin directory in the root path.

api/buf.yaml
lujiajing1126: > It's the standard/common pattern to name the request. skywalking keeps using this pattern to build grpc service.

So if we need to comply with the conventions, we have to change the request and response types as well as delete this `except` section.

hanahmily: Would you fix this issue?

lujiajing1126: > Would you fix this issue?

Sure I can do this later.

PR #26, opened 2021-08-03
Introduce index module with memtable only

hanahmily:
In this PR, I introduce the index module, which only contains the memtable. Other components, for instance, on-disk table and kv storage, will come in follow-up PRs. 

@lujiajing1126 The query module can integrate with the index module once this gets merged. The incomplete index module can provide full features to other modules, but don't write a massive volume data into the index module because all index data is laid in the main memory.


banyand/index/index.go
```diff
@@ -19,36 +19,230 @@ package index
...
+			Name:  []byte(compositeFieldID(object.GetName(), field.Name)),
+			Value: field.Value,
+		}, field.ChunkID))
+	}
+	return err
+}
+
+func (s *service) getShard(series common.Metadata, shardID uint) (*shard, error) {
+	id := compositeSeriesID(series.Spec)
+	ss, ok := s.meta.meta[id]
```
lujiajing1126, line 124: use `s.get(series *apiv1.Metadata)` in order to make use of read lock?

banyand/index/search.go
```diff
@@ -0,0 +1,359 @@
...
+	store := sd.store
+	searcher, hasData := store.Window(startTime, endTime)
+	if !hasData {
+		return roaring.EmptyPostingList, nil
+	}
+	tree, errBuild := buildSearchTree(searcher, indexObjectName, conditions)
+	if errBuild != nil {
+		return nil, err
+	}
+	if s.log.Should(zerolog.DebugLevel) {
```
lujiajing1126, line 60: why we need to check level here?

lujiajing1126 (changes requested):

banyand/index/search.go
hanahmily: will remove it. I have found `Event.Interface` to print `tree` instead of marshaling explicitly, so we don't need the `should` method to reduce the overhead of JSON marshaling.

banyand/index/index.go
hanahmily: done

PR #27, opened 2021-08-04
Integrate Query module with index module

lujiajing1126:

This PR intends to integrate the query module with the index module. Tests have been added to cover both numerical index and textual index.

A small issue left in the last PR has been resolved.

@hanahmily Pls have a review

## Known issues

Basically every field should be indexed, we have to modify the IndexRule textproto?


hanahmily (approved):
LGTM

PR #28, opened 2021-08-05
Fix inaccurate timestamp in the StartTimeIndex store

lujiajing1126:

- [x] First reproduce the issue
- [x] Fix issue


hanahmily (approved):
LGTM

PR #29, opened 2021-08-05
Refactor index startup and remove index field check

lujiajing1126:

### DIFF
--- a/banyand/index/index.go
+++ b/banyand/index/index.go
@@ -123,6 +123,7 @@ func (s *service) Insert(series common.Metadata, shardID uint, field *Field) err
 	}
 	objects, ok := sd.meta[field.Name]
 	if !ok {
+		s.log.Debug().Str("field", field.Name).Msg("field is not indexed")
 		return nil
 	}
 	for _, object := range objects {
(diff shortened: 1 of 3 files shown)


banyand/index/index.go
```diff
@@ -108,7 +123,7 @@ func (s *service) Insert(series common.Metadata, shardID uint, field *Field) err
 	}
 	objects, ok := sd.meta[field.Name]
 	if !ok {
-		return ErrUnknownField
+		return nil
```
hanahmily, line 126: Could you record a debug level message to indicates which fields are ignored?

banyand/index/index.go
```diff
@@ -154,6 +169,28 @@ func (s *service) GracefulStop() {
...
+	})
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err() == nil
+		default:
+			allMatches := true
+			for _, opt := range options {
+				allMatches = allMatches && opt(s.meta.meta)
```
hanahmily, line 184: Could you break the loop once getting a `false` result from `opt()` instead of iterating all options.

banyand/index/index.go
lujiajing1126: Added

banyand/index/index.go
lujiajing1126: Fixed

banyand/index/index.go
```diff
@@ -154,6 +170,31 @@ func (s *service) GracefulStop() {
...
+				if !allMatches {
+					break
+				}
```
hanahmily, line 188:
```suggestion
				if allMatches = opt(s.meta.meta); !allMatches {
					break
				}
```

PR #30, opened 2021-08-07
feat: Implement Write in the liaison

Fine0830:
Implement Write in the liaison


kezhenxu94 (commented):
The folder `skywalking-banyandb/banyand/liaison/data` seems to be copied from https://github.com/grpc/grpc-go/tree/master/examples/data , this is source code dependencies so we'd better note them in the LICENSE file NOW because this is hard to review later when releasing

kezhenxu94:
Also, as we are using Go 1.16 I'm wondering whether it's necessary to copy `banyand/liaison/data/data.go`, maybe we can use built-in [go:embed](https://golang.org/pkg/embed/) since Go 1.16

wu-sheng:
If copy happens, let's keep the original license header, and change license.

banyand/liaison/grpc/grpc.go
```diff
@@ -19,31 +19,54 @@ package grpc
...
 	"github.com/apache/skywalking-banyandb/pkg/bus"
+	"github.com/apache/skywalking-banyandb/pkg/convert"
 	"github.com/apache/skywalking-banyandb/pkg/logger"
+	"github.com/apache/skywalking-banyandb/pkg/partition"
+	"github.com/apache/skywalking-banyandb/pkg/query/logical"
 	"github.com/apache/skywalking-banyandb/pkg/run"
 )
 
+var (
+	tls      = flag.Bool("tls", false, "Connection uses TLS if true, else plain TCP")
```
lujiajing1126, line 51: As I understand, the flags should be defined in the `func (s *Server) FlagSet() *run.FlagSet`? @hanahmily

banyand/liaison/grpc/grpc.go
```diff
@@ -128,3 +166,87 @@ func (s *Server) GracefulStop() {
...
+		if ruleError != nil {
+			return ruleError
+		}
+		if s.seriesInfo.seriesEvent == nil {
+			return errors.New("No seriesEvents")
+		}
+		seriesIDLen := len(s.seriesInfo.seriesEvent.FieldNamesCompositeSeriesId)
+		var str string
+		var arr []string
+		for i := 0; i < seriesIDLen; i++ {
```
lujiajing1126, line 200: I suggest optimizing this loop by,

1) first convert `FieldNamesCompositeSeriesId` to []*FieldRef (i.e. `FieldSubscript` in your case, maybe you can make use of `Schema.CreateRef(names ...string)` to do this),
2) then iterate over fields (`writeEntity.GetEntity().GetFields()`) to compose the seriesID, 

If we do so, only one iteration is necessary.

lujiajing1126 (changes requested):

banyand/liaison/grpc/grpc.go
Fine0830: uh... I am confused. There are two arrays([]*FieldRef and []*Field), can I get results in one iteration? the Field of Write has only Value without Key.

banyand/liaison/grpc/grpc.go
```diff
@@ -128,3 +169,87 @@ func (s *Server) GracefulStop() {
...
+	for {
+		writeEntity, err := TraceWriteServer.Recv()
+		if err == io.EOF {
+			return nil
+		}
+		if err != nil {
+			return err
+		}
+
+		//log.Println("writeEntity:", writeEntity)
```
hanahmily, line 188: pls remove debugging codes.

banyand/liaison/grpc/grpc.go
```diff
@@ -128,3 +169,87 @@ func (s *Server) GracefulStop() {
...
+		metadata := common.Metadata{
+			KindVersion: apischema.SeriesKindVersion,
+			Spec:        writeEntity.GetMetadata(),
+		}
+		schema, ruleError := ana.BuildTraceSchema(context.TODO(), metadata)
+		if ruleError != nil {
+			return ruleError
+		}
+		if s.seriesInfo.seriesEvent == nil {
+			return errors.New("No seriesEvents")
```
hanahmily, line 199: Could you create a global error and return it here?

banyand/liaison/grpc/grpc.go
```diff
@@ -128,3 +169,87 @@ func (s *Server) GracefulStop() {
...
+		ana := logical.DefaultAnalyzer()
+		metadata := common.Metadata{
+			KindVersion: apischema.SeriesKindVersion,
+			Spec:        writeEntity.GetMetadata(),
+		}
+		schema, ruleError := ana.BuildTraceSchema(context.TODO(), metadata)
+		if ruleError != nil {
+			return ruleError
+		}
+		if s.seriesInfo.seriesEvent == nil {
```
hanahmily, line 198: Data race here. Pls leverage `sync.RWMutex`  to protect seriesEvent.

banyand/liaison/grpc/grpc.go
```diff
@@ -128,3 +169,87 @@ func (s *Server) GracefulStop() {
...
+					}
+				}
+			}
+		}
+		str = strings.Join(arr, "")
+		if str == "" {
+			return errors.New("invalid seriesID")
+		}
+		seriesID := []byte(str)
+		shardNum := s.shardInfo.shardEvent.GetShard().GetId()
```
hanahmily, line 229: shardEvent might be nil, pls check whether it's nil before read it. and `sync.RWMutex` is also needed.

banyand/liaison/grpc/grpc_test.go
```diff
@@ -0,0 +1,203 @@
...
+	go func() {
+		for {
+			writeResponse, errRecv := stream.Recv()
+			if errRecv == io.EOF {
+				// read done.
+				close(waitc)
+				return
+			}
+			if errRecv != nil {
+				log.Fatalf("Failed to receive data : %v", errRecv)
```
hanahmily, line 163: to use `assert.NoError`(testify) here

banyand/liaison/grpc/grpc_test.go
```diff
@@ -0,0 +1,203 @@
...
+			}
+			log.Println("writeResponse: ", writeResponse)
```
hanahmily, line 165: to user `assert` to replace `log`

banyand/series/trace/service.go
```diff
@@ -65,7 +72,9 @@ func (s *service) PreRun() error {
 		return err
 	}
 	s.schemaMap = make(map[string]*traceSeries, len(schemas))
+	s.writeListener.schemaMap = make(map[string]*traceSeries, len(schemas))
```
hanahmily, line 75: Could  you assign `s.schemaMap` to `s.writeListener.schemaMap` directly?

banyand/series/trace/service.go
```diff
@@ -65,7 +72,9 @@ func (s *service) PreRun() error {
...
 	s.l = logger.GetLogger(s.Name())
+	s.writeListener.l = logger.GetLogger(s.Name())
```
hanahmily, line 77: pls assign `s.l` to `s.writeListener.l`

hanahmily (changes requested):
The main issue is how to handle flags. Pls, leverage specific functions to set and validate flags.

Fine0830:
Done, please recheck.

banyand/liaison/grpc/grpc.go
```diff
@@ -128,3 +158,90 @@ func (s *Server) GracefulStop() {
...
+		ana := logical.DefaultAnalyzer()
+		metadata := common.Metadata{
+			KindVersion: apischema.SeriesKindVersion,
+			Spec:        writeEntity.GetMetadata(),
+		}
+		schema, ruleError := ana.BuildTraceSchema(context.TODO(), metadata)
+		if ruleError != nil {
+			return ruleError
+		}
+		s.seriesInfo.RWMutex.RLock()
```
lujiajing1126, line 185: You have to release the read lock if it steps into `s.seriesInfo.seriesEvent == nil` condition

banyand/liaison/grpc/grpc.go
```diff
@@ -128,3 +158,90 @@ func (s *Server) GracefulStop() {
...
+			case *v1.Field_Str:
+				arr = append(arr, fmt.Sprint(v.Str.Value))
+			}
+		}
+		str = strings.Join(arr, "")
+		if str == "" {
+			return ErrInvalidSeriesID
+		}
+		seriesID := []byte(str)
+		s.shardInfo.RWMutex.RLock()
```
lujiajing1126, line 218: Same here

banyand/liaison/grpc/grpc.go
```diff
@@ -102,7 +132,9 @@ func (s *Server) Name() string {
 }
 
 func (s *Server) FlagSet() *run.FlagSet {
+	size := 1024 * 1024 * 8
 	fs := run.NewFlagSet("grpc")
+	fs.Int("maxRecMsgSize", size, "the size of max receiving message")
```
lujiajing1126, line 137: where do we use this variable?

lujiajing1126 (changes requested):

banyand/liaison/grpc/grpc.go
Fine0830: Done. It's added in server options.

lujiajing1126:
LGTM!

banyand/liaison/grpc/grpc.go
```diff
@@ -52,6 +77,9 @@ func (s *shardInfo) Rev(message bus.Message) (resp bus.Message) {
 		s.log.Warn().Msg("invalid event data type")
 		return
 	}
+	s.RWMutex.Lock()
+	defer s.RWMutex.Unlock()
+	s.shardEvent = shardEvent
 	s.log.Info().
```
hanahmily, line 83: * Add two separate `mutex` to protect `shardEvent` and `seriesEvent`. They don't want to block each other.
 * Pls parse the `action` of `shardEvent` and `seriesEvent` instead of accepting it directly.

banyand/liaison/grpc/grpc.go
```diff
@@ -128,3 +210,92 @@ func (s *Server) GracefulStop() {
...
+			KindVersion: apischema.SeriesKindVersion,
+			Spec:        writeEntity.GetMetadata(),
+		}
+		schema, ruleError := ana.BuildTraceSchema(context.TODO(), metadata)
+		if ruleError != nil {
+			return ruleError
+		}
+		s.seriesInfo.RWMutex.RLock()
+		if s.seriesInfo.seriesEvent == nil {
+			s.seriesInfo.RWMutex.RUnlock()
```
hanahmily, line 239: How about adding a `s.getSeriesInfo` here like 

```go
func (s *Server) getSeriesInfo() {
    s.RWMutex.RLock()
    defer s.RWMutext.RUlock()
    return s.xxxx
}
```

The above method helps us get rid of checking where to unlock rwmutext.

banyand/liaison/grpc/testdata/server_cert.pem
```diff
@@ -0,0 +1,32 @@
+-----BEGIN CERTIFICATE-----
```
hanahmily, line 1: As @kezhenxu94 mentioned, if you copy https://github.com/grpc/grpc-go/blob/master/examples/data/x509/server_cert.pem, pls add notices and license in banyandb. But they look like only for the test purpose, how about generate them by your hand

hanahmily (changes requested):
flags and mutex issues should be fixed

banyand/liaison/grpc/testdata/server_cert.pem
Fine0830: The `server_cert.pem` and `server_key.pem` are needed in server, the `server_cert.pem` is needed in client. The license of cert and key files are Apache-2.0 License, the same as `skywalking-banyandb`, I need to add they?

banyand/liaison/grpc/grpc.go
```diff
@@ -49,26 +49,35 @@ import (
...
-	ser        *grpclib.Server
-	pipeline   queue.Queue
-	repo       discovery.ServiceRepo
-	shardInfo  *shardInfo
-	seriesInfo *seriesInfo
+	addr               string
+	maxRecvMsgSize     int
+	TLSVal             bool
+	ServerHostOverride string
+	CertFile           string
```
hanahmily, line 66: If we don't export them, switch them to lower case.

banyand/liaison/grpc/testdata/server_cert.pem
hanahmily: To copy source codes needs another process more than lib dependency. It's much easy to generate TLS files by referring to https://bbengfort.github.io/2017/03/secure-grpc/

banyand/liaison/grpc/grpc.go
Fine0830: I export them for  the `grpc_test`.

banyand/liaison/grpc/testdata/server_cert.pem
Fine0830: OK

banyand/liaison/grpc/grpc.go
hanahmily: You could follow the below piece to set those flags instead of exporting them:

```go
tcp.FlagSet().Parse([]string{"--tls=" + xxx})
```

hanahmily (approved):
LGTM

PR #31, opened 2021-08-09
Fix duplicate type names in proto

lujiajing1126:

image


wu-sheng:
Renaming is not the key. Add namespace and java package option.
Take a reference from SkyWalking protocol repository.

lujiajing1126:
> Renaming is not the key. Add namespace and java package option.
> Take a reference from SkyWalking protocol repository.

Sorry, I am confused. I suppose the conflict is due to the duplicated name for both Enum and Message defined in the `schema.proto`.

wu-sheng:
You said in Java, different proto files could have different Java package name, which could avoid this conflicts. 
Duplicate names are common, SkyWalking network proto could conflict with each other or banyandb too.

lujiajing1126:
> You said in Java, different proto files could have different Java package name, which could avoid this conflicts.
> Duplicate names are common, SkyWalking network proto could conflict with each other or banyandb too.

image

Emmm, but they are in the same file?

wu-sheng:
Same name? How go compiling passed? This seems not working in any language.

lujiajing1126:
> Same name? How go compiling passed? This seems not working in any language.

Seems strange, the compiler automatically adds prefixes to the enum values,

image

wu-sheng:
OK, and rename title and make change log corret.
Fix duplicate type names in proto.

lujiajing1126:
> OK, and rename title and make change log corret.
> Fix duplicate type names in proto.

Done.

hanahmily (approved):
It looks like only Go, protobuf native language, doesn't have this issue 😅

wu-sheng:
> It looks like only Go, protobuf native language, doesn't have this issue 😅

It is only go generator unreasonable fix. Like many go tools do.

PR #34, opened 2021-08-13
Move proto files to proper packages

hanahmily:
These changes help OAP use gRPC APIs properly. 


wu-sheng:
The latest requirements for protocol are included here, https://github.com/apache/skywalking/pull/7455

api/proto/banyandb/database/v1/schema.proto
```diff
@@ -44,10 +44,11 @@ message Duration {
 }
 
 enum FieldType {
-    FIELD_TYPE_STRING = 0;
-    FIELD_TYPE_INT = 1;
-    FIELD_TYPE_STRING_ARRAY = 2;
-    FIELD_TYPE_INT_ARRAY = 3;
+    FIELD_TYPE_UNSPECIFIED=0;
```
wu-sheng, line 47: This would break the client side. Notice, only add new index is healthy. So, please update the client side proto, and rebuild.

hanahmily: will pr the main repo soon

hanahmily: https://github.com/apache/skywalking/pull/7462

PR #37, opened 2021-08-17
Set repo CI check requirements

wu-sheng:
@kezhenxu94 Do I use the correct CI task name?

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -31,9 +31,7 @@ github:
   protected_branches:
     main:
       required_status_checks:
-        strict: true
         contexts:
           - Continuous Integration
       required_pull_request_reviews:
         dismiss_stale_reviews: true
-        required_approving_review_count: 1
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,0 +1,39 @@
+  
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+github:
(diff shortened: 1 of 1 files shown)


wu-sheng:
@hanahmily I still keep the `main` branch unprotected, direct push is still working.

.asf.yaml
```diff
@@ -0,0 +1,39 @@
...
+    rebase:  false
+  protected_branches:
+    main:
+      required_status_checks:
+        strict: true
+        contexts:
+          - Continuous Integration
+      required_pull_request_reviews:
+        dismiss_stale_reviews: true
+        required_approving_review_count: 1
```
kezhenxu94, line 39: > @hanahmily I still keep the main branch unprotected, direct push is still working.

This is not the case you said

kezhenxu94 (approved):

.asf.yaml
wu-sheng: Is this going to add branch protection? This was for review requirement

kezhenxu94: > Is this going to add branch protection? This was for review requirement

This config `strict: true` protects the branch.

hanahmily:
What I concern is do I have the legacy to merge any PRs?

.asf.yaml
kezhenxu94: > > Is this going to add branch protection? This was for review requirement
> 
> This config `strict: true` protects the branch.

And this

```
contexts:
  - Continuous Integration
```

According to GitHub

> - [x] Require status checks to pass before merging
> Choose which status checks must pass before branches can be merged into a branch that matches this rule. When enabled, commits must first be pushed to another branch, then merged or pushed directly to a branch that matches this rule after status checks have passed.

wu-sheng:
You could merge others' PRs, not yourself.

wu-sheng:
@hanahmily @kezhenxu94 I removed the review requirement for now. It should be added back once we are closing to our first release. We need to be careful after that.

.asf.yaml
```diff
@@ -0,0 +1,37 @@
...
+    - database
+  enabled_merge_buttons:
+    squash:  true
+    merge:   false
+    rebase:  false
+  protected_branches:
+    main:
+      required_status_checks:
+        contexts:
+          - Continuous Integration
```
kezhenxu94, line 35: These still not allows direct push to main branch

wu-sheng: Yes, but they can merge the PR through pull request alternatively.

kezhenxu94: If we don't want to allow direct push to main branch but allow merging committers' own PRs, this is OK to me

wu-sheng: Currently require CI passed and a committer merge(squash only).

kezhenxu94 (approved):

PR #39, opened 2021-08-17
Add strict: true for branch auto update

wu-sheng:


kezhenxu94 (approved):

wu-sheng:
We need fail fast or something? I still could merge even test fails randomly.

PR #40, opened 2021-08-17
Only run the `result` job after all needed jobs succeeded

kezhenxu94:
@wu-sheng this makes the logical job only run after all needed jobs succeeded, otherwise the logical job won't run and thus the CI will be pending

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -31,7 +31,6 @@ github:
   protected_branches:
     main:
       required_status_checks:
-        strict: true
         contexts:
           - Continuous Integration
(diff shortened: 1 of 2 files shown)


kezhenxu94:
@wu-sheng now it requires review after `strict: true`

wu-sheng:
Noticed too. Let's remove it for now. We mostly have 2 working on codes right now.

wu-sheng (approved):

kezhenxu94:
Well, I run out of idea. Seems `skipped` jobs are consider as passing in check status

PR #42, opened 2021-08-17
Separate docker build files

lujiajing1126:


.github/workflows/publish-docker-pr.yml
```diff
@@ -0,0 +1,54 @@
...
+      - name: Install goimports
+        run: go install golang.org/x/tools/cmd/goimports@latest
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+      - name: Update dependencies 
+        run: GOPROXY=https://proxy.golang.org go mod download
+      - name: Generate codes
+        run: make generate
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
```
kezhenxu94, line 51: I'm very confused why we need so many steps out of Makefile, instead of making them automatically run in Makefile by using dependent tasks so that developers only need to run `make docker` without caring about the GitHub workflow file. The same concern exists in the other workflow file `.github/workflows/publish-docker-main.yml`

.github/workflows/publish-docker-main.yml
```diff
@@ -52,17 +51,13 @@ jobs:
...
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
-      - name: Push docker image
-        if: github.ref == 'refs/heads/main'
+      - name: Build and Push docker image
         run: |
-          make docker.push || make docker.push
\ No newline at end of file
+          make docker || make docker
+          make docker.push || make docker.push
```
kezhenxu94, line 63: The previous workflow failed to push the Docker images is not because `make docker` and `make docker.push` are in different jobs, it's because `make docker` doesn't even build the Docker images, which is not reproducible in my local machine.

See this https://github.com/apache/skywalking-banyandb/runs/3349244592#step:8:378 , there is no Docker images after `make docker`

lujiajing1126: > The previous workflow failed to push the Docker images is not because `make docker` and `make docker.push` are in different jobs, it's because `make docker` doesn't even build the Docker images, which is not reproducible in my local machine.
> 
> See this https://github.com/apache/skywalking-banyandb/runs/3349244592#step:8:378 , there is no Docker images after `make docker`

There is a warning at the very beginning of the step. The docker image built only exists in the current layer. So we need to build and push in one step.

lujiajing1126: https://github.com/apache/skywalking-banyandb/runs/3349154715#step:8:12

kezhenxu94: @lujiajing1126 do you mean this `fatal: No names found, cannot describe anything.
30`? It's a git command to set a variable from what I understand and should not affect the docker task.

lujiajing1126:
```
time="2021-08-17T09:53:39Z" level=warning msg="No output specified for docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load"
```

.github/workflows/publish-docker-pr.yml
lujiajing1126: I suppose we can move `goimport` installation into Makefile and have a better dependency tree

wu-sheng:
I don't think separating is the key point. GHA shares everything in one CI task.
Let's find what it causes.

.github/workflows/publish-docker-main.yml
kezhenxu94: @lujiajing1126 check this https://github.com/apache/skywalking-banyandb/runs/3349370271 , I made your changes in this PR run in [my branch](https://github.com/apache/skywalking-banyandb/tree/test), it doesn't help at all

.github/workflows/publish-docker-pr.yml
kezhenxu94: > I suppose we can move `goimport` installation into Makefile and have a better dependency tree

Not only `goimport`, but also `make generate`

kezhenxu94 (commented):
Also, separating the `publish-docker` workflow into 2 files makes no sense IMO, the reason why we want the workflow to run (partially though) in PRs is that we want to check the commands in that workflow is runnable and don't want it to break after merging. Separating it into 2 files may lead to a situation where developers adjust the file `.github/workflows/publish-docker-pr.yml` to make the CI pass and merge the PRs, but forget to adjust `.github/workflows/publish-docker-main.yml`.

.github/workflows/publish-docker-main.yml
kezhenxu94, line 63: > Also, separating the `publish-docker` workflow into 2 files makes no sense IMO, the reason why we want the workflow to run (partially though) in PRs is that we want to check the commands in that workflow is runnable and don't want it to break after merging. Separating it into 2 files may lead to a situation where developers adjust the file `.github/workflows/publish-docker-pr.yml` to make the CI pass and merge the PRs, but forget to adjust `.github/workflows/publish-docker-main.yml`.

What I suggested is something like this:

```suggestion
          make docker || make docker
          if [[ ${{ github.ref == 'refs/heads/main' ]]; then
            make docker.push || make docker.push
          fi
```

lujiajing1126: I've added `--load` options in Makefile.

Pls check the latest CI log. It seems fine now.

lujiajing1126:
> Also, separating the `publish-docker` workflow into 2 files makes no sense IMO, the reason why we want the workflow to run (partially though) in PRs is that we want to check the commands in that workflow is runnable and don't want it to break after merging. Separating it into 2 files may lead to a situation where developers adjust the file `.github/workflows/publish-docker-pr.yml` to make the CI pass and merge the PRs, but forget to adjust `.github/workflows/publish-docker-main.yml`.

This change has been reverted.

kezhenxu94 (approved):

PR #43, opened 2021-08-21
Fix index of the field while using projections

lujiajing1126:

- [x] Add test case to reproduce the issue
- [x] Bug fix


hanahmily (approved):
LGTM

PR #44, opened 2021-08-21
Refactor binary data projection

lujiajing1126:
Previously, we include `data_binary` in the projection which causes invalid request. The error is reported by the query analyzer becuase every projection field must be a valid field defined in the schema.

However, the dataBinary is indeed a special field in the write entity, so it should equivalently be treated specifically from query side.

This PR intends to refine the query protocol with an additional bool `data_binary` field. After this PR is merged, I will sync this protocol for java client.


hanahmily (dismissed):
LGTM

hanahmily:
@lujiajing1126 pls fix ci failures

lujiajing1126:
> @lujiajing1126 pls fix ci failures

Fixed

hanahmily (approved):
LGTM

PR #45, opened 2021-09-07
Introduce stream and tsdb module

hanahmily:
The changes intend to improve the index subsystem by 

 * Introduce the series index to improve the space utilization
 * Keep the block attached index to solve the high cardinality of the entity.

At the same time, some notions and modules are introduced to replace the old ones:

 * `trace`(`log`) + `storage` -> `stream` + `tsdb`
 * Move index into `pkg`.
 * Update API to `v2`

@lujiajing1126 Please using the new modules and APIs to refactor the query module.

Once the transferring is done, we will remove old modules that are `trace`, `storage`, and `index`.


banyand/stream/stream_query_test.go
```diff
@@ -0,0 +1,532 @@
+// Licensed to Apache Software Foundation (ASF) under one or more contributor
```
hanahmily, line 1: @lujiajing1126 The test cases could help you understand how to use the new `stream` module to query data. They show us three scenarios:

 * Query data by entities which are composited by `service_id`, `service_instance_id`, and `state`
 * Query data by local indices. For example, `duration` and `endpoint_id`. You can sort data by the local indices(`duration`) or the time.
 * Query data by the global index. The `trace_id` is the only case for skywalking trace.

banyand/metadata/schema/data/index_rule_binding.json
```diff
@@ -0,0 +1,25 @@
+{
```
lujiajing1126, line 1: Now we are using `json` instead of `prototext` files?

api/proto/banyandb/stream/v2/write.proto
```diff
@@ -0,0 +1,50 @@
...
+import "banyandb/model/v2/common.proto";
+
+message ElementValue {
+  // element_id could be span_id of a Span or segment_id of a Segment in the context of stream
+  string element_id = 1;
+  // timestamp_nanoseconds is in the timeunit of nanoseconds. It represents
+  // 1) either the start time of a Span/Segment,
+  // 2) or the timestamp of a log
+  google.protobuf.Timestamp timestamp = 2;
+  message TagFamily {
```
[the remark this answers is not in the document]
hanahmily: I prefer to use `TagFamilySpec` in `schema.proto`, in which `TagSpec` is using a similar convention.

banyand/metadata/schema/data/index_rule_binding.json
hanahmily: `json` is supported by IDE widely and properly. That means we can edit these metadata more easily than `prototext`.

BTW, once `metadata` management is ready, we will remove these files finally.

banyand/stream/stream_query_test.go
```diff
@@ -0,0 +1,532 @@
...
+					elements: []string{"5", "3"},
+				},
+			},
+		},
+		{
+			name: "filter",
+			args: queryOpts{
+				entity:    tsdb.Entity{tsdb.AnyEntry, tsdb.AnyEntry, tsdb.AnyEntry},
+				timeRange: tsdb.NewTimeRangeDuration(baseTime, 1*time.Hour),
+				buildFn: func(builder tsdb.SeekerBuilder) {
```
lujiajing1126, line 204: So the searchable tags (local indexes) can be directly processed by the storage layer now?

banyand/metadata/schema/data/index_rule_binding.json
lujiajing1126: Make sense to me

banyand/stream/stream_query_test.go
```diff
@@ -0,0 +1,532 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			shards, errShards := s.Shards(nil)
+			tester.NoError(errShards)
+			err := func() error {
+				for _, shard := range shards {
+					itemIDs, err := shard.Index().Seek(index.Field{
```
lujiajing1126, line 347: global index means it is directly attached to a single shard, right?

banyand/stream/index.go
```diff
@@ -0,0 +1,121 @@
...
+	}()
+}
+
+//TODO: should listen to pipeline in a distributed cluster
+func (s *stream) writeGlobalIndex(ruleIndex indexRule, ref tsdb.GlobalItemID, value *streamv2.ElementValue) error {
+	val, err := getIndexValue(ruleIndex, value)
+	if err != nil {
+		return err
+	}
+	indexShardID, err := partition.ShardID(val, s.schema.ShardNum)
```
lujiajing1126, line 70: The global index may not in the same shard with the actual data?

banyand/kv/kv.go
```diff
@@ -206,9 +229,8 @@ func OpenIndexStore(shardID int, path string, options ...IndexOptions) (IndexSto
 	for _, opt := range options {
 		opt(bdb)
 	}
-	bdb.dbOpts = bdb.dbOpts.WithMaxLevels(1)
-	// Put all values into LSM
-	bdb.dbOpts = bdb.dbOpts.WithVLogPercentile(1.0)
+	bdb.dbOpts = bdb.dbOpts.WithMaxLevels(2)
```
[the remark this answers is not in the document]
hanahmily: Just for debugging, it's a challenge to find a single key from default 7 levels. Maybe I should parameter it.

banyand/stream/stream_query_test.go
hanahmily: Exactly.

banyand/stream/stream_query_test.go
hanahmily: it belongs to a segment in a shard. It means the global indices could be updated from time to time. 

Segment and block are the internal objects, query module could query a segment-attracted index, the global index, from a shard.

banyand/stream/index.go
hanahmily: The query result of a global index might refer to several locations across several shards. That's why we call it the "global" index.

hanahmily:
@lujiajing1126 Do you have some thoughts about these changes? I want to merge it into the main branch since it's changed a lot.

wu-sheng (approved):

lujiajing1126:
> @lujiajing1126 Do you have some thoughts about these changes? I want to merge it into the main branch since it's changed a lot.

I have no problem for this PR.

I've roughly reviewed these changes this Tuesday and Wednesday. I will read the tsdb module these days.

I will setup a meeting next Monday to discuss some more details.

PR #46, opened 2021-09-15
Implements LSM tree index and inverted index

hanahmily:
These changes implemented all kinds of index which will be used in 0.1.0


api/proto/banyandb/common/v2/common.proto
```diff
@@ -34,4 +34,5 @@ message Metadata {
     string group = 1;
     // name of the entity
     string name = 2;
+    uint32 id = 3;
```
hanahmily, line 37: @lujiajing1126 the id will be generated by metadata module. We set it for index rules manually for now.

wu-sheng (approved):

kezhenxu94 (approved):
There seems to be many codes commented out, not sure it's temporary or not needed, if not needed we should consider removing them.

hanahmily:
> There seems to be many codes commented out, not sure it's temporary or not needed, if not needed we should consider removing them.

Certainly, once the new index design is supported by every module, the outdated codes will be removed. I intend to keep them to avoid compiling failures.

PR #49, opened 2021-09-22
Implement stream query v2

lujiajing1126:
This PR implements the stream query v2 protocol while we still keep v1 to make backwards-compatibility temporarily.

A final building block of this PR is the elementID which will land in the upcoming PR by @hanahmily


lujiajing1126:
ElementID is added after #50

banyand/stream/stream_query.go
```diff
@@ -36,11 +38,25 @@ var (
 
 type Query interface {
 	Stream(stream *commonv2.Metadata) (Stream, error)
+	StreamT(stream *commonv2.Metadata) (StreamT, error)
+}
+
+// StreamT is an interface exposed specifically for testing,
+// which also includes methods defined by the interface Stream
+//nolint:golint
+type StreamT interface {
```
hanahmily, line 47: You don't need `StreamT` here. Pls, empower `Stream` instead. 

 * Move `io.Closer` into `Stream`
 * I will provide some paths to build write parameters for testing.

hanahmily (changes requested):
The changes impact too many pieces. Let's slipt it to two sub-tasks:

1. Changes about dependencies: API, stream, trace, and etc.
2. Query v2 implements.

Let's focus on the first one in this.

banyand/stream/stream_query.go
lujiajing1126: After rebase #51 and #52 , this PR has been simpilified.

Pls have a review again.

pkg/query/v2/logical/plan_indexscan.go
```diff
@@ -0,0 +1,244 @@
...
+	projectionFieldRefs [][]*FieldRef
+	entity              tsdb.Entity
+}
+
+func (i *indexScan) Execute(ec executor.ExecutionContext) ([]*streamv2.Element, error) {
+	shards, err := ec.Shards(i.entity)
+	if err != nil {
+		return nil, err
+	}
+	var elements [][]*streamv2.Element
```
hanahmily, line 122: No, we should not load all elements here.

The algorithm should be :

 1. Move all iterator to the first item by invoking their `Next`.
 2. Load all `first` items into a slice. 
 3. Pop up the minimal item through the order value, parse it to element, as the first value in the final result.
 4. Move the iterator whose value is poped in step 3, push the next value of this iterator into the slice.
 5. Repeat 3~4 until the final result reaches the limitation in the criteria.
 
This algorithm intends to use less space by loading elements that will be selected.

pkg/query/v2/logical/plan_tablescan.go
```diff
@@ -0,0 +1,195 @@
+// Licensed to Apache Software Foundation (ASF) under one or more contributor
```
hanahmily, line 1: This plan looks like a subset of indexscan. Could we merge them? or use another structure

pkg/query/v2/logical/plan_traceid.go
```diff
@@ -0,0 +1,193 @@
...
+var _ Plan = (*traceIDFetch)(nil)
+
+type unresolvedTraceIDFetch struct {
+	metadata         *commonv2.Metadata
+	traceID          string
+	projectionFields [][]*Tag
+}
+
+func (t *unresolvedTraceIDFetch) Analyze(s Schema) (Plan, error) {
+	defined, idxRule := s.IndexDefined(NewTag("", s.TraceIDFieldName()))
```
hanahmily, line 46: We don't have to parse trace_id as I mentioned above, parse global index instead.

hanahmily (commented):
Current indexscan, tablescan and trace are not suitable for current stream and tsdb module.

How about using streamscan, tsdbscan instead. This strategy can reflect the underlying structure. Furthermore, the `measure` module can reuse tsdbscan as well.

pkg/query/v2/logical/analyzer.go
```diff
@@ -0,0 +1,262 @@
...
+)
+
+var (
+	ErrFieldNotDefined            = errors.New("field is not defined")
+	ErrInvalidConditionType       = errors.New("invalid pair type")
+	ErrIncompatibleQueryCondition = errors.New("incompatible query condition type")
+	ErrInvalidSchema              = errors.New("invalid schema")
+	ErrIndexNotDefined            = errors.New("index is not define for the field")
+	ErrTraceIDWrongType           = errors.New("trace id type should be string")
+	ErrStateWrongType             = errors.New("state type should be int")
```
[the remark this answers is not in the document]
lujiajing1126: Removed since state is not a special tag now.

banyand/query/v2/query.go
```diff
@@ -0,0 +1,41 @@
...
+
+type Executor interface {
+	run.PreRunner
+}
+
+func NewExecutor(_ context.Context, streamService stream.Service, serviceRepo discovery.ServiceRepo, pipeline queue.Queue) (Executor, error) {
+	return &queryProcessor{
+		streamService: streamService,
+		serviceRepo:   serviceRepo,
+		logger:        logger.GetLogger("query"),
```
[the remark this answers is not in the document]
lujiajing1126: Removed.

pkg/query/v2/logical/analyzer.go
```diff
@@ -0,0 +1,262 @@
...
+	tags := make([]*Tag, len(tagNames))
+	for i, name := range tagNames {
+		tags[i] = NewTag(family, name)
+	}
+	return tags
+}
+
+// GetCompoundName is only used for error message
+func (t *Tag) GetCompoundName() string {
+	return t.familyName + "$" + t.name
```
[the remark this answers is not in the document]
lujiajing1126: `:` is used now. But this compound name is only for error message display. Due to the global uniqueness of tag name, we may not use familyName in most cases.

pkg/query/v2/logical/schema.go
```diff
@@ -0,0 +1,176 @@
...
+
+func (s *schema) Equal(s2 Schema) bool {
+	if other, ok := s2.(*schema); ok {
+		return cmp.Equal(other.fieldMap, s.fieldMap)
+	}
+	return false
+}
+
+// registerField registers the field spec with given tagFamilyName, tagName and indexes.
+func (s *schema) registerField(tagFamilyName, tagName string, tagFamilyIdx, tagIdx int, spec *databasev2.TagSpec) {
```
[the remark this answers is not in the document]
lujiajing1126: Removed

pkg/query/v2/logical/plan_tablescan.go
lujiajing1126: Removed. It has been merged into `IndexScan` now.

pkg/query/v2/logical/common.go
```diff
@@ -19,41 +19,32 @@ package logical
...
-	tagFamilyIdx, tagIdx := fieldRefs[0].Spec.TagFamilyIdx, fieldRefs[0].Spec.TagIdx
-	return func(a, b *streamv2.Element) bool {
-		iTag, jTag := a.GetTagFamilies()[tagFamilyIdx].GetTags()[tagIdx], b.GetTagFamilies()[tagFamilyIdx].GetTags()[tagIdx]
-		lField, _ := getRawTagValue(iTag)
-		rField, _ := getRawTagValue(jTag)
-		comp := bytes.Compare(lField, rField)
+func createComparator(sortDirection modelv2.QueryOrder_Sort) comparator {
+	return func(a, b tsdb.Item) bool {
+		comp := bytes.Compare(a.SortedField(), b.SortedField())
 		if sortDirection == modelv2.QueryOrder_SORT_ASC {
```
hanahmily, line 48: tsdb defaults sort order to ASC, which means Unspecified order should be ASC. please follow this way here

pkg/query/v2/logical/iter.go
```diff
@@ -0,0 +1,104 @@
...
+	// c is the comparator to sort items
+	c comparator
+	// iters is the list of initial Iterator
+	iters []tsdb.Iterator
+	// deq is the deque of the container
+	// 1. When we push a new container, we can normally append it to the tail of the deq,
+	//    and then sort the whole slice
+	// 2. When we pop a new container, we can just pop out the first element in the deq.
+	//    The rest of the slice is still sorted.
+	deq []*container
```
hanahmily, line 49: How about opts to `heap` package in std lib?

pkg/query/v2/logical/iter.go
```diff
@@ -0,0 +1,104 @@
...
+		deq:   make([]*container, 0),
+	}
+	iT.init()
+	return iT
+}
+
+// init function MUST be called while initialization.
+// 1. Move all iterator to the first item by invoking their Next.
+// 2. Load all first items into a slice.
+func (iT *itemIter) init() {
```
hanahmily, line 65: The receiver should be `it` or `i`.

hanahmily (commented):
The new merge-sorting algorithm looks good generally. 

BTW, could you skip current liaison test cases since they are flaky? We should polish them later

pkg/query/v2/logical/iter.go
lujiajing1126: Modified

pkg/query/v2/logical/plan_traceid.go
lujiajing1126: GlobalIndexScan has been added.

pkg/query/v2/logical/analyzer.go
```diff
@@ -0,0 +1,262 @@
...
+}
+
+func (a *Analyzer) Analyze(_ context.Context, criteria *streamv2.QueryRequest, metadata *commonv2.Metadata, s Schema) (Plan, error) {
+	// parse fields
+	plan, err := parseFields(criteria, metadata, s)
+	if err != nil {
+		return nil, err
+	}
+
+	if plan.Type() == PlanTraceIDFetch {
```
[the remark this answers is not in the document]
lujiajing1126: A question here, since global index scan are much simplified, I would like to return unresolved plan here ASAP.

But the indexes to be used are determined after analysis with schema, how can we avoid unnecessary limit and offset here?

pkg/query/v2/logical/iter.go
lujiajing1126: Done

pkg/query/v2/logical/common.go
lujiajing1126: I am skeptical about this default behavior.

lujiajing1126: Fixed

pkg/query/v2/logical/analyzer.go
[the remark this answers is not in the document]
lujiajing1126: Default Limit added to 100.

banyand/query/v2/query.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/apache/skywalking-banyandb/banyand/stream"
+	"github.com/apache/skywalking-banyandb/pkg/logger"
+	"github.com/apache/skywalking-banyandb/pkg/run"
+)
+
+type Executor interface {
+	run.PreRunner
+}
+
+func NewExecutor(_ context.Context, streamService stream.Service, serviceRepo discovery.ServiceRepo, pipeline queue.Queue) (Executor, error) {
```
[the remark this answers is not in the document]
lujiajing1126: Done

hanahmily (approved):
LGTM

PR #52, opened 2021-09-24
Add more functions to support test

hanahmily:
These changes introduce some  functions to support query module testing

### DIFF
--- a/banyand/stream/stream_query.go
+++ b/banyand/stream/stream_query.go
@@ -18,6 +18,8 @@
 package stream
 
 import (
+	"io"
+
 	"github.com/golang/protobuf/proto"
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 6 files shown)


banyand/stream/stream_query.go
```diff
@@ -39,8 +41,12 @@ type Query interface {
 }
 
 type Stream interface {
+	io.Closer
+	Write(value *streamv2.ElementValue) error
```
hanahmily, line 45: Once the write returns an empty error, indices and raw data are all there. @lujiajing1126

hanahmily:
@liuhaoyang Please take a look, do these changes support your need?

banyand/stream/stream_query.go
hanahmily: From your PR's CI failures, the stream tests are flaky. This new write function can fix this defect.

lujiajing1126:
https://github.com/apache/skywalking-banyandb/blob/main/banyand/stream/service.go#L103

Seems issue with this line. I guess we have to setup `writeListener` after this loop?

hanahmily:
> https://github.com/apache/skywalking-banyandb/blob/main/banyand/stream/service.go#L103
> 
> Seems issue with this line. I guess we have to setup `writeListener` after this loop?

Prune it.

wu-sheng (approved):

PR #55, opened 2021-10-10
Feat: metadata registry with embedded etcd

lujiajing1126:
This PR intends to add registry for `Stream`, `IndexRuleBinding` and `IndexRule` with embedded etcd.

@hanahmily Pls have a review.


hanahmily:
@lujiajing1126 Please add `Measure` related APIs in this.

lujiajing1126:
@hanahmily It seems `protoc` is included in the `buf` and what we've installed is actually not used by the `buf generate`.

hanahmily:
> @hanahmily It seems `protoc` is included in the `buf` and what we've installed is actually not used by the `buf generate`.

After some local verifications, I got the same result. How about to remove `protoc` installation process from `base.mk`?

PR #57, opened 2021-10-14
Add measure and topN APIs

hanahmily:
In this PR, I introduced 
 * The measure query and write APIs
 * TopN offline statistics specification
 * TopN query APIs

The measure-relevant APIs are easy to follow. Please refer to the comments to review them. 

Here I will use some examples to represent how to use TopN APIs

## TopN Examples

Firstly, I should define a `TopNAggregation` to generate offline statistics:

```yaml
metadata:
  name: service_cpm
  group: default
source_measure: 
  name: cpm
  group: default
field_name: value
field_value_sort: DESC
// data points from service scope will ignore "service_id"
group_by_tag_names: ["scope", "service_id"] 
...
```

Then building a `TopNRequest` to fetch the top 10 services based on above statistics:

```yaml
metadata:
  name: service_cpm
  group: default
time_range: [start, end]
top_n: 10
agg: MEAN
conditions:
- name: scope
   op: Equal
   value: "service"
```

or to fetch the top 10 instances of a service:

```yaml
metadata:
  name: service_cpm
  group: default
time_range: [start, end]
top_n: 10
agg: MEAN
conditions:
- name: scope
   op: Equal
   value: "service"
- name: service_id
   op: Equal
   value: xxxx
```

At the same time, users could use a different strategy from above: splitting service, endpoints, and instances statistics into several by the criteria property on `TopNAggregation`

@lujiajing1126


api/proto/banyandb/database/v1/schema.proto
```diff
@@ -116,14 +135,39 @@ message Measure {
     repeated FieldSpec fields = 3;
     // entity indicates which tags will be to generate a series and shard a measure
     Entity entity = 4;
-    // shard_num is the number of shards
-    uint32 shard_num = 5;
-    // ttl indicates time to live, how long the data will be cached
-    Duration ttl = 6;
+    // interval_rules define data points writing interval
+    repeated IntervalRule interval_rules = 5;
```
hanahmily, line 139: @wu-sheng As we discussed, interval_rules instrument how frequently data points in a series are written? @lujiajing1126 we should add a new tag in the java client to match these rules.

api/proto/banyandb/measure/v1/query.proto
```diff
@@ -55,21 +55,17 @@ message QueryRequest {
   model.v1.TimeRange time_range = 2;
   // offset is used to support pagination, together with the following limit
   uint32 offset = 3;
```
lujiajing1126, line 57: offset should also be remove?

wu-sheng (approved):

PR #60, opened 2021-11-20
Introduce gorilla encoding for integer values

hanahmily:
This is a variant of the original gorilla encoding:

  * store the start time in the beginning.
  * set the time value to  `1` when the value is present. Then encoding the value with the xor encoder.
  * set the time value to `0` when the value is absent. skipping the value.
  * using pre-defined interval to parse the time based on its offset to the start time.

CC @lujiajing1126


kezhenxu94 (approved):
🙈

PR #62, opened 2021-12-12
Create group on initialization

lujiajing1126:
This PR creates `default` group on initialization.

Besides, duplicated `logger` in stream service and subscription typo in metadata response are fixed.


hanahmily (approved):
LGTM

PR #63, opened 2022-01-05
Introduce preparation workflow for tests

lujiajing1126:
Previously, we setup all components and services which are necessary for tests manually, which may lead to resource leakages if an error is returned. And in some cases, we are using `testify.Require` package which calls `FailNow` at once after assertion failures.

This PR intends to introduce a workflow to prepare components/services before tests, and it could properly handle errors and  even `panic` by calling all registered `stopFunc` by previous steps.

There are still known issues:

1. `fcntl` cases

```
--- FAIL: Test_Stream_Write (5.89s)
    stream_write_test.go:253:
        	Error Trace:	stream_write_test.go:253
        	            				stream_write_test.go:42
        	Error:      	Received unexpected error:
        	            	1 error occurred:
        	            		* failed to open normal store: while opening memtables error: Unable to open mem dir.. Path=/var/folders/p0/8chn4dqn0w1c_yd0dw5byljc0000gn/T/banyandb-test-2643493976/shard-1/seg-20220105/block-1549/lsm/lsm. Error=open /var/folders/p0/8chn4dqn0w1c_yd0dw5byljc0000gn/T/banyandb-test-2643493976/shard-1/seg-20220105/block-1549/lsm/lsm: too many open files

        	Test:       	Test_Stream_Write
FAIL
FAIL	github.com/apache/skywalking-banyandb/banyand/stream	24.626s
```

We have no chance to handle `fcntl` failures while either running `startFunc` (e.g. `createDatabase`) or `stopFunc` (e.g. calling `os.RemoveAll`).

2. `os.Exit`

if someone call `os.Exit` in the middle of the program, the program may exit unexpectedly. A possible compensation may be listen to some well-known signals...

### DIFF
--- a/pkg/test/setup.go
+++ b/pkg/test/setup.go
@@ -20,8 +20,8 @@ package test
 import (
 	"context"
 
-	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
+	"go.uber.org/multierr"
 )
 
 type StartFunc func() error
@@ -46,18 +46,19 @@ type Flow interface {
 	// The timing to do the real shutdown can be determined by users.
 	Shutdown() StopFunc
 
-	ErrorOrNil() error
+	// Error returns all errors returned from startFunc(s)
+	// Nil error imply a successful flow.
+	Error() error
 }
 
 type testFlow struct {
-	err       *multierror.Error
+	err       error
 	stopFuncs []StopFunc
 }
 
 // NewTestFlow creates a flow ready to prepare services/components to be used for testing.
 func NewTestFlow() Flow {
 	return &testFlow{
-		err:       &multierror.Error{},
 		stopFuncs: make([]StopFunc, 0),
 	}
 }
@@ -77,7 +78,7 @@ func (tf *testFlow) RunWithoutSideEffect(ctx context.Context, startFunc StartFun
 }
 
 func (tf *testFlow) Run(ctx context.Context, startFunc StartFunc, stopFunc StopFunc) Flow {
-	if tf.err.ErrorOrNil() != nil {
+	if tf.err != nil {
 		return tf
 	}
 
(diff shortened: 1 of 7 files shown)


pkg/test/setup.go
```diff
@@ -0,0 +1,127 @@
...
+	// If a non-nil error is returned or panic, shutdown must be called at once.
+	// The error will be stored and thus could be checked by the caller later.
+	RunWithoutSideEffect(context.Context, StartFunc) Flow
+
+	// Shutdown does not actually shutdown the flow,
+	// but only returns a function containing all stopFunc(s) to be executed.
+	// The timing to do the real shutdown can be determined by users.
+	Shutdown() StopFunc
+
+	ErrorOrNil() error
```
hanahmily, line 49: Error() is a more conventional name.

pkg/test/setup.go
```diff
@@ -0,0 +1,127 @@
...
+		stopFuncs: make([]StopFunc, 0),
+	}
+}
+
+// Shutdown does not actually call the shutdown functions but only return a composition of all
+// stop functions given by the user thus it allows the user to determine the timing.
+func (tf *testFlow) Shutdown() StopFunc {
+	return func() {
+		for idx := len(tf.stopFuncs) - 1; idx >= 0; idx-- {
+			tf.stopFuncs[idx]()
```
hanahmily, line 70: Can one of stopFuncs stop others when it encounters a panic

hanahmily (commented):
Some nits

pkg/test/setup.go
```diff
@@ -0,0 +1,127 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package test
+
+import (
+	"context"
+
+	"github.com/hashicorp/go-multierror"
```
hanahmily, line 23: Why do you import this lib? we're using `go.uber.org/multierr`.

pkg/test/setup.go
lujiajing1126: Changed

pkg/test/setup.go
lujiajing1126: Oh, I forgot that. Removed

pkg/test/setup.go
lujiajing1126: Yes, it could happen. As I understand, the caller should take care of that, e.g. calling method with a nil pointer may cause a panic.

hanahmily (approved):
LGTM

PR #64, opened 2022-01-06
Open the tsdb on a existing path

hanahmily:
Open the database. If the path contains nothing, tsdb creates a new file layout.

Related to https://github.com/apache/skywalking/issues/8392


hanahmily:
@lujiajing1126 you could load an existing database and don't have trouble testing the OAP integration anymore.

wu-sheng:
Please fix CI.

wu-sheng (approved):

PR #65, opened 2022-01-10
WIP: metadata reload

lujiajing1126:
Currently we have to restart the `banyandb` process to reload the metadata. This PR intends to introduce runtime metadata reload.

## Tasks

- [x] Change schema preload process. Apply PR #63 to measure and query modules.
- [x] Refactor common APIs and etcd implementations to support `update`/`delete` notifications for both stream and measure entities.
- [x] Add test cases: metadata and notifications tests
- [ ] Reload metadata in `stream` and `measure` modules

In the future, I would suggest add something like [`TypeMeta` in Kubernetes](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go) to the entities such as `stream`, `measure`, `indexRule` and `indexRuleBinding` to improve runtime interoperability. Currently, we cannot know the type from interfaces like `proto.Message` except we explicitly do type assertions.


api/proto/banyandb/common/v1/common.proto
```diff
@@ -37,6 +37,20 @@ message Metadata {
...
+    // The referent should be reconciled after the current object is deleted.
+    repeated OwnerReference owner_references = 4;
+}
+
+// OwnerReference contains enough information to let you identify an owning object.
+// The group is omitted since an owning object must be in the same group as the dependent.
+// The kind of the owning object depends implicitly on the kind of the dependent.
+// Currently,
+// IndexRule --(owner_references)--> IndexRuleBinding --(explicitly with subject)--> stream/measure
+message OwnerReference {
```
hanahmily, line 51: We don't need this since indexRuleBindings have all relationships between indexrules and stream/measue.

The API has adopted the many-to-many relationship, but the owner object limits it to a one-to-many type. It conflicts the original design of this API

banyand/metadata/schema/etcd.go
```diff
@@ -130,6 +140,7 @@ func (e *etcdSchemaRegistry) DeleteGroup(ctx context.Context, group string) (boo
 	}
 	keyPrefix := GroupsKeyPrefix + g.GetName() + "/"
 	_, err = e.kv.Delete(ctx, keyPrefix, clientv3.WithRange(incrementLastByte(keyPrefix)))
+	// TODO: which resources are deleted?
```
hanahmily, line 143: What's the TODO means? Do not the resources have the same prefix?

banyand/metadata/schema/etcd.go
```diff
@@ -282,15 +314,104 @@ func (e *etcdSchemaRegistry) UpdateIndexRuleBinding(ctx context.Context, indexRu
...
+	var oldRuleRefs []string
+	old, err := e.GetIndexRuleBinding(ctx, &commonv1.Metadata{
+		Name: indexRuleBinding.GetMetadata().GetName(),
+	})
+
+	if err == nil && old != nil {
+		oldRuleRefs = old.GetRules()
+	}
+
+	err = e.update(ctx, g, &resource{
```
hanahmily, line 346: Why do you tend to update indexrule when updating indexrulebinding? They are both independent objects, don't have to sync with each other.

banyand/metadata/schema/etcd.go
```diff
@@ -402,15 +534,49 @@ func (e *etcdSchemaRegistry) get(ctx context.Context, key string, message proto.
...
 		return err
 	}
-	_, err = e.kv.Put(ctx, key, string(val))
+	_, err = e.kv.Put(ctx, res.Key(), string(val))
 	if err != nil {
 		return err
 	}
+
+	if notification {
+		switch res.typ {
```
hanahmily, line 548: From my perspective, all updating event handling processes should be identical.

banyand/metadata/schema/etcd.go
```diff
@@ -402,15 +534,49 @@ func (e *etcdSchemaRegistry) get(ctx context.Context, key string, message proto.
...
+			})
+		case ResourceIndexRule:
+			for _, ownerRef := range res.GetMetadata().GetOwnerReferences() {
+				rb, innerErr := e.GetIndexRuleBinding(ctx, &commonv1.Metadata{
+					Name:  ownerRef.GetName(),
+					Group: group.GetName(),
+				})
+
+				if innerErr != nil {
+					// TODO: possible inconsistent state?
```
hanahmily, line 567: The handler should be idempotent. The notifier has to retry the handler till it returns no error.

hanahmily (commented):
The indexrule, rulebinding, and steam/measure are all independent. The handler only tasks care of the objects it binds to.

It might cause some inconsistency when updating several objects. That's because the banyandb's APIs are using the eventual consistency model, which helps us get rid of the complicated operation to support a strong consistency model, which is not mandatory for our use cases.

lujiajing1126:
> The indexrule, rulebinding, and steam/measure are all independent. The handler only tasks care of the objects it binds to.
> 
> It might cause some inconsistency when updating several objects. That's because the banyandb's APIs are using the eventual consistency model, which helps us get rid of the complicated operation to support a strong consistency model, which is not mandatory for our use cases.

I see what you meant here. But the question is, when should the stream/measure module reload the schema? since the measure and stream modules are not aware of IndexRuleBinding and IndexRule?

hanahmily:
> > The indexrule, rulebinding, and steam/measure are all independent. The handler only tasks care of the objects it binds to.
> > It might cause some inconsistency when updating several objects. That's because the banyandb's APIs are using the eventual consistency model, which helps us get rid of the complicated operation to support a strong consistency model, which is not mandatory for our use cases.
> 
> I see what you meant here. But the question is, when should the stream/measure module reload the schema? since the measure and stream modules are not aware of IndexRuleBinding and IndexRule?

The metadata model doesn't have to know it. The stream/meausre should though. In the booting phase, stream/measure handles the indexurle/binding by themself. They are responsible for listening to the change events of indexrule and indexrulebinding, then determining whether the updated objects need to be handled.

lujiajing1126:
> > > The indexrule, rulebinding, and steam/measure are all independent. The handler only tasks care of the objects it binds to.
> > > It might cause some inconsistency when updating several objects. That's because the banyandb's APIs are using the eventual consistency model, which helps us get rid of the complicated operation to support a strong consistency model, which is not mandatory for our use cases.
> > 
> > 
> > I see what you meant here. But the question is, when should the stream/measure module reload the schema? since the measure and stream modules are not aware of IndexRuleBinding and IndexRule?
> 
> The metadata model doesn't have to know it. The stream/meausre should though. In the booting phase, stream/measure handles the indexurle/binding by themself. They are responsible for listening to the change events of indexrule and indexrulebinding, then determining whether the updated objects need to be handled.

I see. The logic should be put on the other side. I will close this PR and open a new one.

PR #66, opened 2022-01-13
Reload stream when metadata changes

lujiajing1126:
This PR supersedes the previous PR #65 to allow metadata reload while it put logic mostly in the stream module instead of pursuing a strong-consistent metadata in the previous PR.

As a result, the stream model starts a serially-running background job to continuously reconcile the opened stream (with underlying storage).

Please have a review with the new design @hanahmily

More test cases will be added later. I suppose `Eventually` method is necessary for these kinds of tests.


hanahmily:
> More test cases will be added later. I suppose `Eventually` method is necessary for these kinds of tests.

@lujiajing1126 
https://github.com/apache/skywalking-banyandb/pull/67 will introduce the `Eventually`. and I also noticed that `AfterEach` has a capacity to clean up the testing when facing a `panic`. Is the `flow` you introduced still necessary for testing?

banyand/stream/service.go
```diff
@@ -94,101 +101,266 @@ func (s *service) PreRun() error {
...
+
+func (s *service) OnDelete(m schema.Metadata) {
+	switch m.Kind {
+	case schema.KindStream:
+		s.deleteCh <- &commonv1.Metadata{
+			Name:  m.Name,
+			Group: m.Group,
+		}
+	case schema.KindIndexRuleBinding:
+		// TODO: we only have metadata(group and name) here, what shall we remove?
```
hanahmily, line 189: just ignore them

hanahmily (changes requested):
What I focus on is the process to handle failures of syncing, which is absent in the current stage. The module which is receiving events should ensure it could handle any event successfully. Otherwise, the conflicting metadata must bring us a series of troubles.

banyand/stream/service.go
```diff
@@ -52,23 +53,29 @@ type Service interface {
...
 	pipeline      queue.Queue
 	repo          discovery.ServiceRepo
-	stopCh        chan struct{}
+	// stop channel for the service
+	stopCh chan struct{}
+	// stop channel for the inner worker
+	stopCh2 chan struct{}
+
+	updateCh chan *databasev1.Stream
+	deleteCh chan *commonv1.Metadata
```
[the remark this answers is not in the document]
lujiajing1126: a single event channel is used.

banyand/stream/service.go
```diff
@@ -52,23 +53,29 @@ type Service interface {
...
 	l             *logger.Logger
 	metadata      metadata.Repo
 	root          string
 	pipeline      queue.Queue
 	repo          discovery.ServiceRepo
-	stopCh        chan struct{}
+	// stop channel for the service
+	stopCh chan struct{}
+	// stop channel for the inner worker
+	stopCh2 chan struct{}
```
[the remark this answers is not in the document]
lujiajing1126: Fixed

banyand/stream/service.go
lujiajing1126: Now, I use an option in etcd client to get the previous value while deleting keys. So it is possible to know later which entity has been deleted.

lujiajing1126:
> What I focus on is the process to handle failures of syncing, which is absent in the current stage. The module which is receiving events should ensure it could handle any event successfully. Otherwise, the conflicting metadata must bring us a series of troubles.

Currently, I just skip the processing of the given entity since the failure may probably caused by other modifications, and will be probably reaching consistent in the next round of syncing.

So, I suppose we can only focus on the metadata of the stream/measure, without any dependence on the event type, right?

If we cannot fetch the schema with the given metadata, it naturally means the entity has been deleted, so we delete the opened stream.

lujiajing1126:
> > More test cases will be added later. I suppose `Eventually` method is necessary for these kinds of tests.
> 
> @lujiajing1126 #67 will introduce the `Eventually`. and I also noticed that `AfterEach` has a capacity to clean up the testing when facing a `panic`. Is the `flow` you introduced still necessary for testing?

I agree. We may fully migrate to the `gingko`/`gomega` later in the next PR.

I suppose we can make use of `DeferCleanup`, `BeforeSuite, and `Before/AfterEach` to better organize cleanup codes. You may check the latest commits.

banyand/stream/service.go
hanahmily: nice

hanahmily:
> > What I focus on is the process to handle failures of syncing, which is absent in the current stage. The module which is receiving events should ensure it could handle any event successfully. Otherwise, the conflicting metadata must bring us a series of troubles.
> 
> Currently, I just skip the processing of the given entity since the failure may probably caused by other modifications, and will be probably reaching consistent in the next round of syncing.

FWIK, the changing triggers the metadata syncing. As I mentioned, the metadata update operation rarely happens, which will lead to a continuing inconsistency. How about introducing a periodic syncing strategy as a complementary solution?

banyand/stream/service_test.go
```diff
@@ -0,0 +1,302 @@
...
+			_ = os.RemoveAll(etcdRootDir)
+		})
+
+		// Init Stream Service
+		streamService, err = NewService(context.TODO(), metadataService, repo, pipeline)
+		Expect(err).NotTo(HaveOccurred())
+		rootPath, deferFunc, err := test.NewSpace()
+		Expect(err).NotTo(HaveOccurred())
+		DeferCleanup(func() {
+			deferFunc()
```
hanahmily, line 179: The testing temp dir is removed before the stream module managed in the run-group tends to close the kv storage.
This will cause a panic thrown on the kv closing.

lujiajing1126: Fixed

banyand/stream/service.go
```diff
@@ -94,101 +101,266 @@ func (s *service) PreRun() error {
...
+	case schema.KindStream:
+		s.updateCh <- m.Spec.(*databasev1.Stream)
+	case schema.KindIndexRuleBinding:
+		if m.Spec.(*databasev1.IndexRuleBinding).GetSubject().Catalog == commonv1.Catalog_CATALOG_STREAM {
+			stm, err := s.metadata.StreamRegistry().GetStream(context.TODO(), &commonv1.Metadata{
+				Name:  m.Name,
+				Group: m.Group,
 			})
+			if err != nil {
+				s.l.Error().Err(err).Msg("fail to get subject")
```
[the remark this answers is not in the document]
lujiajing1126: A periodic syncing has been introduced.

hanahmily (approved):

PR #70, opened 2022-01-22
Bucket strategy managment and segment controller

hanahmily:
These changes tend to introduce a bucket-based strategy manager to move segments and blocks forward. The new manager enhancements the segment in this PR. The block's going to implement it in the following up commits.

Caveat: the manager doesn't clean up segments/blocks, which are not the current reporter. Another resource controller will take it according to the memory state. 

### DIFF
--- a/banyand/tsdb/bucket/strategy.go
+++ b/banyand/tsdb/bucket/strategy.go
@@ -96,7 +96,7 @@ func (s *Strategy) Run() {
 			c := s.current.Report()
 			for {
 				select {
-				case status, more := <-c:
+				case status, closed := <-c:
 					if !more {
 						reset()
 						goto bucket
--- a/banyand/tsdb/bucket/strategy.go
+++ b/banyand/tsdb/bucket/strategy.go
@@ -91,38 +91,34 @@ func (s *Strategy) Run() {
 	reset()
 	go func(s *Strategy) {
 		var err error
+	bucket:
+		c := s.current.Report()
 		for {
-		bucket:
-			c := s.current.Report()
-			for {
-				select {
-				case status, closed := <-c:
-					if !closed {
-						reset()
-						goto bucket
-					}
-					ratio := Ratio(status.Volume) / Ratio(status.Capacity)
-					if ratio >= s.ratio && s.next == nil {
-						s.next, err = s.ctrl.Next()
-						if errors.Is(err, ErrNoMoreBucket) {
-							return
-						}
-						if err != nil {
-							s.logger.Err(err).Msg("failed to create the next bucket")
-						}
-					}
-					if ratio >= 1.0 {
-						s.current = s.next
-						s.next = nil
-						goto bucket
-					}
-				case <-s.stopCh:
-					return
+			select {
+			case status, closed := <-c:
+				if !closed {
+					reset()
+					goto bucket
 				}
+				ratio := Ratio(status.Volume) / Ratio(status.Capacity)
+				if ratio >= s.ratio && s.next == nil {
+					s.next, err = s.ctrl.Next()
+					if errors.Is(err, ErrNoMoreBucket) {
+						return
+					}
+					if err != nil {
+						s.logger.Err(err).Msg("failed to create the next bucket")
+					}
+				}
+				if ratio >= 1.0 {
+					s.current = s.next
+					s.next = nil
+					goto bucket
+				}
+			case <-s.stopCh:
+				return
 			}
-
 		}
-
 	}(s)
 }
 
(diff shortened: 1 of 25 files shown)


wu-sheng (dismissed):

banyand/tsdb/bucket/strategy.go
```diff
@@ -0,0 +1,131 @@
...
+		for s.current == nil {
+			s.current = s.ctrl.Current()
+		}
+		s.next = nil
+	}
+	reset()
+	go func(s *Strategy) {
+		var err error
+		for {
+		bucket:
```
lujiajing1126, line 95: I suppose a single `for-select` loop is enough for this case

banyand/tsdb/bucket/strategy.go
```diff
@@ -0,0 +1,131 @@
...
+			c := s.current.Report()
+			for {
+				select {
+				case status, more := <-c:
```
lujiajing1126, line 99:
```suggestion
				case status, closed := <-c:
```

banyand/tsdb/bucket/strategy.go
```diff
@@ -0,0 +1,131 @@
...
+						reset()
+						goto bucket
+					}
+					ratio := Ratio(status.Volume) / Ratio(status.Capacity)
+					if ratio >= s.ratio && s.next == nil {
+						s.next, err = s.ctrl.Next()
+						if errors.Is(err, ErrNoMoreBucket) {
+							return
+						}
+						if err != nil {
```
lujiajing1126, line 110: if err is not nil, probably it is dangerous to proceed within the current loop

banyand/tsdb/shard.go
```diff
@@ -52,66 +54,234 @@ func (s *shard) Index() IndexDatabase {
...
+	}
+	return sc.load(ctx, suffix, segPath)
+}
+
+func (sc *segmentController) load(ctx context.Context, suffix, path string) (seg *segment, err error) {
+	seg, err = openSegment(ctx, suffix, path, sc.intervalRule)
+	if err != nil {
+		return nil, err
+	}
+	{
```
lujiajing1126, line 275: unnecessary pair of braces?

banyand/tsdb/segment.go
```diff
@@ -19,42 +19,74 @@ package tsdb
...
+	interval := s.Duration() >> 4
+	if interval < 100*time.Millisecond {
+		interval = 100 * time.Millisecond
 	}
-	return greaterAndEqualStart && s.endTime.After(ts)
+	go func() {
+		defer close(ch)
+		for {
+			ticker := time.NewTicker(interval)
+			defer ticker.Stop()
```
lujiajing1126, line 54: resource leak here

lujiajing1126 (changes requested):

banyand/tsdb/segment.go
```diff
@@ -19,42 +19,74 @@ package tsdb
...
+	TimeRange
 }
 
-func (s *segment) contains(ts time.Time) bool {
-	greaterAndEqualStart := s.startTime.Equal(ts) || s.startTime.Before(ts)
-	if s.endTime.IsZero() {
-		return greaterAndEqualStart
+func (s *segment) Report() bucket.Channel {
+	ch := make(bucket.Channel)
+	interval := s.Duration() >> 4
```
lujiajing1126, line 46: why we need to divide by 2^4?

banyand/tsdb/bucket/strategy.go
hanahmily: ` s.current.Report()` will return a new channel. We should reuse it by introducing an outer loop.

banyand/tsdb/bucket/strategy.go
hanahmily: The loop keeps calling the `next` continually till it successfully. This piece could fix accidental failures.

banyand/tsdb/shard.go
hanahmily: it tends to reduce the scope of the lock

banyand/tsdb/segment.go
hanahmily: done

banyand/tsdb/segment.go
hanahmily: get rid of setting the interval by hand. The interval = the segment duration / 16

banyand/tsdb/bucket/strategy.go
lujiajing1126: But the outer `for-loop` is not necessary?

```go
var err error
bucket:
c := s.current.Report()
for {
  select { ... }
}
```

lujiajing1126:
LGTM

lujiajing1126 (approved):

PR #71, opened 2022-01-25
Check schema equality while updating

lujiajing1126:
This PR intends to reduce unnecessary update notification while doing updates for schemas.


banyand/metadata/schema/utils.go
```diff
@@ -0,0 +1,140 @@
+// Licensed to Apache Software Foundation (ASF) under one or more contributor
```
hanahmily, line 1: How about naming this file to `checker.go` instead of a meaningless `util`

banyand/metadata/schema/utils.go
```diff
@@ -0,0 +1,140 @@
...
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	commonv1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/common/v1"
+	databasev1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/database/v1"
+)
+
+type equalityChecker func(a, b proto.Message) bool
+
+var (
+	streamEqualityChecker = func(a, b proto.Message) bool {
```
hanahmily, line 32: Do you tend to invoke this checker out of the map? If not, could you remove this redundant variable?

banyand/metadata/schema/utils.go
```diff
@@ -0,0 +1,140 @@
...
+	return a.GetSeconds() == b.GetSeconds() && a.GetNanos() == b.GetNanos()
+}
+
+func subjectEqual(a, b *databasev1.Subject) bool {
+	return a.GetCatalog() == b.GetCatalog() && a.GetName() == b.GetName()
+}
+
+func resourceOptEqual(a, b *databasev1.ResourceOpts) bool {
+	return a.GetShardNum() == b.GetShardNum() && // opts
+		a.GetTtl().GetVal() == b.GetTtl().GetVal() &&
```
hanahmily, line 111: Could `a.GetTtl()` return `nil`?

hanahmily (changes requested):
I agree with the current solution, but an alternative path is to add a `version` field to the `metadata` object to compare a single point instead of several items. What's your opinion on this?

banyand/metadata/schema/utils.go
lujiajing1126: Removed

banyand/metadata/schema/utils.go
lujiajing1126: I've migrated to `protocmp`, https://pkg.go.dev/google.golang.org/protobuf/testing/protocmp.

I suppose this is the much better way to do equality check.

banyand/metadata/schema/utils.go
lujiajing1126: Renamed

lujiajing1126:
Still a bit slow while starting OAP with schema installation, I will check soon

hanahmily:
> 



> > I agree with the current solution, but an alternative path is to add a `version` field to the `metadata` object to compare a single point instead of several items. What's your opinion on this?
> 
> IHMO, I prefer not to believe user input.

yep, your way is the safest.

hanahmily (approved):

PR #74, opened 2022-01-28
Add pprof

lujiajing1126:
This PR intends to add `pprof` for runtime diagnosis

### DIFF
--- a/banyand/prof/pprof.go
+++ b/banyand/prof/pprof.go
@@ -45,7 +45,7 @@ type pprofService struct {
 
 func (p *pprofService) FlagSet() *run.FlagSet {
 	flagSet := run.NewFlagSet("prof")
-	flagSet.StringVar(&p.listenAddr, "pprof-listener-addr", "0.0.0.0:6060", "listen addr for pprof")
+	flagSet.StringVar(&p.listenAddr, "pprof-listener-addr", "127.0.0.1:6060", "listen addr for pprof")
 	return flagSet
 }
 
--- a/banyand/prof/pprof.go
+++ b/banyand/prof/pprof.go
@@ -1,0 +1,73 @@
+// Licensed to Apache Software Foundation (ASF) under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Apache Software Foundation (ASF) licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package prof
+
+import (
+	"net/http"
+	_ "net/http/pprof"
+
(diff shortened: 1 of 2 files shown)


banyand/prof/pprof.go
```diff
@@ -0,0 +1,74 @@
...
+
+type pprofService struct {
+	listenAddr string
+	stopCh     chan struct{}
+	l          *logger.Logger
+}
+
+func (p *pprofService) FlagSet() *run.FlagSet {
+	flagSet := run.NewFlagSet("prof")
+	flagSet.StringVar(&p.listenAddr, "pprof-listener-addr", "0.0.0.0:6060", "listen addr for pprof")
```
wu-sheng, line 48: Exporting to all network access? This seems a security leak.

lujiajing1126: Fixed

banyand/prof/pprof.go
```diff
@@ -0,0 +1,74 @@
...
+
+type pprofService struct {
+	listenAddr string
+	stopCh     chan struct{}
+	l          *logger.Logger
+}
+
+func (p *pprofService) FlagSet() *run.FlagSet {
+	flagSet := run.NewFlagSet("prof")
+	flagSet.StringVar(&p.listenAddr, "pprof-listener-addr", "127.0.0.1:6060", "listen addr for pprof")
```
wu-sheng, line 48: @hanahmily need your suggestion here. Hard code seems not a good practice. WDYT?

lujiajing1126: > @hanahmily need your suggestion here. Hard code seems not a good practice. WDYT?

`127.0.0.1:6060` is the default value for the flag which can be changed with command line option `--prof-listener-addr` at runtime.

wu-sheng: Then where should be documented? I feel this kind of thing will be missed from time to time.

lujiajing1126: > Then where should be documented? I feel this kind of thing will be missed from time to time.

Currently the flag system designed by @hanahmily is self-described,

```
$ ./banyand/build/bin/banyand-server standalone --help
2022-01-28T10:08:32+08:00 | DEBUG | ***register flags**** module:STANDALONE name:METADATA registered:1 total:4
2022-01-28T10:08:32+08:00 | DEBUG | ***register flags**** module:STANDALONE name:STREAM registered:2 total:4
2022-01-28T10:08:32+08:00 | DEBUG | ***register flags**** module:STANDALONE name:GRPC registered:3 total:4
2022-01-28T10:08:32+08:00 | DEBUG | ***register flags**** module:STANDALONE name:PPROF-SERVICE registered:4 total:4
Run as the standalone mode

Usage:
   standalone [flags]

Flags:
      --addr string                  The address of banyand listens (default ":17912")
      --cert-file string             The TLS cert file
  -h, --help                         help for standalone
      --key-file string              The TLS key file
      --logging.env string           the logging (default "dev")
      --logging.level string         the level of logging (default "debug")
      --max-recv-msg-size int        The size of max receiving message (default 10485760)
      --metadata-root-path string    the root path of metadata (default "/tmp")
  -n, --name string                  name of this service (default "standalone")
      --pprof-listener-addr string   listen addr for pprof (default "127.0.0.1:6060")
      --root-path string             the root path of database (default "/tmp")
      --show-rungroup-units          show rungroup units
      --tls                          Connection uses TLS if true, else plain TCP
  -v, --version                      version for standalone
```

all flags registered will be automatically shown in the help messages. I am not sure if this is what you want.

wu-sheng: OK, thanks for sharing. I am good now.

hanahmily (approved):
LGTM

PR #76, opened 2022-01-29
Fix timezone issue

lujiajing1126:
This PR fixes timezone issue introduced in PR #73 

Currently the `time.Local` is used to ensure consistency.

But I am curious why liaison test passed. @hanahmily


hanahmily (approved):

hanahmily:
> But I am curious why liaison test passed. @hanahmily

this affects the folder name instead of the internal time system which is implemented by `time.Time`

PR #77, opened 2022-02-07
Remove unnecessary PATH usage in Makefile(s)

lujiajing1126:
Fix https://github.com/apache/skywalking/issues/8506


lujiajing1126:
According to the official GNU documents,

> Variable and function references in recipes have identical syntax and semantics to references elsewhere in the makefile. They also have the same quoting rules: if you want a dollar sign to appear in your recipe, you must double it (‘$$’). For shells like the default shell, that use dollar signs to introduce variables, it’s important to keep clear in your mind whether the variable you want to reference is a make variable (use a single dollar sign) or a shell variable (use two dollar signs). For example:

So for the case of `mockgen` which is used in the go source codes with `//go:generate mockgen`, we still have to set an additional `PATH`. Then the excerpt above implies a doubled dollar sign must be used for expansion lazily in the shell context.

I've tested locally and it works well. @wu-sheng

wu-sheng (approved):

PR #78, opened 2022-02-11
Extend tsdb to contain streams of a group

hanahmily:
A stream would create a tsdb, which causes the banyandb to occupy too much memory.

This tends to extend a single tsdb's capacity to store all streams in a single group. These changes are:

 * Reuse the tsdb in the stream module.
 * Add Group resource to set the resource options, like shard number, interval rules, etc.
 * Update query, liaison, and other relevant modules.
 * Remove periodic reconciliation between stream and metadata. We should enhance the notification framework of metadata changing to relay the events to stream and other modules. 
 * Remove the measuring module. Given that the changes are too many, we will add them back in following-up commits 

@lujiajing1126 Please verify whether this could reduce memory consumption on integrating with the OAP.


wu-sheng (approved):

lujiajing1126:
I confirm that this branch significantly reduce the memory usage. But some issues,

image

hanahmily:
@lujiajing1126 I tried to fix this issue, pls verify my patch.

lujiajing1126 (approved):
The last commit works, at least for the metadata. We will first merge and then check other issues.

PR #80, opened 2022-02-18
Polish query processing 

lujiajing1126:
1. Check iterators and series returned from all shards,
2. Fix logs

### DIFF
--- a/pkg/query/logical/plan_indexscan_global.go
+++ b/pkg/query/logical/plan_indexscan_global.go
@@ -86,9 +86,10 @@ func (t *globalIndexScan) Execute(ec executor.ExecutionContext) ([]*streamv1.Ele
 	}
 	var elements []*streamv1.Element
 	for _, shard := range shards {
-		elementsInShard, err := t.executeForShard(ec, shard)
-		if err != nil {
-			return elements, err
+		elementsInShard, shardErr := t.executeForShard(ec, shard)
+		if shardErr != nil {
+			// TODO: shall we proceed?
+			return elements, shardErr
 		}
 		elements = append(elements, elementsInShard...)
 	}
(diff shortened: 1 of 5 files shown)


pkg/query/logical/plan_indexscan_global.go
```diff
@@ -86,9 +86,10 @@ func (t *globalIndexScan) Execute(ec executor.ExecutionContext) ([]*streamv1.Ele
 	}
 	var elements []*streamv1.Element
 	for _, shard := range shards {
-		elementsInShard, err := t.executeForShard(ec, shard)
-		if err != nil {
-			return elements, err
+		elementsInShard, shardErr := t.executeForShard(ec, shard)
+		if shardErr != nil {
+			// TODO: shall we proceed?
```
hanahmily, line 91: this fail-fast strategy is good to me

pkg/query/logical/plan_indexscan_local.go
```diff
@@ -173,6 +180,10 @@ func (i *localIndexScan) executeInShard(shard tsdb.Shard) ([]tsdb.Iterator, erro
 		return nil, err
 	}
 
+	if seriesList == nil || len(seriesList) == 0 {
```
hanahmily, line 183: `len()` can handle both empty and nil slices,

hanahmily (commented):
some nits

pkg/query/logical/plan_indexscan_local.go
lujiajing1126: Thx! But it is strange. Why didn't the linter report this issue?

pkg/query/logical/plan_indexscan_global.go
lujiajing1126: `TODO` removed

pkg/query/logical/plan_indexscan_local.go
lujiajing1126: Fixed

hanahmily (approved):

PR #84, opened 2022-02-23
Fix the index iterator overflow

hanahmily:
The index iterator would overflow when the index module stores several streams' indices. 


wu-sheng (approved):

lujiajing1126 (dismissed):

wu-sheng:
Test fails. Please fix.

lujiajing1126 (approved):

PR #87, opened 2022-02-28
Add groupBy to the measure query request

hanahmily:
Add groupBy and aggregation function to the query request:

 * the query request doesn't support sub or nested aggregation 
 * the response's timestamp field is null on returning the aggregated result
 * the result is as same as the `order by` if the request doesn't specify the agg function on grouping


api/proto/banyandb/measure/v1/query.proto
```diff
@@ -62,8 +62,16 @@ message QueryRequest {
...
   model.v1.TagProjection tag_projection = 5;
   message FieldProjection {
-    repeated string name = 1;
+    repeated string names = 1;
   }
   // field_projection can be used to select fields of the data points in the response
   FieldProjection field_projection = 6;
+  message GroupBy {
+    model.v1.TagProjection tag_projection = 1;
+    string field_name = 2;
```
wu-sheng, line 71: Is this name used in field projection?

api/proto/banyandb/measure/v1/query.proto
```diff
@@ -62,8 +62,16 @@ message QueryRequest {
...
+  }
+  // group_by groups data points based on their field value for sepcific tags and this field'name
```
wu-sheng, line 73:
```suggestion
  // group_by groups data points based on their field value for a specific tag and use field_name as the projection name
```

api/proto/banyandb/measure/v1/query.proto
```diff
@@ -62,8 +62,16 @@ message QueryRequest {
...
+  GroupBy group_by = 7;
+  // agg aggregates data points grouped by tags and a field's name
+  model.v1.AggregationFunction agg = 8;
```
wu-sheng, line 76: We need a projection name for `agg`.

api/proto/banyandb/measure/v1/topn.proto
```diff
@@ -60,7 +55,7 @@ message TopNRequest {
     // top_n set the how many items should be returned in each list.
     int32 top_n = 3;
     // agg aggregates lists grouped by field names in the time_range
-    AggregationFunction agg = 4;
+    model.v1.AggregationFunction agg = 4;
     // criteria select counters.
     repeated model.v1.Condition conditions = 5;
```
wu-sheng, line 60: We need a `sort` field to match `TopNAggregation#field_value_sort#UNSPECIFIED`

lujiajing1126:
In the current api design, as I understand, we can only have one single aggregation on a single field, `SELECT AVG(field_A) FROM measure_A GROUP BY tag_A`.

An alternative approach would be to separate field projection as `one_of`,

1. Direct Field Projection with a field name, `SELECT latency ...`
2. Indrect Field Projection with a field name, aggregation method and an alias, e.g. `SELECT AVG(latency) as avg_latency,  MAX(latency) as max_latency ...`

while `groupBy` is only a single tag name. WDYT?

wu-sheng:
> In the current api design, as I understand, we can only have one single aggregation on a single field, SELECT AVG(field_A) FROM measure_A GROUP BY tag_A.

This is correct. 

> An alternative approach would be to separate field projection as one_of,

I am a little confused about this `one_of`. (2) is clear, I am not sure what does (1) mean.

lujiajing1126:
> > In the current api design, as I understand, we can only have one single aggregation on a single field, SELECT AVG(field_A) FROM measure_A GROUP BY tag_A.
> 
> This is correct.
> 
> > An alternative approach would be to separate field projection as one_of,
> 
> I am a little confused about this `one_of`. (2) is clear, I am not sure what does (1) mean.

I meant the following message,

```proto
message FieldProjection {
    message Direct {
        string col = 1;
    }
    // EXPR(COL) AS ALIAS
    message Indirect {
        enum Expr {
            EXPR_UNSPECIFIED = 0;
            EXPR_COUNT = 1;
            EXPR_SUM = 2;
            EXPR_MIN = 3;
            EXPR_MAX = 4;
            EXPR_AVG = 5;
        }
        string col = 1;
        string alias = 2;
        Expr expr = 3;
    }
    oneof spec {
        // direct reference to the column
        Direct direct_ref = 1;
        // indirect reference with manipulation
        Indirect indirect_ref = 2;
    }
}
```

wu-sheng:
@hanahmily WDYT? If we prefer the current simple version, I am fine. @lujiajing1126 's proposal seems more closing to a traditional SQL style, but I assume we wouldn't go that far.

api/proto/banyandb/measure/v1/query.proto
```diff
@@ -62,8 +62,20 @@ message QueryRequest {
...
   FieldProjection field_projection = 6;
+  message GroupBy {
+    model.v1.TagProjection tag_projection = 1;
+    string field_name = 2;
+  }
+  // group_by groups data points based on their field value for a specific tag and use field_name as the projection name
+  GroupBy group_by = 7;
+  message Aggregation {
+    model.v1.AggregationFunction function = 1;
+    string field_name = 2;
```
wu-sheng, line 77: To be specific, please be clear about the rule, should this `field_name` be in `field_projection`? This is sensitive from user perspective.

api/proto/banyandb/measure/v1/query.proto
```diff
@@ -62,8 +62,23 @@ message QueryRequest {
...
+    model.v1.TagProjection tag_projection = 1;
+    // field_name must be one of fields indicated by field_projection
+    string field_name = 2;
+  }
+  // group_by groups data points based on their field value for a specific tag and use field_name as the projection name
+  GroupBy group_by = 7;
+  message Aggregation {
+    model.v1.AggregationFunction function = 1;
+    // field_name must be one of files indicated by the field_projection
+    string field_name = 2;
```
wu-sheng, line 80: So, in the general case, this field name is the source field but also aggregated based on `function`. Like `alias` == `original name`, right?

hanahmily:
Your approach mixes up field filtering and aggregation that will cause two executors, aggregation and index scanner, to access the same place. Both of them have to use switch-case or if-else to distinguish the `oneof`. Splitting them to different places would simplify the implementation of both executors. Every field in the QueryRequest dedicates to an executor. We don't tend to share them between executors. 

Furthermore, as sheng mentioned, this approach you proposed is similar to the SQL syntax. But the design principle of this API is to easily implement the query module, instead of the SQL variant which is easy to read by the human. In another word, the API could be the parsed result of SQL.

api/proto/banyandb/measure/v1/query.proto
hanahmily: yep. field_name helps the aggregation executor extract the target from the previous executor in the query processing pipeline.

wu-sheng:
@lujiajing1126 I don't think carefully about the points you and Gao's discussion. But I want to mention the basic idea of BanyanDB. In general, I think we don't run forward to the SQL syntax tree, which is too complex for us. Ultimately, even you are going to provide a SQL-like interface, we will set a very strict rule compared to SQL92(or some SQL grammar)

api/proto/banyandb/measure/v1/query.proto
wu-sheng: OK, work for me.

lujiajing1126:
> Your approach mixes up field filtering and aggregation that will cause two executors, aggregation and index scanner, to access the same place. Both of them have to use switch-case or if-else to distinguish the `oneof`. Splitting them to different places would simplify the implementation of both executors. Every field in the QueryRequest dedicates to an executor. We don't tend to share them between executors.
> 
> Furthermore, as sheng mentioned, this approach you proposed is similar to the SQL syntax. But the design principle of this API is to easily implement the query module, instead of the SQL variant which is easy to read by the human. In another word, the API could be the parsed result of SQL.

From the prospective of query engine, it makes sense to me though users may probably make mistakes more often.

Generelly, LGTM.

wu-sheng (approved):

hanahmily:
>  users may probably make mistakes more often.

May our SQL variant help the end-user 😄

wu-sheng:
@lujiajing1126 I just have a DM discussion with Gao, I think generally current design is perfectly matched the analyzed executors. It may be better to you talk with him directly to follow the context.
Right now, `group-by` is just creating a set for the aggregation, otherwise, aggregation happens(if exist) to the whole filtered data set by criteria, tag_projection and field_projection

PR #89, opened 2022-03-01
Add measure query

lujiajing1126:
This PR introduces basic measure query feature with local index scan.

The implementation is based on,

1. No global index for measure
2. No limit and offset for measure

As we've discussed, GroupBy and Aggregation will come after this PR.


lujiajing1126:
> > No global index for measure
> 
> Could you explain a little about this? I think only stream(a.k.a trace and log) has global index for trace id, is this `global indext` meaning?

Exactly. We can define a global index in the schema, e.g. traceID in Segment. It is not hard-coded in the program.

lujiajing1126:
> > > > No global index for measure
> > > 
> > > 
> > > Could you explain a little about this? I think only stream(a.k.a trace and log) has global index for trace id, is this `global indext` meaning?
> > 
> > 
> > Exactly. We can define a global index in the schema, e.g. traceID in Segment. It is not hard-coded in the program.
> 
> Potentially, should consider disabling the global index in measure? Just discussing.

I agree. Currently, we don't impose any constraint on the schema.

PR #90, opened 2022-03-07
Add a prefix to the root path

hanahmily:
Add the prefix in case of both modules boot up at the same time to conflict with each other.

### DIFF
--- a/banyand/liaison/grpc/stream_test.go
+++ b/banyand/liaison/grpc/stream_test.go
@@ -55,7 +55,7 @@ var _ = Describe("Stream", func() {
 	})
 	It("is a plain server", func() {
 		By("Verifying an empty server")
-		flags := []string{"--root-path=" + rootPath, "--metadata-root-path=" + metadataPath}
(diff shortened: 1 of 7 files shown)


lujiajing1126:
In the flagSet, I guess we should use different names since the pflag system dose not accept duplicated flag name.

hanahmily:
> In the flagSet, I guess we should use different names since the pflag system dose not accept duplicated flag name.
> 
> https://github.com/apache/skywalking-banyandb/blob/7f957627bb893c9bb9360174e32fc886e1f71b64/pkg/run/run.go#L200-L204

thanks for pointing out this.

lujiajing1126:
> > In the flagSet, I guess we should use different names since the pflag system dose not accept duplicated flag name.
> > https://github.com/apache/skywalking-banyandb/blob/7f957627bb893c9bb9360174e32fc886e1f71b64/pkg/run/run.go#L200-L204
> 
> thanks for pointing out this.

Then we have to adapt all test cases 😢

hanahmily:
> Then we have to adapt all test cases 😢

Fixed all cases

lujiajing1126 (approved):
LGTM

PR #92, opened 2022-03-10
Support measure aggregation

lujiajing1126:
1. Add GroupBy and Aggregation ops to measure query,
2. Separate aggregation to a standalone package to be reused for topn aggregation


wu-sheng:
Great to see aggregation and groupBy query are being added.

hanahmily (approved):

PR #93, opened 2022-03-19
Improve the measure module due to the integration

hanahmily:
* API: move "interval" from Group to Measure
 * API: introduce "ID" tag type
 * Test: add skywalking style mock data


hanahmily:
@lujiajing1126 the new measure mocking data could be as a reference how to implement the OAP storage plugin.

wu-sheng (approved):
If @lujiajing1126 feels good.

lujiajing1126 (approved):
LGTM

PR #94, opened 2022-03-26
Support top function on measure querying

hanahmily:
This top function is distinct from the approximate top_n strategy based on the pre-calculation. This feature leverages the query model totally to do ad-hoc top queries.


wu-sheng (approved):

hanahmily:
@lujiajing1126 please use this function to support top-n queries in OAP to implement the java client. The pre-defined version will be introduced in the next iteration.

PR #99, opened 2022-04-04
Add measure tests to liaison

lujiajing1126:
Add integration tests for Measure.


hanahmily (dismissed):
LGTM

wu-sheng:
Tests fail, please recheck.

wu-sheng (approved):

PR #100, opened 2022-04-10
Return error created from status

lujiajing1126:
Fix https://github.com/apache/skywalking/issues/8830
And it should land before https://github.com/apache/skywalking-banyandb-java-client/pull/10

1. `InvalidArgument` with detailed message is return if `group` is empty,
2. If no entity is found for `?GetRequest`, `NotFound` will be returned,
3. If existing entity is found for `?CreateRequest`, `AlreadyExists` will be returned.

Currently, `?UpdateRequest` will always succeed whenever entity exists, i.e. existing resource will be overwritten or new resource will be created.


hanahmily:
> Currently, `?UpdateRequest` will always succeed whenever entity exists, i.e. existing resource will be overwritten or new resource will be created.

I prefer to return `NotFound` for an unknown entity.

banyand/metadata/schema/etcd.go
```diff
@@ -192,7 +191,7 @@ func (e *etcdSchemaRegistry) get(ctx context.Context, key string, message proto.
 	return nil
 }
 
-func (e *etcdSchemaRegistry) update(ctx context.Context, metadata Metadata) error {
+func (e *etcdSchemaRegistry) update(ctx context.Context, metadata Metadata, allowOverwrite bool) error {
```
hanahmily, line 194: How about splitting this into `update` and `create`. they have different preconditions.

banyand/metadata/schema/error.go
```diff
@@ -0,0 +1,59 @@
...
+// which describes violations in a client request.
+func BadRequest(field, desc string) error {
+	v := &errdetails.BadRequest_FieldViolation{
+		Field:       field,
+		Description: desc,
+	}
+	br := &errdetails.BadRequest{}
+	br.FieldViolations = append(br.FieldViolations, v)
+	st, err := statusGRPCInvalidArgument.WithDetails(br)
+	if err != nil {
```
hanahmily, line 52: Err, which indicates `statusGRPCInvalidArgument`'s code is 0, is always nil. Let's ignore it.

hanahmily (commented):
It seems that metadata should provide fine-grained operations for managing resources.

lujiajing1126:
A `create` method has been added, please have a check again @hanahmily

banyand/metadata/schema/error.go
hanahmily: @lujiajing1126 AFAIK, you don't have to handle this error.

hanahmily:
@lujiajing1126 it looks good to me expect for the error handler

banyand/metadata/schema/error.go
lujiajing1126: > @lujiajing1126 AFAIK, you don't have to handle this error.

Fixed.

hanahmily (approved):

PR #104, opened 2022-04-16
Make sure index rule ID is set by

hanahmily:
* Add a default value based on 32 bits CRC to the ID
  * Check the ID on getting and listing index rules

Fixed https://github.com/apache/skywalking/issues/8846

### DIFF
--- a/banyand/metadata/schema/index.go
+++ b/banyand/metadata/schema/index.go
@@ -21,7 +21,6 @@ import (
 	"context"
 	"hash/crc32"
 
-	"github.com/pkg/errors"
 	"google.golang.org/protobuf/proto"
 
 	commonv1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/common/v1"
@@ -31,7 +30,6 @@ import (
 var (
 	IndexRuleBindingKeyPrefix = "/index-rule-bindings/"
 	IndexRuleKeyPrefix        = "/index-rules/"
-	ErrEmptyIndexRuleID       = errors.New("the index rule id is empty")
 )
 
 func (e *etcdSchemaRegistry) GetIndexRuleBinding(ctx context.Context, metadata *commonv1.Metadata) (*databasev1.IndexRuleBinding, error) {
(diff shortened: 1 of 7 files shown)


banyand/metadata/schema/index.go
```diff
@@ -19,16 +19,19 @@ package schema
 
 import (
 	"context"
+	"hash/crc32"
 
 	"google.golang.org/protobuf/proto"
 
 	commonv1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/common/v1"
 	databasev1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/database/v1"
+	"github.com/pkg/errors"
```
lujiajing1126, line 28: the location of this import line is wrong

banyand/metadata/schema/index.go
```diff
@@ -93,6 +96,9 @@ func (e *etcdSchemaRegistry) GetIndexRule(ctx context.Context, metadata *commonv
 	if err := e.get(ctx, formatIndexRuleKey(metadata), &entity); err != nil {
 		return nil, err
 	}
+	if entity.Metadata.Id == 0 {
+		return nil, errors.WithMessagef(ErrEmptyIndexRuleID, "index rule: %v", metadata)
```
lujiajing1126, line 100: Shall we return grpc status code since it it API related

banyand/metadata/schema/index.go
```diff
@@ -108,12 +114,21 @@ func (e *etcdSchemaRegistry) ListIndexRule(ctx context.Context, opt ListOpt) ([]
 	}
 	entities := make([]*databasev1.IndexRule, 0, len(messages))
 	for _, message := range messages {
-		entities = append(entities, message.(*databasev1.IndexRule))
+		entity := message.(*databasev1.IndexRule)
+		if entity.Metadata.Id == 0 {
+			return nil, errors.WithMessagef(ErrEmptyIndexRuleID, "index rule: %v", entity.Metadata)
```
lujiajing1126, line 119: same here

lujiajing1126 (changes requested):

banyand/metadata/schema/index.go
hanahmily: forget to run `make pre-push`

banyand/metadata/schema/index.go
hanahmily: Make sense. I will introduce `DataLoss` to indicate this kind of data corruption.

lujiajing1126 (approved):
LGTM

PR #106, opened 2022-04-18
Introduce the clock mock to test block controller

hanahmily:
This tends to introduce a clock mock to convince the tsdb block controller testing.

  * Add a mock clock to mock `time` package.
  * Remove the  `millisecond` interval time unit which is only for testing.

fixes https://github.com/apache/skywalking/issues/8844

### DIFF
--- a/banyand/tsdb/block.go
+++ b/banyand/tsdb/block.go
@@ -82,7 +82,7 @@ func newBlock(ctx context.Context, opts blockOpts) (b *block, err error) {
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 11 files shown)


lujiajing1126 (approved):

banyand/tsdb/tsdb_suite_test.go
```diff
@@ -34,6 +34,6 @@ func TestTsdb(t *testing.T) {
 var _ = BeforeSuite(func() {
 	Expect(logger.Init(logger.Logging{
 		Env:   "dev",
-		Level: "info",
+		Level: "debug",
```
lujiajing1126, line 37: Shall we revert this back?

lujiajing1126 (approved):

PR #110, opened 2022-04-26
Introduce gorilla encoder to measure

hanahmily:
Assemble gorilla encoder/decoder to integer fields of a measure.

fix https://github.com/apache/skywalking/issues/8847

This fixes the server-side of https://github.com/apache/skywalking/issues/8948 at the same time.

### DIFF
--- a/banyand/measure/measure.go
+++ b/banyand/measure/measure.go
@@ -19,6 +19,7 @@ package measure
 
 import (
 	"context"
+	"time"
 
 	commonv1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/common/v1"
 	databasev1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/database/v1"
@@ -27,6 +28,7 @@ import (
 	"github.com/apache/skywalking-banyandb/pkg/logger"
 	"github.com/apache/skywalking-banyandb/pkg/partition"
 	pbv1 "github.com/apache/skywalking-banyandb/pkg/pb/v1"
+	"github.com/apache/skywalking-banyandb/pkg/timestamp"
 )
 
 // a chunk is 1MB
@@ -44,6 +46,7 @@ type measure struct {
 	entityLocator          partition.EntityLocator
 	indexRules             []*databasev1.IndexRule
 	indexWriter            *index.Writer
+	interval               time.Duration
 }
 
 func (s *measure) GetSchema() *databasev1.Measure {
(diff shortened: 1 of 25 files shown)


banyand/measure/measure.go
```diff
@@ -70,10 +73,14 @@ func (s *measure) Close() error {
 	return s.indexWriter.Close()
 }
 
-func (s *measure) parseSpec() {
+func (s *measure) parseSpec() (err error) {
 	s.name, s.group = s.schema.GetMetadata().GetName(), s.schema.GetMetadata().GetGroup()
 	s.entityLocator = partition.NewEntityLocator(s.schema.GetTagFamilies(), s.schema.GetEntity())
 	s.maxObservedModRevision = pbv1.ParseMaxModRevision(s.indexRules)
+	if s.schema.Interval != "" {
```
lujiajing1126, line 80: It is possible that the `Interval` is empty?

banyand/measure/measure_write.go
```diff
@@ -196,8 +195,8 @@ func (w *writeCallback) Rev(message bus.Message) (resp bus.Message) {
 	return
 }
 
-func familyIdentity(name string, flag byte) []byte {
-	return bytes.Join([][]byte{[]byte(name), {flag}}, nil)
+func familyIdentity(name string, flag []byte) []byte {
+	return bytes.Join([][]byte{tsdb.Hash([]byte(name)), flag}, nil)
```
lujiajing1126, line 199: It seems strange to use `bytes.Join` here?

pkg/timestamp/duration.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"strconv"
+	"strings"
+	"time"
+)
+
+// ParseDuration parses a duration string.
+// A duration string is a possibly signed sequence of
+// decimal numbers, each with optional fraction and a unit suffix,
+// such as "300ms", "-1.5h" or "2h45m".
```
lujiajing1126, line 29: Emmm, in which scenarios, we should support negative duration?

banyand/measure/measure.go
hanahmily: absolutely, the interval of `service_traffic` is empty.

banyand/measure/measure_write.go
hanahmily: Any thoughts here?  I applied this pattern everywhere.

pkg/timestamp/duration.go
hanahmily: No special scenario. It's an extension of the std time module, and I don't tend to remove the negative support.

banyand/measure/measure.go
lujiajing1126: > absolutely, the interval of `service_traffic` is empty.

No. I suppose for `service_traffic`, it is `Minute`.

banyand/measure/measure_write.go
lujiajing1126: > Any thoughts here? I applied this pattern everywhere. 

more or less OK for me. Let's optimize later~

banyand/measure/measure.go
wu-sheng: @lujiajing1126 All `*Traffic` data doesn't put time dimensionality into the ID. Basically, traffic is only TTL relative.

lujiajing1126: > @lujiajing1126 All `*Traffic` data doesn't put time dimensionality into the ID. Basically, traffic is only TTL relative.

Is that because we don't have any `field` for `*Traffic`? 

Since interval is mostly for GORILLA encoding, it is meaningless in such scenarios

hanahmily: The interval is bound to the field encoding algorithm, not to the TTL.

lujiajing1126 (approved):

PR #114, opened 2022-05-15
Correct the behaviour of group by

hanahmily:
The grouping operation should remove duplicated data points, but it fails. This change tends to fix the incorrect behavior.


lujiajing1126:
What about `groupBy` with count aggregation?

Since we don't have any field, it seems to be impossible to reuse the existing field.

wu-sheng:
For auto complete case, we don't need count, we need the group by keys in the results.
For event case, we need pagination only.
We don't have to support all cases in current stage.

lujiajing1126 (approved):

PR #117, opened 2022-05-16
Enhance the measure query

hanahmily:
Introduced a series of enhancements:

* Add paging operation: limit, offset and order by
* Default time range criterion of the stream querying to [0, maxnanosecond]


api/proto/banyandb/measure/v1/query.proto
```diff
@@ -90,4 +90,10 @@ message QueryRequest {
   } 
   // top limit the result based on a particular field
   Top top = 9;
+  // offset is used to support pagination, together with the following limit
+  uint32 offset = 10;
```
lujiajing1126, line 94: But what's the difference between `top` and `limit/orderBy`?

pkg/timestamp/nano.go
```diff
@@ -69,6 +72,7 @@ var (
 
 	ErrTimeOutOfRange     = errors.Errorf("time is out of range %d - %d", MinNanoTime, MaxNanoTime)
 	ErrTimeNotMillisecond = errors.Errorf("time is not millisecond precision")
+	ErrTimeEmpty          = errors.Errorf("time is empty")
```
lujiajing1126, line 75: These errors should contain gRPC status codes. Currently client cannot recognize these error types.

lujiajing1126 (changes requested):

api/proto/banyandb/measure/v1/query.proto
wu-sheng: It seems we should merge these two?

hanahmily: `top` operation is based on `field` while `limit/order by` is on `tag`.  They are different.

lujiajing1126: > `top` operation is based on `field` while `limit/order by` is on `tag`. They are different.

Oh, I see. Then which one has high priority? Or they are exclusive?

hanahmily: We can apply them at the same time. `limit` returns the results based on `top`'s output.

hanahmily: I will add more details in the comments.

pkg/timestamp/nano.go
hanahmily: done

api/proto/banyandb/measure/v1/query.proto
hanahmily: done

lujiajing1126: Normally, we don't use topN mixed with limit/offset right?

lujiajing1126:
After this PR is merged, shall we pass a default time range in the client? The logic is moved to the server, right?

api/proto/banyandb/measure/v1/query.proto
hanahmily: Exactly. We don't have such a use case in OAP integration.

hanahmily:
> After this PR is merged, shall we pass a default time range in the client? The logic is moved to the server, right?

Yes, the server will provide a default range.

lujiajing1126 (approved):
LGTM now

PR #127, opened 2022-06-02
Fix protobuf  file typo

remGigo:
Fix protobuf  file typo

### DIFF
--- a/api/proto/banyandb/property/v1/rpc.proto
+++ b/api/proto/banyandb/property/v1/rpc.proto
@@ -56,7 +56,7 @@ message GetResponse {
 }
 
 message ListRequest {
-    banyandb.common.v1.Metadata contaner = 1;
+    banyandb.common.v1.Metadata container = 1;
 }
 
 message ListResponse {
--- a/api/proto/banyandb/property/v1/property.proto
+++ b/api/proto/banyandb/property/v1/property.proto
@@ -28,7 +28,7 @@ import "banyandb/model/v1/query.proto";
 
 // Metadata is for multi-tenant use
 message Metadata {
-    // container is created when it recevies the first property
+    // container is created when it receives the first property
     common.v1.Metadata container = 1;
     // id identifies a property
     string id = 2; 


lujiajing1126:
You have to generate `*.proto.go` files locally and commit them.

api/proto/banyandb/property/v1/rpc.proto
```diff
@@ -56,7 +56,7 @@ message GetResponse {
 }
 
 message ListRequest {
-    banyandb.common.v1.Metadata contaner = 1;
+    banyandb.common.v1.Metadata container = 1;
```
lujiajing1126, line 59: When you change this file, the references in the source codes also have to be changed.

remGigo: I see, and I wonder is "make generate" at root directory all i need to do ? By the way, if i run "make generate", there are over a hundred files that have their format changed, most of which are removed with some whitespace lines in the "import" module. Is that ok?

lujiajing1126: > I see, and I wonder is "make generate" at root directory all i need to do ? By the way, if i run "make generate", there are over a hundred files that have their format changed, most of which are removed with some whitespace lines in the "import" module. Is that ok?

A hundred files? Not possible, it will only make necessary changes. I guess in your case, only `rpc.pb.go` will change.

remGigo: emmm, this is the result of "make generate", it may adjust the format ?

lujiajing1126: The formatted codes (in 3 blocks) are generated by `goimports`. Have you installed the latest `goimports` globally?

hanahmily:
Please run `make generate` and `make format` to pass the check test.

hanahmily:
Then you have to modify this reference https://github.com/apache/skywalking-banyandb/blob/main/banyand/liaison/grpc/property.go#L63

api/proto/banyandb/property/v1/rpc.proto
hanahmily: To issue `make format` after generating codes.

hanahmily:
I fixed this in https://github.com/apache/skywalking-banyandb/pull/128

api/proto/banyandb/property/v1/rpc.proto
remGigo: > The formatted codes (in 3 blocks) are generated by `goimports`. Have you installed the latest `goimports` globally and make sure it can be found within the `PATH`?
> 
> Pls refer to the official document, https://pkg.go.dev/golang.org/x/tools/cmd/goimports

I ignored goimports error information. My fault, thanks!!

PR #128, opened 2022-06-02
Add NOTICE to binary package and fix some typo

hanahmily:


wu-sheng (dismissed):

wu-sheng:
@lujiajing1126 I think client lib should follow this fix too? Although gRPC/protobuf makes this change compatible with previous versions.

lujiajing1126 (approved):

wu-sheng (approved):

lujiajing1126:
> @lujiajing1126 I think client lib should follow this fix too? Although gRPC/protobuf makes this change compatible with previous versions.

Java Client does not have this typo. While moving code there, I've already fixed it.

https://github.com/apache/skywalking-banyandb-java-client/blob/main/src/main/proto/banyandb/v1/banyandb-property.proto#L77

hanahmily:
> @lujiajing1126 I think client lib should follow this fix too? Although gRPC/protobuf makes this change compatible with previous versions.

After checking the java-client's proto files, I didn't find similar types. It seems they don't affect the interoperations between client and server.

lujiajing1126:
> > @lujiajing1126 I think client lib should follow this fix too? Although gRPC/protobuf makes this change compatible with previous versions.
> 
> After checking the java-client's proto files, I didn't find similar typo. It seems they don't affect the interoperations between client and server.

Yes. The protocol itself only cares about the position and the type of the field.

PR #129, opened 2022-06-03
Update Go to 1.18

lujiajing1126:
Update Go to 1.18. Close https://github.com/apache/skywalking/issues/9169

Notable dependencies upgrade,

- `buf`: from 1.0.0 to 1.5.0
- `protobuf` and gen tools: from 1.17.1 to 1.18.0
- `klauspost/compress`: from 1.13.1 to 1.15.6
- `grpc` and gen tools: from 1.39.0 to 1.47.0

PR #130, opened 2022-06-06
Add bydbctl and ui projects

hanahmily:
Fixes https://github.com/apache/skywalking/issues/9162


banyand/webapp/webapp.go
```diff
@@ -0,0 +1,153 @@
...
+type service struct {
+	listenAddr string
+	mux        *http.ServeMux
+	stopCh     chan struct{}
+	l          *logger.Logger
+}
+
+func (p *service) FlagSet() *run.FlagSet {
+	flagSet := run.NewFlagSet("")
+	flagSet.StringVar(&p.listenAddr, "webapp-listener-addr", ":17913", "listen addr for webapp")
```
lujiajing1126, line 55: We will server gRPC-web protocol on this port or `17912`?

hanahmily: I prefer this port. The frontend could access API through `/api/*`

wu-sheng:
Because you are adding vue into this repo as well as further release, you have to keep binary license updated as well.

.github/workflows/publish-docker.yml
```diff
@@ -58,12 +63,12 @@ jobs:
...
         uses: docker/login-action@v1.10.0
-        if: github.ref == 'refs/heads/main'
+        # if: github.ref == 'refs/heads/main'
         with:
           registry: ${{ env.HUB }}
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Push docker image
-        if: github.ref == 'refs/heads/main'
+        # if: github.ref == 'refs/heads/main'
```
wu-sheng, line 72: TODO, are these temp changing?

hanahmily: It's only for testing and this PR is not ready for review

banyand/webapp/webapp.go
lujiajing1126: I agree. Use one protocol for a single port may be the best practice.

banyand/webapp/webapp.go
```diff
@@ -0,0 +1,153 @@
...
+	p.l = logger.GetLogger(p.Name())
+	fSys, err := fs.Sub(ui.DistContent, "dist")
+	if err != nil {
+		return err
+	}
+	p.mux = http.NewServeMux()
+	httpFS := http.FS(fSys)
+	fileServer := http.FileServer(http.FS(fSys))
+	serveIndex := serveFileContents("index.html", httpFS)
+	p.mux.Handle("/", intercept404(fileServer, serveIndex))
```
lujiajing1126, line 77: Then we have to mount `/api` here.

Shall we consider some web framework like `gin`?

hanahmily: I didn't find any reason to import `gin` or other frameworks.

lujiajing1126: > I didn't find any reason to import `gin` or other frameworks.

I see, currently the `http` stdlib is enough here. I'm not familiar with web-related ecosystem in Go.

hanahmily: stdlib aligned our requirement which is to serve the SPA frontend and RESTFul APIs. The main repo selected a similar webapp integration path. The difference between them is that java stdlib doesn't provide a handy HTTP server and we have to adapt to the spring ecosystem.

dist/LICENSE
```diff
@@ -174,6 +174,20 @@
       of your accepting any such warranty or additional liability.
 
 
+========================================================================
```
hanahmily, line 177: @kezhenxu94 . I failed to include the license dep checking into CI due to the order of groups cannot stay the same. Skywalking-eye changed the content on rerunning `dep resolve`. Any thoughts?

dist/LICENSE
```diff
@@ -201,58 +216,6 @@ BSD-3-Clause and Apache-2.0 and MIT and BSD-3-Clause and MIT licenses
 
     github.com/klauspost/compress v1.15.6 BSD-3-Clause and Apache-2.0 and MIT and BSD-3-Clause and MIT
 
-========================================================================
```
wu-sheng, line 204: Are these removed recently?

dist/LICENSE
lujiajing1126: I also want to mention this issue since in the last PR #129 there's huge amount of changes due to the unstable order.

dist/LICENSE
hanahmily: They might relate to https://github.com/apache/skywalking-banyandb/pull/129

hanahmily: As I mentioned at https://github.com/apache/skywalking-banyandb/pull/130#discussion_r890748705, if skywalking eye could output coherent results, we can add the license generator to the CI to find these changes ASAP

dist/LICENSE
hanahmily: We should not rerun `make license-dep` except for releasing for now.

dist/LICENSE
wu-sheng: @kezhenxu94 @fgksgf @mrproliu Any suggestion?

wu-sheng (approved):

lujiajing1126 (approved):
LGTM

dist/LICENSE
kezhenxu94: Should be fixed in https://github.com/apache/skywalking-eyes/pull/115

PR #132, opened 2022-06-13
Do some chores

hanahmily:
* Introduce more strict checkers
  * Replace yarn with npm
  * Add node version checker
  * Bump up skywalking-eye


wu-sheng:
Please fix CI.

lujiajing1126:
Why we have to use `yarn` over `npm`?

ui/tools/checkversion.js
```diff
@@ -0,0 +1,32 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+const fs = require("fs");
+
+const nvmVersion = fs.readFileSync(".node-version").toString().trim();
```
lujiajing1126, line 22: I suppose we can impose version check by setting the `engine` field in `package.json`

https://docs.npmjs.com/cli/v8/configuring-npm/package-json#engines

ui/.node-version
```diff
@@ -0,0 +1 @@
+16.15.0
```
lujiajing1126, line 1: I don't think `.node-version` is a common dotfile for node version selection.

Probably `.nvmrc` is a better dotfile which can be read by [`nvm`](https://github.com/nvm-sh/nvm#nvmrc).

And we may use `lts/gallium` as the content of `.nvmrc`.

lujiajing1126 (changes requested):

hanahmily:
> Why we have to use `npm` over `yarn`?

There are two motivations:

  * skywalking-eye opts to `npm` to generate license files.
  * UI contributors are using the element UI which prefers `npm`.

lujiajing1126:
> > Why we have to use `npm` over `yarn`?
> 
> There are two motivations:
> 
> * skywalking-eye opts to `npm` to generate license files.
> * UI contributors are using the element UI which prefers `npm`.

Make sense. But I'm still confused with the `.node-version` file.

It there any toolkit which follows this dotfile?

ui/tools/checkversion.js
hanahmily: `engine` is better to place the version. I updated the script

ui/.node-version
hanahmily: Replaced it with `package.json`

hanahmily:
> > > Why we have to use `npm` over `yarn`?
> > 
> > 
> > There are two motivations:
> > 
> > * skywalking-eye opts to `npm` to generate license files.
> > * UI contributors are using the element UI which prefers `npm`.
> 
> Make sense. But I'm still confused with the `.node-version` file.
> 
> It there any toolkit which follows this dotfile?

I removed this file.

lujiajing1126 (approved):

PR #136, opened 2022-06-18
Add streaming API and topN aggregator

lujiajing1126:
This PR tends to introduce a simple stream processing API and implementation for TopN aggregation.

## Design

`Flow` is an abstraction for the streaming processes, with following `operator`(s),

- `Source`: provide data stream for the `Flow`. As we've discussed before, it should be a listener consuming `Measure` write request continuously. Later, we could use a global binlog/WAL.
- `Mapper`: `func(T) R` which transforms an element from `T` to `R`
- `Filter`: `func(T) bool` which predicates whether an element should be passed to the downstream
- `Windows`: Currently only `SlidingEventTimeWindows` is implemented
- `Sink`: the place to write final result. We have to write the final result, e.g. TopN into a separate `Measure` storage.

### Filter

```go
s := flow.New(tt.input).
    Filter(func(i int) bool {
        return i%2 == 0
    }).
    To(snk)
```

The `Filter` operator allows us to filter by `criteria` set in `TopNAggregation`.

### Mapper

```go
s := flow.New(tt.input).
    Map(func(i int) int {
        return i * 2
    }).
    To(snk)
```

The `Mapper` operator allows us to extract `field` from the record and transform it by `groupBy` operation.

**We currently do not have a separate `keyBy` (for example `Apache/Flink`) operation to do `groupBy` for simplicity**

### Windows

Generally, the split of the window can be related to the "time", which could be either of the following concepts,

- Event Time
- Processing Time

image

The above graph from `Flink` community discriminates these concepts. But in our case, the only "time" we care about is `EventTime` which represents the exact moment the record is produced, since we need to use the `EventTime` to drive the timely flush of the aggregation results, e.g. TopN ranks.

image
Sliding windows can fulfill our requirement in the sense that we need to flush the data more frequently.

It means the flush interval should be much smaller than the interval of the real data points. For example in OAP, the downsampling rate can be `MINUTE` while the flush timer is set to 25 seconds by default.

Technically, the `SlidingEventWindow` is built on,

- A `PriorityQueue` maintains records which have not yet been emitted,
- A `PriorityQueueSet` maintains all registered (depulicated) timers which will be triggered later

### TopN

With the above semantics, we can impl `TopN` as a window aggregation function,

```go
flow.New(tt.input).
    Filter(...). // where
    Mapper(...) // select and groupBy 
    Window(NewSlidingTimeWindows(time.Minute*1, time.Second*15)).
    TopN(10, OrderBy(modelv1.SORT_DESC), ...) // TopN with parameters
    To(snk)
```

TopN is implemented with the help of a `TreeMap` which maps sortedKey to the collection of records.


hanahmily:
Thanks for your input. This streaming framework is so elegant. How about adopting it into:

* TopN aggregation(the original goal)
* Query module. We have a plan to support streaming-style query API to fix #8531 and other complicated querying scenarios. 

If these are the cases, we could have capacities to switch a query process between ad-hoc and precalculation according to its characters, which means a unified codebase to support two kinds of `TopN` is possible.

wu-sheng:
To implement the query in the streaming pattern, we need to make sure the flow instance should be created lightly in the runtime. 
Such as in the OAP streaming kernel, it is heavy as it is only performance first and immutable in the runtime(only to be set in the booting stage).

Here, you may face two cases, for persistent, the stream is performance first, and for query, in the streaming-style query(a.k.a. query condition driven streaming), would require the flexibility and lightweight.

The in-memory queue(including the time window you mentioned) and GC cost would be relative to this topic.

lujiajing1126:
Thanks both of you for the useful comments.

> Thanks for your input. This streaming framework is so elegant. How about adopting it into:
> 
> * TopN aggregation(the original goal)
> * Query module. We have a plan to support streaming-style query API to fix #8531 and other complicated querying scenarios.
> 
> If these are the cases, we could have capacities to switch a query process between ad-hoc and precalculation according to its characters, which means a unified codebase to support two kinds of `TopN` is possible.

Yes. This is exactly what I was thinking while I was designing and writing this submodule. But we may face the situation mentioned by @wu-sheng,

> To implement the query in the streaming pattern, we need to make sure the flow instance should be created lightly in the runtime. Such as in the OAP streaming kernel, it is heavy as it is only performance first and immutable in the runtime(only to be set in the booting stage).
> 
> Here, you may face two cases, for persistent, the stream is performance first, and for query, in the streaming-style query(a.k.a. query condition driven streaming), would require the flexibility and lightweight.
> 
> The in-memory queue(including the time window you mentioned) and GC cost would be relative to this topic.

Currently, the `flow` system (for OLAP) will start several go-routines, i.e. in-memory queue, for background jobs and timers. This is likely to hurt the performance of OLTP.

I will try to investigate the possibility and see to what extend we can reuse these components.

wu-sheng:
> Currently, the flow system (for OLAP) will start several go-routines, i.e. in-memory queue, for background jobs and timers. This is likely to hurt the performance of OLTP.

I think the most component impacting this is the queue. The queue usually increases the throughput capability by using the bulk mode, but it increases the latency and costs a lot of memory/lock/flag/signal.

I would like to recommend to establish the flow with different APIs to adopt read(more flexible) and write(higher performance and immutable). In the query, more things should be done by concurrency with the final result report. So, once the sliding window and most in-memory queue are limited in write scenario, we should be safer.

And also, two API sets are better for dynamic and immutable read/write scenarios.

lujiajing1126:
A simple (incomplete) batch flow API is added. 

Since we've changed several time for the query module in the development of `v0.1.0` to support some extra features like `groupBy`, `topN`, it now looks like patchwork. The new batchAPI implementation is fully based on Iterator, i.e. `Iterator[T]` which takes advantage of generics introduced in go `1.18`,

```go
snk := sink.NewItemSlice()
flow := New(sources.NewMeasure(m, timestamp.NewInclusiveTimeRangeDuration(baseTs, 10*time.Minute)), metaSvc).
	Filter([]*modelv1.Criteria{pb.And("default", pb.Eq("id", pb.TagTypeID("1")))}).
	To(snk)

err := flow.OpenSync()
```

Now, `streaming` and `batch` have similar APIs,

```go
type Flow interface {
	Filter(f interface{}) Flow
	Map(f interface{}) Flow
	Window(WindowAssigner) WindowedFlow
	Offset(int) Flow
	Limit(int) Flow
	To(interface{}) Flow
	OpenAsync() <-chan error
	OpenSync() error
}
```

In batch API,

- `Source` contains `Measure` or `Stream` and time-range,
- `Filter` handles both sharding keys and indexed tag query. And it is designed to resolve https://github.com/apache/skywalking/issues/8513, the underlying operators already support `OR`. In the next PR, we could introduce API changes for `OR` criteria.

wu-sheng:
So, could we say this is a streaming writing API skeleton only. Not relate to read for now, right?

lujiajing1126:
> So, could we say this is a streaming writing API skeleton only. Not relate to read for now, right?

Batch Read API is introduced in the last commit, but it is still incomplete now. Only filter (with time-range, indexed tags and sharding keys) is implemented. Just an attempt to unify streaming write and query.

I think we can focus on the streaming write part in this PR and then the query part in the next one.

wu-sheng:
Make sense.

lujiajing1126:
Generally ready for review. I will implement query soon.

But still some questions to be discussed @hanahmily ,

### How to control the window parameter for the sliding window

We have two parameters,

1. Window Size <=> real interval of the data point
2. Slide Size <=> flush interval and the size of the batch

### How to shard and serialize topN results

Currently, I use a different `Entity` prefix for `topN`, i.e. `${measure.name}.${topNAggregation.name}`. And this is **fixed** which means we do not set any sharding key for the record and thus the actual shard is now hardcoded to `0`.

And the ranking result is now a slice of `Tuple2` struct in go,

```go
type Tuple2 {
  First interface{}
  Second interface{}
}
```

Overall, according to the current data types we have for the `Tag`, `[]Tuple2` can be serialized to `Binary`. But how to serialize each of the `Tuple2`, maybe `json`? This choice would determine the deserialization in the query.

### How to abstract the storage for bound schemas, like `TopNAggregation`

The bound schema means this schema is somehow relevant to a `Measure` or `Stream`, for example, the `TopNAggregation` must have a `Measure` as its source. And since the results of `TopN` have to be persistent, we can always transform the `TopN` to a "virtual" `Measure`.

Currently, I only introduced an extra method in the `Measure` interface for query the bound schema like `TopN`,

`CompanionShards(metadata *commonv1.Metadata) ([]tsdb.Shard, error)`

the signature implies we do not have sharding keys for bound schema.

pkg/flow/api/ds.go
```diff
@@ -0,0 +1,24 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package api
+
+type Data []any
+
+func Row(args ...any) Data {
```
hanahmily, line 22: Could you instance Data with `Data{...}` instead of this function.

pkg/flow/api/types.go
```diff
@@ -0,0 +1,98 @@
...
+// under the License.
+
+package api
+
+import (
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+type Flow interface {
+	Filter(f interface{}) Flow
```
hanahmily, line 25: Is `func(interface{}) bool` better for `f`?

pkg/flow/api/types.go
```diff
@@ -0,0 +1,98 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+package api
+
+import (
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+type Flow interface {
```
hanahmily, line 24: Could you add comments to all exported types? Now we have this checker in bydbctl, I would introduce it to other packages soon.

pkg/flow/api/types.go
```diff
@@ -0,0 +1,98 @@
...
+	Filter(f interface{}) Flow
+	Map(f interface{}) Flow
```
hanahmily, line 26: how about opting to `func (i interface{}) interface{}` instead `interface`?

pkg/flow/api/types.go
```diff
@@ -0,0 +1,98 @@
...
+	TopN(topNum int, opts ...any) Flow
+	Aggregate(aggrFunc AggregateFunction) Flow
+}
+
+type Window interface {
+	MaxTimestamp() int64
+}
+
+type WindowAssigner interface {
+	AssignWindows(int64) ([]Window, error)
```
hanahmily, line 45: Is the parameter a time range?

pkg/flow/api/types.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+//go:generate mockgen -destination=./aggregation_func_mock.go -package=api github.com/apache/skywalking-banyandb/pkg/flow/api AggregateFunction
+type AggregateFunction interface {
```
hanahmily, line 49: From current implements, it is can be simplified by `type AggregateFunction func([]interface{})interface{}`.

pkg/flow/api/types.go
```diff
@@ -0,0 +1,98 @@
...
+	Add([]interface{})
+	GetResult() interface{}
+}
+
+type StreamRecord struct {
+	ts           int64
+	hasTimestamp bool
+	data         interface{}
```
hanahmily, line 57: You could leverage type parameters to implement this struct.

```
type StreamRecord[T any] struct {
    ....
    data T
}
```

pkg/flow/api/types.go
```diff
@@ -0,0 +1,98 @@
...
+
+//go:generate mockgen -destination=./aggregation_func_mock.go -package=api github.com/apache/skywalking-banyandb/pkg/flow/api AggregateFunction
+type AggregateFunction interface {
+	Add([]interface{})
+	GetResult() interface{}
+}
+
+type StreamRecord struct {
+	ts           int64
+	hasTimestamp bool
```
hanahmily, line 56: I didn't find any references from external packages? Is it redundant?

hanahmily (changes requested):
Reviewed the `flow/api` packages.

pkg/flow/api/ds.go
lujiajing1126: Fixed

pkg/flow/api/types.go
lujiajing1126: Comments added

pkg/flow/api/types.go
lujiajing1126: Comment added.

The parameter is the timestamp, i.e. eventTime in streaming

pkg/flow/api/types.go
lujiajing1126: This is extensively used in the tests.

Shall we remove this? For example, we assign a default timestamp.

pkg/flow/api/types.go
lujiajing1126: I don't think it is easy to take advantage of generic here.

pkg/flow/api/types.go
lujiajing1126: I would explain "any" type here.

There are mainly two reasons,

1. With `any`, we can use two flavors of methods with or without `context.Context`: `func (i interface{}) interface{}` and `func (ctx context.Context, i interface{}) interface{}`
2. With `any`, we can unify the batch query and streaming processing,

```go
func (flow *batchFlow) Filter(f interface{}) api.Flow {
	if unresolvedCriteria, ok := f.([]*modelv1.Criteria); ok {
		flow.conditionOp = &conditionalFilterOperator{
			f:                  flow,
			unresolvedCriteria: unresolvedCriteria,
		}
	} else {
		flow.appendErr(errors.New("Filter can only accept []*modelv1.Criteria"))
	}
	return flow
}
```

The same reason for `Filter`.

pkg/flow/api/types.go
lujiajing1126: Refactored

banyand/measure/measure_topn.go
```diff
@@ -0,0 +1,568 @@
...
+	for rankNum, tuple := range tuples {
+		fieldValue := tuple.V1.(int64)
+		data := tuple.V2.(api.Data)
+		err = multierr.Append(err, t.writeData(eventTime, timeBucket, fieldValue, data, rankNum))
+	}
+	return err
+}
+
+func (t *topNStreamingProcessor) writeData(eventTime time.Time, timeBucket string, fieldValue int64, data api.Data, rankNum int) error {
+	// HACK: use rankNumber as the nanoseconds to avoid overwrite items with same entity
```
lujiajing1126, line 140: To avoid data overwrite, I currently add rankNumber as nanoseconds. For example, we want to save a snapshot of the TopN ranking at a given moment,

| Rank | GroupValue | FieldValue |
| ----- | ----------- | ---- |
| 1  | `Entity_1` |  10 |
| 2 | `Entity_2` | 5 |
| 3 | `Entity_1` |  2 |

Since we use `GroupValue`(s) as the entity, if we store (1) and (3) using the same eventTime, only the one saved later can be read in the query.

An alternative would be add rankNumber as a part of the `entity` but probably cause worse compression ratio? @hanahmily any suggestion?

banyand/measure/measure_topn.go
```diff
@@ -0,0 +1,568 @@
...
+			},
+			op:  cond.GetOp(),
+			val: v.Id.GetValue(),
+		}
+	default:
+		return nil
+	}
+}
+
+func (manager *topNProcessorManager) buildMapper(fieldName string, groupByNames ...string) (measureProcessorMapper, error) {
```
lujiajing1126, line 391: Currently, `groupByNames` is required. Shall we consider a pre-`TopN Aggregation` without groupBy?

This would affect the layout we save for the TopN "virtual" measure.

pkg/flow/api/types.go
hanahmily: > With any, we can use two flavors of methods with or without context.Context: func (i interface{}) interface{} and func (ctx context.Context, i interface{}) interface{}

I want to pass the context parameter all the time. If you noticed that, more and more libs and projects are introducing the context as a mandatory argument.

> With any, we can unify the batch query and streaming processing

`Filter`, `Map`, and `Reduce` are streaming standard functions. Their parameter should be an "action". Passing the `criteria` would confuse the API's users. 

You could use the function to pass `modelv1.Criteria`. If you have performance concerns, the compiler is smart, optimizing this function.

pkg/flow/api/types.go
hanahmily: Would you elaborate on it? this StreamRecoard is a container that conforms to the rules of generic practices.

pkg/flow/api/types.go
hanahmily: I prefer to remove it. The test cases should leverage exported types to verify the behaviors.

banyand/measure/measure_topn.go
hanahmily: Could you map data into a single group in case of an missing group? If that, you don't have to refactor too many points.

banyand/measure/measure_topn.go
lujiajing1126: Now the rankNumber is used as a part of the `Entity` for consistency.

pkg/flow/api/types.go
lujiajing1126: Removed

pkg/flow/api/types.go
lujiajing1126: Since Golang does not support generic covariant/contravariant, which means we cannot convert interface{} with underlying type `StreamRecord[int]` to `StreamRecord[any]`, thus it is not possible to represent the concrete type in the middle of the processing, like in `SlidingWindow`,

```go
func (s *SlidingTimeWindows) receive() {
	for elem := range s.in {
		record := elem.(api.StreamRecord)
// not possible to do something like
// record := elem.(api.StreamRecord[any])
...
```

banyand/measure/measure_topn.go
lujiajing1126: > Could you map data into a single group in case of an missing group? If that, you don't have to refactor too many points.

To be clear, you meant mapping all data into a single group if groupByNames is not specified, right?

pkg/flow/api/types.go
lujiajing1126: > > With any, we can use two flavors of methods with or without context.Context: func (i interface{}) interface{} and func (ctx context.Context, i interface{}) interface{}
> 
> I want to pass the context parameter all the time. If you noticed that, more and more libs and projects are introducing the context as a mandatory argument.
> 
> > With any, we can unify the batch query and streaming processing
> 
> `Filter`, `Map`, and `Reduce` are streaming standard functions. Their parameter should be an "action". Passing the `criteria` would confuse the API's users.
> 
> You could use the function to pass `modelv1.Criteria`. If you have performance concerns, the compiler is smart, optimizing this function.

I'm having a try. The first is easy to be implemented. But the problem occurs within the query context.

Currently, `[]*modelv1.Criteria` is used to build the `tsdb.ItemFilter`. With the parameter type `func(context.Context, any) bool`, it is impossible to return `[]*modelv1.Criteria`

lujiajing1126: Found an ugly way out. Anyway, we may improve in the future.

hanahmily: If that, a new method is better here. Why could you insist reuse `Filter`?

pkg/flow/api/types.go
hanahmily: Make sense to me. Thanks for your clarification.

pkg/flow/api/types.go
lujiajing1126: Batch API has been removed

hanahmily (changes requested):
Reivewed streaming/flow.go

pkg/flow/types.go
```diff
@@ -0,0 +1,97 @@
...
+	Filter(UnaryOperation[bool]) Flow
+	// Map is used to transform data
+	Map(UnaryOperation[any]) Flow
+	// Window is used to split infinite data into "buckets" of finite size.
+	// Currently, it is only applicable to streaming context.
+	Window(WindowAssigner) WindowedFlow
+	// To pipes data to the given sink
+	To(sink interface{}) Flow
+	// OpenAsync opens the flow in the async mode for streaming scenario.
+	OpenAsync() <-chan error
```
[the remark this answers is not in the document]
lujiajing1126: Renamed

pkg/flow/types.go
```diff
@@ -0,0 +1,97 @@
...
+	// The parameter f can be either predicate function for streaming,
+	// or conditions for batch query.
+	Filter(UnaryOperation[bool]) Flow
+	// Map is used to transform data
+	Map(UnaryOperation[any]) Flow
+	// Window is used to split infinite data into "buckets" of finite size.
+	// Currently, it is only applicable to streaming context.
+	Window(WindowAssigner) WindowedFlow
+	// To pipes data to the given sink
+	To(sink interface{}) Flow
```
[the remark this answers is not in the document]
lujiajing1126: Moved

pkg/flow/types.go
```diff
@@ -0,0 +1,97 @@
...
+	// OpenAsync opens the flow in the async mode for streaming scenario.
+	OpenAsync() <-chan error
+}
+
+type WindowedFlow interface {
```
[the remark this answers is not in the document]
lujiajing1126: Added

pkg/flow/types.go
```diff
@@ -0,0 +1,97 @@
...
+	TopN(topNum int, opts ...any) Flow
+	Aggregate(aggrFunc AggregateFunction) Flow
+}
+
+type Window interface {
```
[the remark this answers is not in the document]
lujiajing1126: Added

pkg/flow/types.go
```diff
@@ -0,0 +1,97 @@
...
+	MaxTimestamp() int64
+}
+
+type WindowAssigner interface {
+	// AssignWindows assigns a slice of Window according to the given timestamp, e.g. eventTime.
+	AssignWindows(timestamp int64) ([]Window, error)
+}
+
+type AggregateFunction func([]interface{}) interface{}
```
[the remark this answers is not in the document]
lujiajing1126: Added

pkg/flow/types.go
```diff
@@ -0,0 +1,97 @@
...
+
+type StreamRecord struct {
```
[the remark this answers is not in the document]
lujiajing1126: Added

pkg/flow/streaming/flow.go
```diff
@@ -0,0 +1,175 @@
...
+	streamingApi "github.com/apache/skywalking-banyandb/pkg/flow/streaming/api"
+	"github.com/apache/skywalking-banyandb/pkg/flow/streaming/sources"
+)
+
+var _ flow.Flow = (*streamingFlow)(nil)
+
+type streamingFlow struct {
+	// sourceParam and sinkParam are generic source and sink
+	sourceParam interface{}
+	sinkParam   interface{}
```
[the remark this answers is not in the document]
lujiajing1126: Removed

pkg/flow/streaming/flow.go
```diff
@@ -0,0 +1,175 @@
...
+	// check and build sink type
+	if err := f.buildSink(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (f *streamingFlow) buildSource() error {
+	if f.sourceParam == nil {
```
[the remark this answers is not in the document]
lujiajing1126: Not used anymore

pkg/flow/streaming/flow.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	// check primitive
+	srcType := reflect.TypeOf(f.sourceParam)
+	switch srcType.Kind() {
+	case reflect.Slice:
+		f.source = sources.NewSlice(f.sourceParam)
+	case reflect.Chan:
+		f.source = sources.NewChannel(f.sourceParam)
+	}
```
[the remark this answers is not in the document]
lujiajing1126: Refactored

pkg/flow/types.go
```diff
@@ -0,0 +1,97 @@
...
+type WindowedFlow interface {
+	TopN(topNum int, opts ...any) Flow
+	Aggregate(aggrFunc AggregateFunction) Flow
+}
+
+type Window interface {
+	MaxTimestamp() int64
+}
+
+type WindowAssigner interface {
```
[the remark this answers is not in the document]
lujiajing1126: Added

pkg/flow/streaming/sources/reader.go
```diff
@@ -0,0 +1,100 @@
...
+	"context"
+	"io"
+
+	"github.com/apache/skywalking-banyandb/pkg/flow"
+	streamingApi "github.com/apache/skywalking-banyandb/pkg/flow/streaming/api"
+)
+
+var _ streamingApi.Source = (*sourceReader)(nil)
+
+type sourceReader struct {
```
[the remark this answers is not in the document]
lujiajing1126: Totally removed

pkg/flow/streaming/sources/slice.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/pkg/errors"
+
+	"github.com/apache/skywalking-banyandb/pkg/flow"
+	streamingApi "github.com/apache/skywalking-banyandb/pkg/flow/streaming/api"
+)
+
+var _ streamingApi.Source = (*sourceSlice)(nil)
+
+type sourceSlice struct {
```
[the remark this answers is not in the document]
lujiajing1126: Moved to `pkg/test`

pkg/flow/priority_queue_set.go
```diff
@@ -0,0 +1,136 @@
+// Licensed to Apache Software Foundation (ASF) under one or more contributor
```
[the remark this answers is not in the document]
lujiajing1126: No. Normally the priority queue is based on slices like the example from the official go doc.

But we need deduplication for this data structure, so `_set` means deduplication here.

pkg/flow/streaming/flow.go
```diff
@@ -0,0 +1,175 @@
...
+type streamingFlow struct {
+	// sourceParam and sinkParam are generic source and sink
+	sourceParam interface{}
+	sinkParam   interface{}
+
+	ctx    context.Context
+	source streamingApi.Source
+	sink   streamingApi.Sink
+	ops    []streamingApi.Operator
+	drain  chan error
```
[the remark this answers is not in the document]
lujiajing1126: Test cases have been added to verify this error chan while a single goroutine is started to consume this channel to avoid too many waiting goroutines.

`TearDown` is also used to ensure resource recycling during the shutdown.

lujiajing1126:
> Reivewed streaming/flow.go

All suggestion applied

pkg/flow/priority_queue_set.go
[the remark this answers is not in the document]
hanahmily: You named the type `PriorityQueue`  which is not the file's name. 

The `queue` and `set` share an identical notion, `container`. We don't have to combine them. How about naming it `UniquePriorityQueue`?

pkg/flow/streaming/flow.go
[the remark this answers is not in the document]
hanahmily: Where to close this chan?

pkg/flow/priority_queue_set.go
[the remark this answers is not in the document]
lujiajing1126: > You named the type `PriorityQueue` which is not the file's name.
> 
> The `queue` and `set` share an identical notion, `container`. We don't have to combine them. How about naming it `UniquePriorityQueue`?

Renamed to `DedupPriorityQueue`

pkg/flow/streaming/flow_sliding_window.go
```diff
@@ -0,0 +1,342 @@
...
+func (s *SlidingTimeWindows) receive() {
+	s.Add(1)
+	defer s.Done()
+	for elem := range s.in {
+		record := elem.(flow.StreamRecord)
+		// assume records are consumed one by one in strict time order
+		s.currentWatermark = record.TimestampMillis()
+		s.purgeOutdatedWindows()
+
+		assignedWindows, err := s.AssignWindows(record.TimestampMillis())
```
hanahmily, line 193: It's tricky here. Why does it purge a window repeatly? The very beginning window in a interval gets more than one purging.

hanahmily (commented):
the flow_window is a bit in the weeds for me. Could you simplify it a bit?

pkg/flow/types.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+func (sr StreamRecord) Data() interface{} {
+	return sr.data
+}
+
+//go:generate mockgen -destination=./inlet_mock.go -package=flow github.com/apache/skywalking-banyandb/pkg/flow Inlet
+// Inlet represents a type that exposes one open input.
+type Inlet interface {
+	In() chan<- interface{}
```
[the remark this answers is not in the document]
lujiajing1126: Fixed

pkg/flow/types.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// Outlet represents a type that exposes one open output.
+type Outlet interface {
+	Out() <-chan interface{}
```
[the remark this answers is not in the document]
lujiajing1126: Fixed

pkg/flow/streaming/flow_unary.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func (f *streamingFlow) Map(mapper flow.UnaryOperation[any]) flow.Flow {
+	return f.Transform(mapper)
+}
+
+// Transform represents a general unary transformation
+// For example: filter, map, etc.
+func (f *streamingFlow) Transform(op flow.UnaryOperation[any]) flow.Flow {
+	f.ops = append(f.ops, newUnaryOp(op, 1))
```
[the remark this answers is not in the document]
lujiajing1126: Added

pkg/flow/streaming/flow_sliding_window.go
```diff
@@ -0,0 +1,342 @@
...
+		return windows, nil
+	}
+	return nil, errors.New("invalid timestamp from the element")
+}
+
+// getWindowStart calculates the window start for a timestamp.
+func getWindowStart(timestamp, windowSize int64) int64 {
+	remainder := timestamp % windowSize
+	// handle both positive and negative cases
+	if remainder < 0 {
```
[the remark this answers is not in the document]
lujiajing1126: unnecessary check removed

pkg/flow/streaming/flow_sliding_window.go
```diff
@@ -0,0 +1,342 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/apache/skywalking-banyandb/pkg/flow"
+)
+
+type TriggerResult uint8
+
+const (
+	FIRE TriggerResult = 1 << iota
+	CONTINUE
```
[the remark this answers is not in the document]
lujiajing1126: Changed

pkg/flow/streaming/flow_sliding_window.go
```diff
@@ -0,0 +1,342 @@
...
+		return
+	}
+}
+
+func (s *SlidingTimeWindows) receive() {
+	s.Add(1)
+	defer s.Done()
+	for elem := range s.in {
+		record := elem.(flow.StreamRecord)
+		// assume records are consumed one by one in strict time order
```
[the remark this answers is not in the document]
lujiajing1126: More comments added. Watermark update logic has changed.

pkg/flow/streaming/flow_sliding_window.go
```diff
@@ -0,0 +1,342 @@
...
+	ctx.RegisterEventTimeTimer(window.MaxTimestamp())
+	return CONTINUE
+}
+
+func (t EventTimeTrigger) OnEventTime(time int64, window timeWindow, ctx TriggerContext) TriggerResult {
+	if time == window.MaxTimestamp() {
+		return FIRE
+	}
+	return CONTINUE
+}
```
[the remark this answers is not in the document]
lujiajing1126: Fixed

pkg/flow/streaming/flow_sliding_window.go
```diff
@@ -0,0 +1,342 @@
...
+	default:
+		s.f.drainErr(errors.New("aggregation is not supported"))
+	}
+	return s.f
+}
+
+type SlidingTimeWindows struct {
+	flow.ComponentState
+	// For SlidingTimeWindows
+	currentWindow timeWindow
```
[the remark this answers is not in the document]
lujiajing1126: Removed

pkg/flow/streaming/flow_sliding_window.go
```diff
@@ -0,0 +1,342 @@
...
+		slidingSlices := s.queue.Slice(windowBottomIndex, slideUpperIndex)
+		// if we've collected some items, reallocate queue
+		if len(slidingSlices) > 0 {
+			remainingItems := s.queue.Slice(slideUpperIndex, windowUpperIndex)
+			// reset the queue
+			var err error
+			s.queue, err = s.queue.WithNewItems(remainingItems)
+			if err != nil {
+				// TODO: drain error
+				panic("drain error")
```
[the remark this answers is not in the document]
lujiajing1126: Refactored with a `errorHandler`

pkg/flow/streaming/flow.go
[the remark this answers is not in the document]
lujiajing1126: `drainErr` function has been switched to sync mode, which means error channel must be properly handled after `Open`.

So we can safely close `errorChan` after all components are closed.

lujiajing1126:
In order to mitigate the data race, I have to use a RWLock in the core path of streaming processor. To further avoid this performance hit, I would suggest changing the `Subscriber` contract,

```
type Subscriber interface {
	Subscribe(topic Topic, listener MessageListener) (Subscription, error)
}

type Subscription interface {
    Cancel()
}
```

`Cancel()` should a sync method, which await after the worker goroutine safely returns. Then the subscriber can ensure there is no incoming messages and thus can safely shutdown the subscriber.

lujiajing1126:
@hanahmily Could you please take a look at this CI failure? I cannot reproduce it on my laptop

hanahmily:
> @hanahmily Could you please take a look at this CI failure? I cannot reproduce it on my laptop

It seems the data race checking failure of flow

lujiajing1126:
> > @hanahmily Could you please take a look at this CI failure? I cannot reproduce it on my laptop
> 
> It seems the data race checking failure of flow

It seems to be unstable. The data race is not reported every time.

hanahmily:
> In order to mitigate the data race, I have to introduce a `sync.RWLock` in the core path of streaming processor. To further avoid this performance hit, I would suggest changing the `Subscriber` contract,
> 
> ```
> type Subscriber interface {
> 	Subscribe(topic Topic, listener MessageListener) (Subscription, error)
> }
> 
> type Subscription interface {
>     Cancel()
> }
> ```
> 
> `Cancel()` should be a sync method, which awaits until the worker goroutine safely returns. Then the subscriber can make sure that there is no incoming messages and thus can safely shutdown the subscriber.

Do you regard `flow/streaming/sink/slice.go`? It seems to be a testing helper which can't affect the main process of the flow. BTW, it should be in a testing package.

hanahmily:
> > > @hanahmily Could you please take a look at this CI failure? I cannot reproduce it on my laptop
> > 
> > 
> > It seems the data race checking failure of flow
> 
> It seems to be unstable. The data race is not reported every time.

yep. data race checking is flaky all the time. Try to use ` ginkgo run --race --until-it-fails` to find the root.

pkg/flow/streaming/flow_sliding_window.go
```diff
@@ -0,0 +1,356 @@
...
+			}
+		}
+		item := &TimestampedValue{elem, 0}
+		s.queueMu.Lock()
+		heap.Push(s.queue, item)
+		s.queueMu.Unlock()
+	}
+	close(s.purgedWindow)
+	close(s.done)
+	close(s.out)
```
hanahmily, line 216: You don't have to close this chan since it have a single sender.

pkg/flow/streaming/flow_sliding_window.go
```diff
@@ -0,0 +1,356 @@
...
+	go s.receive()
+	// start emitting
+	s.Add(1)
+	go s.emit()
+
+	return nil
+}
+
+func (s *SlidingTimeWindows) emit() {
+	defer s.Done()
```
hanahmily, line 117: You can close `out` here to avoid the data race.

lujiajing1126:
After discussion on the weekly meeting, to reduce the memory footprint, a new design has been introduced to replace the previous one,

1. a single TopN aggregator is created and maintained for each TimeWindow,
2. Tumbling window is used instead of the old sliding window

The reason why I introduced sliding window is to cache batches of records and flush them by a smaller time interval (compared with the DataPoint interval, e.g. 1 min). But now it is not necessary in the sense that we can cache by TopN aggregator and merge them to the so-called accumulator.

The TopN results can be then extracted from the accumulator. The memory usage is now controlled by the following parameters,

1. Cache entry number of the TopN aggregator,
2. Lateness

The max total memory consumption will be linearly varying with the cache entry number * aggregator number, of which the latter is determined by `lateness`

hanahmily:
@lujiajing1126 Since we redesigned the process and this PR conflicted with the main branch, how about closing this and raising another PR?

PR #137, opened 2022-06-21
Fix several flaws of the toolchain

hanahmily:
* Improve compatibility with non-Unix OS by a go-based script
* Add gofumpt to format go sources
* Drop formatting generated sources
* Add package-lock.json to firm dependencies


hanahmily:
> Seems the newly generated `*.pb.go` files do not follow the three-part rules,
> 
> 1. stdlib
> 2. third-party libs
> 3. self references

I ignored them on purpose. The generated sources belong to `skip-files`: https://github.com/apache/skywalking-banyandb/pull/137/files#diff-b7d9dafa65300010103ed9e65e449f0399ebac7c9aef38130d49983259ab366aR18-R22

We don't have to edit or polish them since they declare `DO NOT EDIT.` in their headers.

lujiajing1126 (approved):
LGTM

PR #140, opened 2022-06-29
Install golangci-lint by go install

hanahmily:
Install golangci-lint through `go install` to tackle the issue that `wget` might fail silently. 

@WuChuSheng1 Please test this patch on your local VM. 


lujiajing1126:
I have issue with `make pre-push`. This command does not actually run `lint` check

hanahmily:
> I have issue with `make pre-push`. This command does not actually run `lint` check

fixed it.

lujiajing1126 (approved):

PR #148, opened 2022-07-29
Remove term metadata store

hanahmily:
Fixes https://github.com/apache/skywalking/issues/9405

This change frees up ~20% of memory. 


lujiajing1126 (approved):
LGTM. A question: why do we intend to initially convert raw value bytes to a hash id?

hanahmily:
For a reduced term dictionary key. The raw key length could be more than 8
bytes  which is the fixed length of a ID

PR #153, opened 2022-08-09
Parameterize memory size and some key improvements

hanahmily:
These changes introduce several parameters to control memory usage. I assigned some defaults to them, and now the initial memory is around 2GB 🎉 

There are a series of improvements:

* Remove `run.Chan`, which could cause the data race. We don't have to close a channel more than notifying the receivers. @lujiajing1126 Please don't try to use it. 
* Introduce a self-spinning lock to optimize the life-cycle of the block.
* Fix a tiny bug about parsing the lsm index's key. You never encounter errors in testing the `measure` package. 


banyand/kv/kv.go
```diff
@@ -126,6 +126,17 @@ func TSSWithFlushCallback(callback func()) TimeSeriesOptions {
 	}
 }
 
+func TSSWithMemTableSize(size int64) TimeSeriesOptions {
+	return func(store TimeSeriesStore) {
+		if size < 1 {
```
lujiajing1126, line 131: Do we have some minimal requirement for the mem table?

banyand/kv/kv.go
```diff
@@ -160,6 +171,9 @@ func OpenTimeSeriesStore(shardID int, path string, options ...TimeSeriesOptions)
 	}
 	// Put all values into LSM
 	btss.dbOpts = btss.dbOpts.WithVLogPercentile(1.0)
+	if btss.dbOpts.MemTableSize < 8<<20 {
```
lujiajing1126, line 174: `8<<20` seems to be a magic number... What does it mean?

lujiajing1126 (changes requested):
We should make default values constants or variables and add comments for them rather than directly use them everywhere.

Instead of `2<<20` or `8<<20`, I prefer `1<<21` or `1<<24` which seems to be clearer.

banyand/kv/kv.go
hanahmily: No, we don't limit any values to more than 1 bit.

lujiajing1126:
> * Remove `run.Chan`, which could cause the data race. We don't have to close a channel more than notifying the receivers. @lujiajing1126 Please don't try to use it.

Not quite follow this argument. I remember I'm not using this toolkit.

banyand/kv/kv.go
hanahmily: 8MB

hanahmily:
> We should make default values constants or variables and add comments for them rather than directly use them everywhere.

Make sense, I will tweak them. 

> Instead of `2<<20` or `8<<20`, I prefer `1<<21` or `1<<24` which seems to be clearer.

`8<<20` indicates the literal(8) and the unit(1<<20=1024*1024=1MB), which is more precise and straightforward than `1<<24`.

banyand/tsdb/metric.go
```diff
@@ -66,11 +66,11 @@ func (s *shard) runStat() {
 }
 
 func (s *shard) stat() {
-	defer func() {
-		if r := recover(); r != nil {
-			s.l.Warn().Interface("r", r).Msg("recovered")
-		}
-	}()
+	// defer func() {
```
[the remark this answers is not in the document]
hanahmily: No. comment them by mistake.

hanahmily:
Fixes https://github.com/apache/skywalking/issues/9383

banyand/tsdb/index/writer.go
```diff
@@ -76,36 +77,37 @@ func NewWriter(ctx context.Context, options WriterOptions) *Writer {
...
 
 func (s *Writer) Write(value Message) {
 	go func(m Message) {
-		s.ch.Write(m)
+		s.ch <- m
 	}(value)
 }
 
 func (s *Writer) Close() error {
-	return s.ch.Close()
```
lujiajing1126, line 92: Since we removed `Close()` here, is it possible we still have many goroutines waiting to write to the channel?

hanahmily: We can use close to notify the receivers, not the senders. The senders will be waiting to write. That is the purpose of this process.

wu-sheng (approved):

banyand/tsdb/index/writer.go
lujiajing1126: > We can use close to notify the receivers, not the senders. The senders will be waiting to write. That is the purpose of this process.

Yes, we can use some stop channel to notify the receivers, but what about the senders?

> They won't block the GC from collecting them.

From this issue, https://github.com/golang/go/issues/19702, Go may not be clever enough 
 now to collect blocking goroutines.

lujiajing1126 (approved):
LGTM. We can create an issue to track goroutine leak issue later.

banyand/tsdb/index/writer.go
wu-sheng: Automatically closing always has this kind of bugs in GC phases.
Be careful. We don't want a leak 😄

PR #155, opened 2022-08-16
[Draft]Tweak query APIs

hanahmily:
* Add logical operation to criteria to support OR
* Add full-text searching


hanahmily:
@wu-sheng @lujiajing1126 @wankai123 Please take a look at the primitive updates

api/proto/banyandb/model/v1/query.proto
```diff
@@ -68,8 +72,21 @@ message Condition {
 
 // tag_families are indexed.
 message Criteria {
-    string tag_family_name = 1;
-    repeated model.v1.Condition conditions = 2;
+    repeated Expression exps = 1;
```
lujiajing1126, line 75: What does the repeated imply here? And or OR?

lujiajing1126 (changes requested):

api/proto/banyandb/database/v1/schema.proto
```diff
@@ -158,6 +158,15 @@ message IndexRule {
     Location location = 4;
     // updated_at indicates when the IndexRule is updated
     google.protobuf.Timestamp updated_at = 5;
+    enum Analyzer {
+        ANALYZER_UNSPECIFIED = 0;
+        ANALYZER_KEYWORD = 1;
+        ANALYZER_STANDARD = 2;
+        ANAKYZER_SIMPLE = 3;
+        ANAKYZER_STOP = 4;
```
wu-sheng, line 166: Could you specify these four types?

api/proto/banyandb/model/v1/query.proto
wu-sheng: I want to recheck here too. If Expression is planned to support nested, then Expression should be a hierarchy structure, rather than a `repeated`. This seems an intermediate status

hanahmily: Add inner expression here

api/proto/banyandb/database/v1/schema.proto
hanahmily: Add comments

api/proto/banyandb/model/v1/query.proto
wu-sheng: Once the nested expression mechanism is added, we don't need repeatable exps here. Eventually, we need one exp at the top level.

wu-sheng: @hanahmily I think this has not been 100% resolved, the `repeated` is still arguable.

hanahmily: Push down "repeated" fields, that could support nested operations.

api/proto/banyandb/model/v1/query.proto
```diff
@@ -68,8 +72,23 @@ message Condition {
...
+        LOGICAL_OP_AND = 1;
+        LOGICAL_OP_OR = 2; 
+    }
+    // op is a logial operation
+    LogicalOp op = 1;
+    // tag_family_name indicates the tag family to which the tag name in the condition belongs.
+    string tag_family_name = 2;
+    // condition is a binary operation
+    model.v1.Condition condition = 3; 
 }
```
wu-sheng, line 92: In our discussion, as you want to separate logic expression and op condition, I want to propose this one,

```proto

// tag_families are indexed.
message Criteria {
   enum LogicalOp {
        LOGICAL_OP_UNSPECIFIED = 0;
        LOGICAL_OP_AND = 1;
        LOGICAL_OP_OR = 2; 
    }
    // op is a logial operation
   LogicalOp op = 1;
   repeated LogicalExpression les = 2;
}

// LogicalExpression supports logical operation
message LogicalExpression {
    value oneof {
        ConditionExpression = 1;
        Criteria = 2;
    }
}

message ConditionExpression {
    // tag_family_name indicates the tag family to which the tag name in the condition belongs.
    string tag_family_name = 1;
    // condition is a binary operation
    model.v1.Condition condition = 2; 
}
```

hanahmily: Thanks for pointing out the missing operator between inner and outer. I will polish the API a bit.

wu-sheng (approved):
The new query schema makes sense to me.

lujiajing1126 (approved):
API LGTM

hanahmily:
Great. We could implement follow up feats based on these changes.

hanahmily:
We will submit other PRs based on this.

PR #158, opened 2022-08-24
Fix data race

hanahmily:


wu-sheng (approved):

lujiajing1126 (approved):
LGTM. But how could the data race happen? Block closed before goroutine start?

PR #163, opened 2022-08-25
Bump vue

hanahmily:


lujiajing1126 (changes requested):
This version upgrade breaks the `vue-cli-service serve`

image

hanahmily:
Let's keep it as 4.x since there are a lot of changes to bump up to 5.x

PR #164, opened 2022-08-25
Add streaming API and topN aggregator

lujiajing1126:
As a continuation of PR #136


pkg/flow/streaming/flow_sliding_window.go
```diff
@@ -0,0 +1,334 @@
...
+	// For TumblingTimeWindows
+	// Unit: Millisecond
+	size int64
+	// lateness is the maximum allowed lateness for incoming elements
+	// Unit: Millisecond
+	lateness int64
+
+	// guard snapshots
+	snapshotsMu sync.Mutex
+	snapshots   map[int64]flow.AggregationOp
```
hanahmily, line 86: This map will cause OOM due to the lack of releasing

pkg/flow/streaming/flow_sliding_window.go
```diff
@@ -0,0 +1,334 @@
...
+
+	return nil
+}
+
+func (s *TumblingTimeWindows) emit() {
+	defer s.Done()
+	for w := range s.purgedWindow {
+		if err := func(window timeWindow) error {
+			s.snapshotsMu.Lock()
+			defer s.snapshotsMu.Unlock()
```
hanahmily, line 131: Locking a channel writing is not good practice.

pkg/flow/streaming/flow_sliding_window.go
```diff
@@ -0,0 +1,334 @@
...
+	timerHeap *flow.DedupPriorityQueue
+
+	// aggregationFactory is the factory for creating aggregation operator
+	aggregationFactory flow.AggregationOpFactory
+
+	// For api.Operator
+	in           chan flow.StreamRecord
+	out          chan flow.StreamRecord
+	done         chan struct{}
+	purgedWindow chan timeWindow
```
hanahmily, line 101: The sender and receiver are two routines. That means you don't have to use a channel to sync them. To merge them is more convince way.

pkg/flow/streaming/flow_sliding_window.go
```diff
@@ -0,0 +1,334 @@
...
+func eventTimeTriggerOnElement(window timeWindow, ctx TriggerContext) TriggerResult {
+	if window.MaxTimestamp() <= ctx.GetCurrentWatermark() {
+		// if the watermark is already past the window fire immediately
+		return FIRE
+	}
+	ctx.RegisterEventTimeTimer(window.MaxTimestamp())
+	return CONTINUE
+}
+
+type TriggerContext interface {
```
hanahmily, line 288: What's this exported interface used for? Its invokers are from this package. I don't find other implements more than triggerContext.

pkg/flow/streaming/flow_topn.go
```diff
@@ -0,0 +1,171 @@
...
+		if topNAggrFunc.sortKeyExtractor == nil {
+			s.f.drainErr(errors.New("sortKeyExtractor must be specified"))
+		}
+		if topNAggrFunc.sort == modelv1.Sort_SORT_DESC {
+			topNAggrFunc.comparator = func(a, b interface{}) int {
+				return utils.Int64Comparator(b, a)
+			}
+		} else {
+			topNAggrFunc.comparator = utils.Int64Comparator
+		}
```
hanahmily, line 56: The order should default to DESC if users don't specific one.

hanahmily (commented):
The flow process is good for me expect for some nits

pkg/flow/streaming/flow_topn.go
lujiajing1126: Fixed. And according to our last discussion, an Enum is introduced in `streaming` package.

pkg/flow/streaming/flow_sliding_window.go
lujiajing1126: Removed redundent interface

pkg/flow/streaming/flow_sliding_window.go
lujiajing1126: Merged

pkg/flow/streaming/flow_sliding_window.go
lujiajing1126: Fixed

pkg/flow/streaming/flow_sliding_window.go
lujiajing1126: Fixed

pkg/flow/streaming/flow_sliding_window.go
```diff
@@ -0,0 +1,311 @@
...
+	s.Wait()
+	return nil
+}
+
+func (s *TumblingTimeWindows) Exec(downstream flow.Inlet) {
+	s.Add(1)
+	go flow.Transmit(&s.ComponentState, downstream, s)
+}
+
+func NewTumblingTimeWindows(size time.Duration) *TumblingTimeWindows {
```
hanahmily, line 220: An interval will generate several windows based on the slide size. The final result relays on the last window which might contain a subset of the complete list.

The lateness flag will cause an unexpected result. For example, if a single element comes late, the flow creates a new window as the final result. This interval will have a new list with an element instead of the previous one. 

The window should hold all elements in an interval:

* Create a window to represent an interval.
* Receive all elements from this interval, sort them, and leave the top n items based on the strategy.
* Send a snapshot of this window to the sinker on the slide passed. 
* Purge this window when the watermark passes its max time. You can wait a while when the lateness is set to collect some elements come late.

A window can not be created more than once otherwise a new window has to load data flushed lastest.

hanahmily (changes requested):

pkg/flow/streaming/flow_sliding_window.go
lujiajing1126: > An interval will generate several windows based on the slide size.

No, it is not correct. Currently, our implementation is Tumbling window. An element can only be assign to a single window instead of multiple ones.

hanahmily:
```
t.errCh = t.streamingFlow.Window(streaming.NewTumblingTimeWindows(t.slideSize)).
```

But the window's size equals a slide instead of an interval. 

Even the window's size matches an interval. How do you flush based on the slide size of an interval? I found a single size in the current window here.

hanahmily: The slide is neccecry. We should flush the snapshot to the disk several times in an interval. Especially for a big interval: one hour, one day, etc.

lujiajing1126: > ```
> t.errCh = t.streamingFlow.Window(streaming.NewTumblingTimeWindows(t.slideSize)).
> ```
> 
> But the window's size equals a slide instead of an interval.

The Sliding size means how often we flush the window in order to persist TopN results. Probably the name is confusing?

> Even the window's size matches an interval. How do you flush based on the slide size of an interval? I found a single size in the current window here.

No. Just as the previous implementation, the flush interval, a.k.a the slideSize, is equal to 40% of the measure interval.

We will assign a proper timestamp based on the measure interval and the flush timestamp for each final TopN result set.

> The slide is neccecry. We should flush the snapshot to the disk several times in an interval. Especially for a big interval: one hour, one day, etc.

The answer for the first sentence is for sure: we still have a "slide". For large interval, we may find a better solution.

lujiajing1126: > A window can not be created more than once otherwise a new window has to load data flushed lastest.

I agree that a window can be only created once. This could produce more precise result. I will check the current impl.

hanahmily: all right, I noticed `snapshots` contains a list of aggregation results. A window is just a flush event.

hanahmily: This is still an issue.

lujiajing1126: > This is still an issue.

Fixed now. PTAL again.

lujiajing1126:
@hanahmily Now the test cases are stable. PTAL. 

The only issue left is supporting `UNSPECIFIED` sort direction. I will finish it soon.

hanahmily:
@lujiajing1126 It's ready to review?

lujiajing1126:
> @lujiajing1126 It's ready to review?
> 
> 

Yes.

banyand/measure/measure_topn.go
```diff
@@ -0,0 +1,681 @@
...
+
+func (manager *topNProcessorManager) buildFilter(criteria *modelv1.Criteria) (flow.UnaryFunc[bool], error) {
+	// if criteria is nil, we handle all incoming elements
+	if criteria == nil {
+		return func(_ context.Context, dataPoint any) bool {
+			return true
+		}, nil
+	}
+
+	f, err := manager.buildFilterForCriteria(criteria)
```
hanahmily, line 391: `pkg/query/logical/tag_filter` provided a enhanced filter. Please use it simplify the manager.

banyand/metadata/metadata.go
```diff
@@ -31,6 +31,12 @@ import (
 	"github.com/apache/skywalking-banyandb/pkg/run"
 )
 
+// TopNFilter provides methods to find a specific topNAggregation which can be attached to the given resource.
+// Currently, the resource can only be Measure
+type TopNFilter interface {
+	TopNAggregations(ctx context.Context, source *commonv1.Metadata) ([]*databasev1.TopNAggregation, error)
```
hanahmily, line 37: Move it to `MeasureRegistry` to align the API's structure.

banyand/query/processor_topn.go
```diff
@@ -0,0 +1,436 @@
...
+	entityMap := make(map[string]int)
+	entity := make([]tsdb.Entry, 1+len(topNSchema.GetGroupByTagNames()))
+	entity[0] = tsdb.AnyEntry
+	for idx, tagName := range topNSchema.GetGroupByTagNames() {
+		entityMap[tagName] = idx + 1
+		// fill AnyEntry by default
+		entity[idx+1] = tsdb.AnyEntry
+	}
+	for _, pairQuery := range conditions {
+		// TODO: check op?
```
hanahmily, line 146: Yep. Raise an error if it's not `EQ` and the tag is not in `group_by_tag_names`

banyand/measure/measure_write.go
```diff
@@ -55,6 +56,15 @@ func (s *measure) Write(value *measurev1.DataPointValue) error {
 		close(waitCh)
 		return err
 	}
+	// send to stream processor
```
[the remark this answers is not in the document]
lujiajing1126: We have a `topNProcessCallback` to listen from the pipeline.

The `Write` method here should both,
1) write measure to the storage and,
2) send it for streaming processing.

THIS IS ONLY FOR TESTING.

banyand/measure/measure_topn.go
hanahmily: Yep. It's not easy, but we have to. The match operation is complex, we should not provide two variants here. The code will become more complicated to maintain from time to time. Furthermore, the topn filter supports a subset of criteria. We must enhance it in the future if we still select this path.

A convenient way is to transform `[]*model_v1.TagFamilyForWrite` to `[]*model_v1.TagFamily` by appending `name` to them.

lujiajing1126:
Tests added

hanahmily (approved):
LGTM. Thanks for submitting such an outstanding contribution. 💯

wu-sheng:
Finally we get this merged. Congrats.

PR #166, opened 2022-08-29
Introduce full text searching to stream and measure

hanahmily:
* API: Add `Analyzer` to `IndexRule`
* API: Add the` Match` operator to the binary conditions
* Inverted Index supports searching analyzed terms

Related to https://github.com/apache/skywalking/issues/9417


wu-sheng (approved):
APIs are good to me.

PR #170, opened 2022-09-02
[BREAKING CHANGES] Introduce OR logical operation

hanahmily:
These commits will introduce:

* [x] Introduce OR logical operation to APIs
* [x] Add tree-based IndexFilter and TagFilter to filter items
* [x] Update relevant Stream query components
* [x] Update relevant TSDB query components
* [x] Update relevant Measure query components

The `Criteria` update brings a lot of changes to query testing of `Liaison`, `Stream`, `Measure`, and `Query` modules. This PR will merge them to `test/integration` to reduce the duplicated cases and accelerate the testing process.

* [x] Move Stream query test cases
* [x] Move Mesure query test cases
* [x] Move Query execution cases
* [x] Move Liaison test cases

We should tweak The analysis test cases of Query due to the tricky compare operations. I will skip them since we don't have the analysis result query APIs. The follow-up PRs will bring them back.


wu-sheng (approved):
The protocol had been reviewed before.

pkg/query/logical/expr_literal.go
```diff
@@ -90,32 +89,26 @@ type int64ArrLiteral struct {
...
+		return 0, slices.Equal(i.arr, o.arr)
 	}
 	return 0, false
 }
 
-func (i *int64ArrLiteral) BelongTo(tagValue *modelv1.TagValue) bool {
-	intValue := tagValue.GetInt()
-	if intValue != nil {
-		return slices.Contains(i.arr, intValue.Value)
+func (i *int64ArrLiteral) BelongTo(other LiteralExpr) bool {
```
lujiajing1126, line 99: The semantics of `BelongTo` is not consistent for all `ComparableExpr`,

1. For singled values, it mean if `this` is an element of a slice, or it is equal to the element,
2. For slices, it means if the given element is a subset of `this` slice, which inverses the relationship.

pkg/query/logical/index_filter.go
```diff
@@ -0,0 +1,647 @@
...
+	"strings"
+
+	"github.com/apache/skywalking-banyandb/api/common"
+	database_v1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/database/v1"
+	model_v1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/model/v1"
+	"github.com/apache/skywalking-banyandb/banyand/tsdb"
+	"github.com/apache/skywalking-banyandb/pkg/convert"
+	"github.com/apache/skywalking-banyandb/pkg/index"
+	"github.com/apache/skywalking-banyandb/pkg/index/posting"
+	"github.com/pkg/errors"
```
lujiajing1126, line 33: A format issue

pkg/query/logical/expr_literal.go
hanahmily: yes, it sounds like a bug here.

PR #172, opened 2022-09-10
Stream and measure schema management

WuChuSheng1:
update vue-cli@5.0.8


wu-sheng:
I thought we are going to use vue-cli, as it is not updated anymore.

Anything changed?

WuChuSheng1:
> I thought we are going to use vue-cli, as it is not updated anymore.
> 
> Anything changed?

I know Vue cli will no longer be maintained, so my suggestion is to use vite. However, since element UI is not compatible with vite, only element plus is compatible with vite; However, element plus requires vue3, so I suggest upgrading the UI to vite + vue3 + element plus; But it takes time, so I want to use it for the time being vue-cli@5.0.8 Carry out front end and back-end joint debugging. I will upgrade the project from next week.

wu-sheng:
As this would be a part of official 0.2.0 release, it is better we at least use reasonable tech stack, rather than an expired time. @hanahmily told me, he will need more time to make server ready to release, you should still have time

WuChuSheng1:
> As this would be a part of official 0.2.0 release, it is better we at least use reasonable tech stack, rather than an expired time. @hanahmily told me, he will need more time to make server ready to release, you should still have time

Yes, please rest assured that I will use a reasonable technology stack to upgrade the project before releasing version 0.2.0

hanahmily:
> As this would be a part of official 0.2.0 release, it is better we at least use reasonable tech stack, rather than an expired time. @hanahmily told me, he will need more time to make server ready to release, you should still have time

The updating will affect too many places than expected. We have to upgrade to vue3 to replace vuecli. If we put them in a single commit, it will be too hard to review. @WuChuSheng1 will update them in the follow-up PRs.

hanahmily:
@WuChuSheng1 Please run `make license-dep` in the root path to fix license-related issues.

PR #177, opened 2022-09-28
Introduce two key feats to tsdb

hanahmily:
* Add the retention controller to support the group TTL. Fixes https://github.com/apache/skywalking/issues/9453
* Allow writing cold data whose timestamp is not in the current block. This PR is a primitive implement to https://github.com/apache/skywalking/issues/8815


hanahmily:
@wu-sheng @lujiajing1126 All checks passed.

wu-sheng (approved):

PR #180, opened 2022-09-28
add measure and group cmd

sacloudy:
- add stream query test (failed to pass)
- add measure cmd and test
- add group cmd and test


hanahmily:
@sacloudy Please add `start` and `end` flags to `query` commands.

hanahmily (approved):
LGTM

PR #183, opened 2022-10-06
Fix test cases with Eventually semantics

lujiajing1126:

This PR tends to fix test cases failures due to async write.

Before this PR, integration tests randomly failed.


test/cases/measure/data/data.go
```diff
@@ -26,16 +26,16 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
-	g "github.com/onsi/ginkgo/v2"
-	gm "github.com/onsi/gomega"
+	. "github.com/onsi/ginkgo/v2"
```
hanahmily, line 29: ci might fails

lujiajing1126: > ci might fails

Oh, I see the linter issues... Let me revert back this part

lujiajing1126: But It seems we miss format for the test package?

hanahmily (approved):

test/cases/measure/data/data.go
hanahmily: No. It doesn't ignore it. But we have to make the order deterministic.

PR #186, opened 2022-10-10
Add Exist endpoints to all services

hanahmily:
Fixes https://github.com/apache/skywalking/issues/8965


api/proto/banyandb/database/v1/rpc.proto
```diff
@@ -233,6 +264,8 @@ service IndexRuleRegistryService {
       get:"/v1/index-rule/schema/lists/{group}"
     };
   };
+  // Exist doesn't expose an HTTP endpoint. Please use HEAD method to touch Get instead
+  rpc Exist(IndexRuleRegistryServiceExistRequest) returns (IndexRuleRegistryServiceExistResponse);
```
wu-sheng, line 268: Why isn't `exist` exposed as a HTTP endpoint?

hanahmily: Elasticsearch's `Exist` function uses HTTP's `HEAD` method to query the server index path. If the code is 404, the return value is false. 200 returns true.

Our new `Exist` services always return 200. If a client wants to use the `HEAD` method, it could touch the current `Get` services. They will work as Elasticsearch's behavior.

hanahmily: These changes goal is to provide a fancy way to interact with the Java client to avoid throwing a GRPCRuntimeExpection.

If a client leverages HTTP Restful-APIs, the `Get` services are better options for them to issue `HEAD` requests.

wu-sheng: I think there is no point to follow ElasticSearch behavior, banyandb has no logic connection with them 
We just need our internally consistent. If this is a service, exposing to get and response as a JSON seems all right.

wu-sheng (approved):

PR #188, opened 2022-10-10
Fix having semantic inconsistency

hanahmily:
Fixes https://github.com/apache/skywalking/issues/9631


wu-sheng:
What is the difference between belongTo and contain?

hanahmily:
> What is the difference between belongTo and contain? Two directions?

`BelongTo` implements `In`. `Contains` is for `Having`.

wu-sheng (approved):

PR #192, opened 2022-10-15
This PR is mainly used to solve the bugs caused by upgrading the technology stack

WuChuSheng1:


wu-sheng:
A 40k size jpg seems unreasonable huge. Could you try a smaller version?

hanahmily:
> A 40k size jpg seems unreasonable huge. Could you try a smaller version?

Compressed it to ~4kb

hanahmily (approved):

PR #195, opened 2022-10-22
Add bydbctl's examples

hanahmily:
The CRUD and query examples are based on `bydbctl`. 

@lujiajing1126 @wankai123 you could use this command line tool to query schemas and data from server.


wu-sheng:
Please update menu file accordingly, https://github.com/apache/skywalking-banyandb/blob/main/docs/menu.yml

There are a lot of removed and new added menu items. Also, we don't have ctl docs on the menu.

docs/crud/property/schema.md
```diff
@@ -0,0 +1,124 @@
...
+    str:
+      value: "hello"
+- key: state
+  value:
+    str:
+      value: "succeed"
+EOF
+```
+
+The operation supports updating partial tags.
```
wu-sheng, line 63: The create and update seem not to match. Then the partial updating actually is a creating?

docs/crud/property/schema.md
```diff
@@ -0,0 +1,124 @@
...
+  shard_num: 2
+  block_interval:
+    unit: UNIT_HOUR
+    num: 2
+  segment_interval:
+    unit: UNIT_DAY
+    num: 1
+  ttl:
+    unit: UNIT_DAY
+    num: 7
```
wu-sheng, line 33: In the UI template use case, the TTL should not be set. Could you be more specific that this is optional?

docs/crud/property/schema.md
wu-sheng: - `name: container1` -> `name: sw`
- `id: property1` -> `id: ui_template`

I think in the real use case, the name should `ui_template`, and id should be `General-Service` for an example

wu-sheng: Use `sw` or `default` as group should be fine, (in codes, it is `sw`). But it should be consistent in these examples.

docs/crud/stream/schema.md
```diff
@@ -0,0 +1,109 @@
...
+  name: default
+catalog: CATALOG_STREAM
+resource_opts:
+  shard_num: 2
+  block_interval:
+    unit: UNIT_HOUR
+    num: 2
+  segment_interval:
+    unit: UNIT_DAY
+    num: 1
```
wu-sheng, line 29: A small discussion, would we support block_size and segment_size in the future or now?
In a typical use case, size-oriented policy is widely used for storage/disk operation.

wu-sheng: Of course, in those cases, we should discuss TTL policy as well. Because they are relative logically.

docs/crud/measure/schema.md
```diff
@@ -0,0 +1,123 @@
...
+  tags:
+  - name: id
+    type: TAG_TYPE_ID
+  - name: entity_id
+    type: TAG_TYPE_STRING
+fields:
+- name: total
+  field_type: FIELD_TYPE_INT
+  encoding_method: ENCODING_METHOD_GORILLA
+  compression_method: COMPRESSION_METHOD_ZSTD
```
wu-sheng, line 58: TBD, these options and explanations should be added in the concept chapters when you fix those docs.
Including how data layout in a KV system.

docs/crud/measure/schema.md
```diff
@@ -0,0 +1,123 @@
...
+  ttl:
+    unit: UNIT_DAY
+    num: 7
+EOF
+```
+
+The group creates two shards to store data points. Every day, it would create a
+segment that will generate a block every 2 hours.
+
+The data in this group will keep 7 days.
```
wu-sheng, line 38: Should we mention how would process various dimensions(day/hour/minute) measures? Because this is a unique design from OAP, and most time-series DB doesn't have this kind of support.

docs/crud/property/schema.md
hanahmily: done

docs/crud/property/schema.md
hanahmily: The `ResourceOpt` should be left empty here. I also remove `requried` validation from APIs.

hanahmily:
> Please update menu file accordingly, https://github.com/apache/skywalking-banyandb/blob/main/docs/menu.yml
> 
> There are a lot of removed and new added menu items. Also, we don't have ctl docs on the menu.

Fixed

docs/crud/stream/schema.md
hanahmily: We will set the maximum number of bytes of storage blocks to retain in a shard. Some algorithms are based on time, and can not work with fixed-size segments/blocks. Theoretically, The oldest blocks will be removed first. 

If TTL conflicts with the max size, they could work separately. For instance, a block that doesn't reach its time end can be removed by the size limitation controller. This strategy is a canonical way and opted for by several TSDBs.

docs/crud/measure/schema.md
hanahmily: Will add them in separate docs.

docs/crud/measure/schema.md
hanahmily: SW's dimensions link to the measure's `interval`, which is for the compression algorithm of the database. We could mention it in skywalking's banyandb storage part.

docs/crud/stream/schema.md
wu-sheng: The typical setup could be a threshold for the minimal size of the block, which is expected to be pre-allocated by the operation team.

wu-sheng:
Good for me, wait for the tests to pass. Once the architecture doc is added, we should be good to do 0.2.0 release.

lujiajing1126 (approved):

PR #200, opened 2022-10-31
Fix panic with single entity-related condition

hanahmily:


lujiajing1126 (approved):
LGTM. Change title to something like "Fix panic with single entity-related condition"?

PR #208, opened 2022-11-09
Improve workflows and building tools

hanahmily:
* Rearrange steps of workflows
* Introduce building tools cache to throttle up the workflow
* Introduce dependency review

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,17 +27,8 @@ jobs:
     name: Check
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/setup-node@v3
-        with:
-          node-version: 16.15
-      - name: Install Go
-        uses: actions/setup-go@v2
-        with:
-          go-version: 1.19
       - name: Check out code into the Go module directory
-        uses: actions/checkout@v2
-      - name: Check License Header
-        run: make license-check
+        uses: actions/checkout@v3
       - uses: actions/cache@v3
         id: cache-go
         with:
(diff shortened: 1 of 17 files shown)


.github/workflows/ci.yml
```diff
@@ -27,18 +27,10 @@ jobs:
...
-        with:
-          go-version: 1.19
       - name: Check out code into the Go module directory
-        uses: actions/checkout@v2
-      - name: Check License Header
-        run: make license-check
-      - uses: actions/cache@v3
+        uses: actions/checkout@v3
+      - name: Cache Go Modules
+        uses: actions/cache@v3
```
wu-sheng, line 33: Check whether you need the timeout setting, https://github.com/apache/skywalking/pull/9894

hanahmily: @Superskyyy Do we need this patch here? It looks like an OAP-related variable.

wu-sheng: It isn't. It is GitHub action tool parameter. That segment mean cache segment.
I didn't find the original post, it was in other discussions. @Superskyyy Could you share the source?

kezhenxu94: https://github.com/apache/skywalking/pull/9863#issuecomment-1298596785

Superskyyy: Hello, the timeout setting is not related to OAP itself but GitHub actions.

It should be used at any GitHub action step which involves downloading artifact previously stored by the actions/cache@3. Otherwise, it might get stuck for an hour. For example, in the OAP repo case it's used to retrieve the cached infra-e2e build.

Superskyyy: See a case here https://github.com/apache/skywalking/actions/runs/3368331192/jobs/5588961894

Fix (the timeout variable) is documented here.
https://github.com/actions/cache/blob/main/workarounds.md#cache-segment-restore-timeout

Superskyyy: sorry thought I clicked comment but was pending :|

wu-sheng (approved):

PR #217, opened 2022-11-18
Move group as the first doc of concepts

wu-sheng:
image

By following this, the group is the top concept, so CRUD docs should follow them as well.

@lujiajing1126 We seem to need some docs to introduce which `group`, `measure`, `stream`, `properties` are used in OAP server?

### DIFF
--- a/docs/menu.yml
+++ b/docs/menu.yml
@@ -28,6 +28,8 @@ catalog:
         path: "/concept/data-model"
   - name: "CRUD Operations"
     catalog:
+      - name: "Group"
+        path: "/crud/group"
       - name: "Measure"
         catalog:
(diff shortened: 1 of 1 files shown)


lujiajing1126:
> @lujiajing1126 We seem to need some docs to introduce which `group`, `measure`, `stream`, `properties` are used in OAP server?

For example, we can add some docs in this page? https://skywalking.apache.org/docs/main/v9.2.0/en/setup/backend/backend-storage/

wu-sheng:
Yes, we don't have to list all, but we should provide parameters to make users known for CLI. 
Because we don't have a UI to visualize them in an easy way.

hanahmily (approved):

PR #223, opened 2022-11-29
Fix issues found through onboarding the showcase env

hanahmily:
These changes contain issues found by integrating the showcase workloads and the BanyanDB server.

## API
- Add a field to the interval writing message for debugging.

## Bugs:
- Correct the overlapping of blocks/segments
- Correct block start time
- Correct the docID of the inverted index

## Enhancement:
- Add more debug info
- Refactor logs
- Update block and segment prefix templates
- Add series literal to debug info
- Compute log output only if the debug log is enabled

## Chores
- Bump up OAP


wu-sheng (approved):
It is good to see showcase helps.

wu-sheng (approved):

PR #224, opened 2022-12-04
Add UI for creating and editing tagfamilies and tags. Change some UI styles.

WuChuSheng1:
This PR is mainly for add UI for creating and editing tagfamilies and tags, and to change some UI styles.

**1. add UI for creating and editing tagfamilies and tags.**
- The original 'new resources' UI：

image

  - The old UI does not provide the function of adding and editing tagfamilies and tags,This is not good for the user experience.So I added a UI for this, which provides new and editing functions.
  - First, you can right-click the group and click 'new resources' to enter the dialog box of adding' resources'：

image
image

  - Obviously, it provides add, delete, edit and batch delete functions.
  - You can click the 'Add the tagfamilies' button to enter the dialog box for adding' tagfamilies'：

image
It provides add, delete, edit and batch delete functions too.

**Tips:
It is worth noting that the UI is not connected to the background interface temporarily, but it does not affect the original new 'resources' function. I need to know whether the newly added 'resources' interface provides the newly added tagfamily function, and what is its data structure? Maybe you can help me?**

-  change some UI styles：
  - The old right click menu:

image

  - The new right click menu:

image

  - The old dialog:

image
image

  - The new dialog:

image
image
image


hanahmily:
@WuChuSheng1 
>  need to know whether the newly added 'resources' interface provides the newly added tagfamily function, and what is its data structure?

The API keeps as before and doesn't provide a separate tag family CRUD operations.

As I mentioned, the "Resource" is a representation notion for UI. APIs only contain Stream and Measure, there is no "Resource" object in the API's specification.

ui/src/components/Database/Aside/components/DialogResources/TagDialog.vue
```diff
@@ -0,0 +1,168 @@
...
+    data.dialogVisible = props.visible
+})
+watch(() => props.group, () => {
+    data.form.metadata.group = props.group
+})
+
+function confirmForm() {
+    ruleForm.value.validate((valid) => {
+        if (valid) {
+            emit('confirm', data.form)
```
hanahmily, line 84: Could you add validation to check the tag's name is unique in all tag families? And the tag family's name can not be duplicated as well.

hanahmily (changes requested):

ui/src/components/Database/Aside/components/DialogResources/TagDialog.vue
WuChuSheng1: of course

hanahmily (approved):

PR #225, opened 2022-12-05
Drop desperate lint rule 

hanahmily:
Remove the desperate lint rule and dead codes.


.github/workflows/flaky-test.yml
```diff
@@ -25,6 +25,7 @@ env:
 
 jobs:
   test:
+    if: github.repository == 'apache/skywalking-banyandb'
```
wu-sheng, line 28: Is this intentional?

wu-sheng (dismissed):

.github/workflows/flaky-test.yml
hanahmily: Yes, tend to block PR to run this test.

wu-sheng (approved):

PR #227, opened 2022-12-11
Enhance the inverted index, metadata and logger

hanahmily:
There are several enhancements in this PR:

## Improve the inverted index

`SeriesID` is a mandatory field that is scanned in each equal operation. In this PR, I composite `SeriesID` and `Term` to a composite index. The equivalent process gets rid of roaringbitmap's intersection operation. The new path improves the ID matching's performance. 

The side effect is: If the tag is marked as an `Analyzer`, the inverted index engine will split this composited index for analyzing the `Term`. That if you do an equal query on a tag with an `Analyzer`, the performance would be worse a bit.

This is related to https://github.com/apache/skywalking/issues/10119.

## Metadata 

The Etcd-powered metadata registry closing process will wait till all running operations are done. It will prevent the retry interceptor from repeatedly connecting to a closed server, which will slow down the testing.

## Logger

Etcd uses the Zap logger, while others pick up the Zerologger. The PR introduces a Zap config generator that outputs a similar log configuration with the Zerolog. 

In the development mode, both loggers will print their stacks and callers will help developers diagnose the code.

### DIFF
--- a/banyand/Dockerfile
+++ b/banyand/Dockerfile
@@ -34,8 +34,6 @@ ENV GO111MODULE "on"
 WORKDIR /src
 COPY go.* ./
 RUN go mod download
-RUN GOBIN=/bin go install github.com/grpc-ecosystem/grpc-health-probe@v0.4.11 \
-    && chmod 755 /bin/grpc-health-probe
 
 FROM base AS builder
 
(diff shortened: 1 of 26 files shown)


wu-sheng:
> If the tag is marked as an Analyzer

What does this mean? Is this a new attribute of the tag? Is there any docs update about this?

hanahmily:
> > If the tag is marked as an Analyzer
> 
> What does this mean? Is this a new attribute of the tag? Is there any docs update about this?

No new attributes here. It means a tag's index rule sets a specific Analyzer in https://github.com/apache/skywalking-banyandb/blob/main/docs/api-reference.md#indexruleanalyzer.

wu-sheng (approved):

PR #229, opened 2022-12-13
Defining "IN" and "NOT IN" APIs in query module.Fixes apache/skywalking [BanyanDB]#9759

DevPJ9:
Description
Added "In" and "Not_In" parsers in "index_filter.go" and "tag_filter.go" so that the query module can implement them.


hanahmily:
@DevPJ9 The codes seem good. Could you add a test case to verify it? The `or` operation could be transformed to `in` easily. You can refer to https://github.com/apache/skywalking-banyandb/blob/main/test/cases/measure/data/input/linked_or.yaml to write the case.

test/cases/measure/data/input/in.yaml
```diff
@@ -0,0 +1,56 @@
...
+        right:
+          le:
+            op: "LOGICAL_OP_in"
+            left:
+              condition:
+                name: "id"
+                op: "BINARY_OP_EQ"
+                value:
+                  id:
+                    value: "unknown"
```
hanahmily, line 56:
```suggestion
criteria:
  condition:
    name: "id"
    op: "BINARY_OP_IN"
    value:
      str_array:
        value: ["4", "5", "unknown"]
```

hanahmily (changes requested):

hanahmily (approved):
LGTM. Thank you for your contribution.

PR #239, opened 2023-01-10
Unify TagFilter for Streaming scenario

lujiajing1126:
This PR refactors `TagFilter` interface for better performance and replace the incomplete filter implementation in the TopN aggregation.

The new `TagFilter` fully takes advantage of parsed schema and thus improves O(N) match to O(1).

Closes https://github.com/apache/skywalking/issues/9770


lujiajing1126:
I've reverted changes made to the `BuildSchema()` in the `logical/measure` package, but extracted the underlying `TagSpecRegistry` to allow simply using the fine-grained component. This helps overcome the cyclic import issue without breaking the current design.

Further, I've polished the write part by sharing the measure write method among different cases,

 - write **real** measure entities,
 - write **virtual** measure entities, i.e. `FlowRecord` generated by `TopNAggregation`, by transforming to a `DataPointForWrite` entity.

The latter would help further optimization, e.g. fully separate `topN` to another package. But this required much work to be done, for example, the current `Repository` is high coupled with a so-called `ResourceSupplier` which is unexpected in the original design.

So I think this PR is sufficient to resolve the current issue. We could stop here.

banyand/measure/measure_write.go
```diff
@@ -60,7 +61,7 @@ func (s *measure) write(md *commonv1.Metadata, shardID common.ShardID, entity []
 	if err != nil {
 		return err
 	}
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	ctx, cancel := context.WithTimeout(context.TODO(), 5*time.Second)
```
hanahmily, line 64: Why do you change it to `TODO`? This context is the top-level, `Background` is better here.

hanahmily (changes requested):
The simplification is much better than before. But introducing the fake measure seems not a good idea.

banyand/measure/measure_write.go
lujiajing1126: There is no difference between `TODO` and `Background` from the implementation. Anyway, reverted.

lujiajing1126:
I've refactored the wring path of `TopN Aggregation` results by creating new internal writing request to the pipeline. And every TopN schema is now bounded to a real Measure schema which is used to listen and write TopN entities.

The virtual/real measure concepts are still there since the measure of `TopN` is incomplete.

banyand/measure/measure_write.go
```diff
@@ -39,17 +39,18 @@ import (
 
 var errMalformedElement = errors.New("element is malformed")
 
-func (s *measure) write(md *commonv1.Metadata, shardID common.ShardID, entity []byte, entityValues tsdb.EntityValues, value *measurev1.DataPointValue) error {
+func (s *measure) write(sm *databasev1.Measure, shardID common.ShardID, entity []byte, entityValues tsdb.EntityValues,
```
hanahmily, line 42: You had better remove the first parameter `sm`. We could get it from `s` directly.

banyand/liaison/grpc/registry.go
```diff
@@ -435,7 +437,37 @@ type topNAggregationRegistryServer struct {
 func (ts *topNAggregationRegistryServer) Create(ctx context.Context,
 	req *databasev1.TopNAggregationRegistryServiceCreateRequest,
 ) (*databasev1.TopNAggregationRegistryServiceCreateResponse, error) {
-	if err := ts.schemaRegistry.TopNAggregationRegistry().CreateTopNAggregation(ctx, req.GetTopNAggregation()); err != nil {
+	topNSchema := req.GetTopNAggregation()
```
hanahmily, line 440: Move the measure creation operation into `topNProcessorManager.start`. Liaison is a proxy and should not implement complicated functions.

If there is no such measure, the manager should create a new one. The manager should check the on-disk measure's specs. If it's different from expectation, it could get updated.

banyand/liaison/grpc/registry.go
```diff
@@ -435,7 +437,37 @@ type topNAggregationRegistryServer struct {
...
+	// create a "virtual" measure for TopN result
+	topNMeasure := &databasev1.Measure{
+		Metadata: topNSchema.GetMetadata(),
+		Interval: sourceMeasure.GetInterval(),
+		TagFamilies: []*databasev1.TagFamilySpec{
+			{
+				Name: measure.TopNTagFamily,
+				Tags: []*databasev1.TagSpec{
+					{
+						Name: "measureID",
```
hanahmily, line 458: use a snake case here: `measure_id`

banyand/liaison/grpc/registry.go
```diff
@@ -453,12 +485,13 @@ func (ts *topNAggregationRegistryServer) Update(ctx context.Context,
 func (ts *topNAggregationRegistryServer) Delete(ctx context.Context,
 	req *databasev1.TopNAggregationRegistryServiceDeleteRequest,
 ) (*databasev1.TopNAggregationRegistryServiceDeleteResponse, error) {
-	ok, err := ts.schemaRegistry.TopNAggregationRegistry().DeleteTopNAggregation(ctx, req.GetMetadata())
-	if err != nil {
+	okTopNDeleted, errDelTopN := ts.schemaRegistry.TopNAggregationRegistry().DeleteTopNAggregation(ctx, req.GetMetadata())
+	okMeasureDeleted, errDelMeasure := ts.schemaRegistry.MeasureRegistry().DeleteMeasure(ctx, req.GetMetadata())
```
hanahmily, line 489: We don't have to support the cascade deletion. We don't introduce transactional operations. The operation is too dangerous.

banyand/liaison/grpc/registry.go
```diff
@@ -435,7 +437,37 @@ type topNAggregationRegistryServer struct {
...
+	topNSchema := req.GetTopNAggregation()
+	if err := ts.schemaRegistry.TopNAggregationRegistry().CreateTopNAggregation(ctx, topNSchema); err != nil {
+		return nil, err
+	}
+
+	sourceMeasure, err := ts.schemaRegistry.MeasureRegistry().GetMeasure(ctx, topNSchema.GetSourceMeasure())
+	if err != nil {
+		return nil, err
+	}
+	// create a "virtual" measure for TopN result
```
hanahmily, line 449: "virtual" is a common notion. How about using `derived` or `corresponding` instead?

hanahmily (changes requested):
some nits

banyand/liaison/grpc/registry.go
lujiajing1126: Fixed

banyand/liaison/grpc/registry.go
lujiajing1126: This part is moved to `measure` pkg. PTAL

banyand/liaison/grpc/registry.go
lujiajing1126: Removed

banyand/liaison/grpc/registry.go
lujiajing1126: Use `derived` now

banyand/liaison/grpc/registry.go
lujiajing1126: We cannot do this since `topNProcessorManager` will start in the `PreRun` stage, but the measure notifier is registered in the `Serve` stage which will not be called if new measures are create.....Any idea?

lujiajing1126: I think we should avoid creating new measures while loading.......

banyand/measure/measure_write.go
lujiajing1126: Removed

banyand/liaison/grpc/registry.go
hanahmily: Placing the creation in the loading phase ensures the measure always exists. Restarting the server will create the target measure if it's missing. As I mentioned, banyandb doesn't support transactional operations. Creating more than one resource in a task is too dangerous.

There are two options here:

1. Move `processorManager.start` to the `Serve` phase. That means you should iterate all measures to start their embedded `processorManager`
2. Introduce Etcd's transactional operation. Move the creation operation to the `metadata` module instead of `liaison`.

lujiajing1126: > 1. Move `processorManager.start` to the `Serve` phase. That means you should iterate all measures to start their embedded `processorManager`

I take the first approach which is straight forward.

> 2. Introduce Etcd's transactional operation. You could leverage a transaction to create topN resource and its derived measure by moving the creation operation to the `metadata` module instead of `liaison`.

We should support transactional op which would be helpful for https://github.com/apache/skywalking/issues/10010. The current client impl could cause incomplete metadata registration in some edge cases.

hanahmily: We might support the transactional operations on fixing 10010

hanahmily (approved):

PR #241, opened 2023-01-17
fix metadata register bug.

HHoflittlefish777:


banyand/metadata/schema/group.go
```diff
@@ -87,6 +88,9 @@ func (e *etcdSchemaRegistry) DeleteGroup(ctx context.Context, group string) (boo
 }
 
 func (e *etcdSchemaRegistry) CreateGroup(ctx context.Context, group *commonv1.Group) error {
+	if group.UpdatedAt != nil {
```
lujiajing1126, line 91: I suppose we can move `UpdatedAt` field assignment to the `create` and `update` methods

lujiajing1126 (changes requested):
Shall we also update e2e cases to verify these changes against the latest OAP. [The current commit](https://github.com/apache/skywalking/commit/1335a48f1c034abc1fe24f6197ee7acfa3118bf0) is from Nov, 2022.

api/proto/banyandb/database/v1/schema.proto
```diff
@@ -88,25 +89,25 @@ enum CompressionMethod {
...
-  repeated TagFamilySpec tag_families = 2;
+  repeated TagFamilySpec tag_families = 2 [(validate.rules).repeated.min_items = 1];
   // fields denote measure values
   repeated FieldSpec fields = 3;
   // entity indicates which tags will be to generate a series and shard a measure
-  Entity entity = 4;
+  Entity entity = 4 [(validate.rules).message.required = true];
   // interval indicates how frequently to send a data point
   // valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h", "d".
   string interval = 5;
```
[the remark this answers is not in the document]
hanahmily: RegExp might work, but I never like such expressions that are a bit tricky. How about validating it in the metadata model by `timestamp.ParseDuration`?

banyand/metadata/schema/group.go
hanahmily: Refection? Would you be able to elaborate more on this?

lujiajing1126:
```go
// TypeMeta defines the identity and type of an Event.
type TypeMeta struct {
	Name  string
	Group string
	Kind  Kind
}

// Metadata wrap dedicated serialized resource and its TypeMeta.
type Metadata struct {
	Spec Spec
	TypeMeta
}
```

Reflection is not necessary. In the `Metadata`, we've already known its type. Use `switch...case` is enough

api/proto/banyandb/database/v1/schema.proto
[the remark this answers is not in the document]
lujiajing1126: I agree. Validating it in the code would be more flexible.

banyand/metadata/schema/group.go
hanahmily: Type assertion is not a good idea that introduces cycle dependence to the codes. Even if they lay in the same package, it is not a good practice.
I suggest leaving it alone.

HHoflittlefish777:
The pr has some error rollback in local environment, so I would like to pull new request same as it.

HHoflittlefish777:
please review in https://github.com/apache/skywalking-banyandb/pull/246

PR #248, opened 2023-02-06
Adding support for creating streams

WuChuSheng1:
**This PR is to merge  create the stream function.**
- **In this function, you can right-click the menu and click "new resources" button.**
image
- **Then, It will enter the create stream page.**
image
- **On this page, you can set the name of the stream.**
image
- **Of course, This page provides the function of editing Tag Families. You can click the "Add" button to edit.**
image
- **You can add multiple Tag Families and Tags. It is worth noting that Tags are unique.**
image
- **Of course, you can also edit and delete the existing Tag Family.**
image
- **When you click the "submit" button, a new stream will be created, and the creation page will be closed and automatically enter the new stream.**
image
- **When you do not need the stream, you can right-click the menu and click the "delete" button to delete it.**
image

**It is worth noting that at present, only new streams are supported, and stream editing is temporarily not supported.**


Superskyyy:
Please don't form your PR title like this, it gets carried over to the squash and merge phase (will be used as the Top commit message) Maybe try something more straightforward like `Adding support for creating streams` (just an example)

WuChuSheng1:
> Please don't form your PR title like this, it gets carried over to the squash and merge phase (will be used as the Top commit message) Maybe try something more straightforward like `Adding support for creating streams` (just an example)

ok, Thank you

hanahmily (changes requested):
1. At the top of creating the page, "Catalog" should be in place of "Type". "Operation" could replace "Operator"
2. Remove the prefix, "TAG_TYPE_", from the tag type's literal. For example, "INT" could substitute for "TAG_TYPE_INT".
3. Deletion procedure should get confirmation from the user before deleting a stream. A confirm dialog should pop up when the user clicks the delete button.

WuChuSheng1:
> 1. At the top of creating the page, "Catalog" should be in place of "Type". "Operation" could replace "Operator"
> 2. Remove the prefix, "TAG_TYPE_", from the tag type's literal. For example, "INT" could substitute for "TAG_TYPE_INT".
> 3. Deletion procedure should get confirmation from the user before deleting a stream. A confirm dialog should pop up when the user clicks the delete button.

ok

hanahmily:
I think this looks good to me. But we have to wait till v0.3.0 is released, which is in the voting process.

hanahmily (approved):

WuChuSheng1:
> I think this looks good to me. But we have to wait till v0.3.0 is released, which is in the voting process.

Thank you very much for your approval.

PR #253, opened 2023-02-19
Write-ahead Logging interface

HHoflittlefish777:
Write-ahead Logging interface design.


pkg/wal/wal.go
```diff
@@ -0,0 +1,46 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+// Package wal (Write-ahead logging) is an independent component to ensure data reliability.
+package wal
+
+// Log stands for a write-ahead logging instance.
+type Log struct{}
```
hanahmily, line 22: Does it implement the WAL interface? If that, make it unexported.

pkg/wal/wal.go
```diff
@@ -0,0 +1,46 @@
...
+
+// Options for Write-ahead Logging.
+type Options struct{}
+
+// Segment stands for a segment instance of Write-ahead log.
+type Segment struct{}
```
hanahmily, line 28: This should be an interface instead of a dedicated type.

pkg/wal/wal.go
```diff
@@ -0,0 +1,46 @@
...
+
+// Wal include exposed interfaces.
+type Wal interface {
```
hanahmily, line 31:
```suggestion
// WAL includes exposed interfaces.
type WAL interface {
```

pkg/wal/wal.go
```diff
@@ -0,0 +1,46 @@
...
+	// New creates a Log instance in the specified root directory.
+	New(path string, opts *Options) (*Log, error)
```
hanahmily, line 33: Could you transfer it to a package-scoped function? This way is a go conventional path to build a constructor.

pkg/wal/wal.go
```diff
@@ -0,0 +1,46 @@
...
+	// Write request to the WAL buffer.
+	// It will return synchronously when the request write in the WAL buffer,
+	// and trigger asynchronous callback when return when the buffer is flushed to disk successfully.
+	Write(data []byte) (func(), error)
+	// Read specified segment by index.
+	Read(index int) (*Segment, error)
```
hanahmily, line 39: Where can we get the `index`?

pkg/wal/wal.go
```diff
@@ -0,0 +1,46 @@
...
+	// ReadAllSegments operation reads all segments.
+	ReadAllSegments() ([]*Segment, error)
+	// Rotation closes the open segment and opens a new one, returning the closed segment details.
+	Rotation() (*Segment, error)
```
hanahmily, line 43:
```suggestion
	Rotate() (*Segment, error)
```

hanahmily (changes requested):

pkg/wal/wal.go
HHoflittlefish777: Rotate can get detail of segment, index is the segment id.

hanahmily: How can a user get the segment id?

pkg/wal/wal.go
```diff
@@ -0,0 +1,46 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+// Package wal (Write-ahead logging) is an independent component to ensure data reliability.
+package wal
+
+// Options for Write-ahead Logging.
```
hanahmily, line 21:
```suggestion
// Options for creating Write-ahead Logging.
```

pkg/wal/wal.go
```diff
@@ -0,0 +1,46 @@
...
+type Options struct{}
+
+// Segment stands for a segment instance of Write-ahead log.
```
hanahmily, line 24:
```suggestion
// Segment allows reading underlying segments that hold WAl entities 
```

hanahmily (changes requested):

pkg/wal/wal.go
HHoflittlefish777: Use rotate api can return a segment detail，and user can get segment id.

pkg/wal/wal.go
```diff
@@ -0,0 +1,46 @@
...
+type Segment interface{}
+
+// WAL includes exposed interfaces.
+type WAL interface {
+	// Write request to the WAL buffer.
+	// It will return synchronously when the request write in the WAL buffer,
+	// and trigger asynchronous callback when return when the buffer is flushed to disk successfully.
```
hanahmily, line 31:
```suggestion
	// Write a logging entity.
	// It will return immediately when the data is written in the buffer,
	// The returned function will be called when the entity is flushed on the persistent storage.
```

pkg/wal/wal.go
```diff
@@ -0,0 +1,46 @@
...
+	Write(data []byte) (func(), error)
+	// Read specified segment by index.
```
hanahmily, line 33:
```suggestion
	// Read specified segment by SegmentID.
```

pkg/wal/wal.go
```diff
@@ -0,0 +1,46 @@
...
+	Read(index int) (*Segment, error)
+	// ReadAllSegments operation reads all segments.
```
hanahmily, line 35:
```suggestion
	// ReadAllSegments reads all segments sorted by their creation time in ascending order.
```

pkg/wal/wal.go
```diff
@@ -0,0 +1,46 @@
...
+	ReadAllSegments() ([]*Segment, error)
+	// Rotate closes the open segment and opens a new one, returning the closed segment details.
+	Rotate() (*Segment, error)
+	// Delete the specified segment.
+	Delete(index int) error
+}
+
+// New creates a Log instance in the specified root directory.
```
hanahmily, line 43:
```suggestion
// New creates a WAL instance in the specified path.
```

pkg/wal/wal.go
```diff
@@ -0,0 +1,46 @@
...
+func New(_ string, _ *Options) (*WAL, error) {
```
hanahmily, line 44:
```suggestion
func New(_ string, _ Options) (WAL, error) {
```

pkg/wal/wal.go
```diff
@@ -0,0 +1,46 @@
...
+// Package wal (Write-ahead logging) is an independent component to ensure data reliability.
+package wal
+
+// Options for Write-ahead Logging.
+type Options struct{}
+
+// Segment stands for a segment instance of Write-ahead log.
+type Segment interface{}
+
+// WAL includes exposed interfaces.
```
hanahmily, line 27:
```suggestion
// WAL denotes a Write-ahead logging. 
// Modules who want their data reliable could write data to an instance of WAL. 
// A WAL combines several segments, ingesting data on a single opened one. 
// Rotating the WAL will create a new segment, marking it as opened and persisting previous segments on the disk.
```

hanahmily (changes requested):

pkg/wal/wal.go
hanahmily: here is an example, https://github.com/apache/skywalking-banyandb/blob/main/api/common/id.go#L31

pkg/wal/wal.go
```diff
@@ -0,0 +1,54 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+// Package wal (Write-ahead logging) is an independent component to ensure data reliability.
+package wal
+
+// SegmentID identity a segment in a WAL.
```
hanahmily, line 21:
```suggestion
// SegmentID identifies a segment in a WAL.
```

hanahmily (changes requested):

hanahmily (approved):

wu-sheng:
@hanahmily Do we release0.3.1 patch first? Or you plan to cherry pick and release?

hanahmily:
> @hanahmily Do we release0.3.1 patch first? Or you plan to cherry pick and release?

I will hold this PR for a while till 0.3.1 is released. I want to fix the flaky case before releasing it in case of some unrevealed bugs.

hanahmily (approved):

PR #256, opened 2023-02-21
Update change log

hanahmily:


wu-sheng (approved):
Let's sync to the main repo testing and verify this version before release.

PR #257, opened 2023-02-25
[OAP Integration] Register missing TopN registry service

lujiajing1126:
Closes https://github.com/apache/skywalking/issues/10472


hanahmily:
@wu-sheng We might release 0.3.2 ? From past experience, a round integration will bring several patches to the server and java-client.

wu-sheng:
Yes, we need  new release. Considering new feature added, should we consider 0.4.0?

hanahmily:
> Yes, we need new release. Considering new feature added, should we consider 0.4.0?

maybe, the WAL's interface is merged into the main. If the implementation gets merged, 0.4.0 is mandatory.

lujiajing1126:
Basically, it works for multiple groups.

Still two issues,

### How can we serialize `seriesID` of the original measure

The current protocol for `seriesID` in the response is `string` as `name` field. In order to deserialize it properly, we may need a more sophisticated design.

### How to elegantly disable NULL group

From the issue https://github.com/apache/skywalking/issues/10472, quoted

> If you don't want to generate such a group, TopNAggregation{criteria:service_id!=null}. != is the BINARY_OP_NE. The result would be like this: service1:[endpoint1:100...], service2:[endpoint2:90]...

I totally disagree with this design which would mix a nested filter for source measure with options of `TopN` streaming operator. We may add `disableNullGroup` to the `TopNAggregation` schema.

hanahmily:
> ### How can we serialize `seriesID` of the original measure
> The current protocol for `seriesID` in the response is `string` as `name` field. In order to deserialize it properly, we need a more sophisticated design.
>


A string name is not appropriate for now.

An array wraps `model.v1.Tag` seems better than a sterilized string. The key of a tag could be from the source measure's schema.

Here is my suggestion for this API:

```
message Item {
    repeated model.v1.Tag entity = 1;
    model.v1.FieldValue value = 2;
  }
```

> I totally disagree with this design which would mix a nested filter for source measure with options of `TopN` streaming operator. We may add `disableNullGroup` to the `TopNAggregation` schema.
> 

Our nested filter's implementation is based on the query processor. That's why we push https://github.com/apache/skywalking/issues/9770 to the streaming framework.

Mixing them up is our long-term target. Finally, banyandb should provide a SQL variant to filter data in the streaming flow.

lujiajing1126:
> A string name is not appropriate for now.
> 
> An array wraps `model.v1.Tag` seems better than a sterilized string. The key of a tag could be from the source measure's schema.

This part is implemented in the last commit.

> Our nested filter's implementation is based on the query processor. That's why we push [apache/skywalking#9770](https://github.com/apache/skywalking/issues/9770) to the streaming framework.
> 
> Mixing them up is our long-term target. Finally, banyandb should provide a SQL variant to filter data in the streaming flow.

I agree with this roadmap. So when a measure is received, we have to duplicate it by setting all groups tags to NULL (in order to filter it with filter)? It seems an implicit logic.

lujiajing1126:
After discussion, we've decided not to have the NULL group as an exception. Instead, NULL group will simply a case of fuzzy search if no condition is given. This will be properly handled in the query stage (not in the steaming stage any more)

wu-sheng:
What do you mean NULL group?

lujiajing1126:
> What do you mean NULL group?

In https://github.com/apache/skywalking/issues/10472, @hanahmily mentioned this concept,

> The null group collects data points without tags specified in group_by_tag_names. The null group will help OAP to generate global and sub scopes with a single schema specification.

wu-sheng:
@hanahmily Could you explain what is the issue? Why it would lead to `a case of fuzzy search`?

hanahmily:
@lujiajing1126 NULL group is a case of [fuzzy search](https://www.ibm.com/docs/en/informix-servers/12.10?topic=modifiers-fuzzy-searches), which is a term of search engine(ES, lucence ,and etc).

The NULL group collects data points w/o grouping tags or w/ tags, which are set NULL values.

lujiajing1126:
> @lujiajing1126 NULL group is a case of [fuzzy search](https://www.ibm.com/docs/en/informix-servers/12.10?topic=modifiers-fuzzy-searches), which is a term of search engine(ES, lucence ,and etc).

I mean from the internal implementation, NULL group support is basically a "fuzzy" search for any tag values. The structure of seriesID is consist of the following Entity, `{metadata_names}_{sortDirection}_{rankNo}_{tagValues}`

- `{metadata_names}` is the source measure name and topN schema name,
- `{sortDirection}` is used by `field_value_sort` in the `TopNRequest`,
- `{rankNo}` the rank number of the record,
- `{tagValues}` represents multiple entities. The size is equal to the length of `groupByTagName` in the schema.

Then for NULL group, the query stage will search for `{metadata_names}_{sortDirection}_{rankNo=*}_{tagValues=*}`. In another word, it will post aggregate all groups in the query stage. So the NULL group is now a virtual concept. According to  @hanahmily , we can do it based on the following reasons,

- In OAP now, in almost any case, `SERVICE_ID` is always specified. It means global TopN for all services may not be used.
- The current implementation is more resilient, e.g. support query with a subset of groupBy tags

wu-sheng:
`{metadata_names}_{sortDirection}_{rankNo=*}_{tagValues=*}` Please don't follow this way. Pre-aggregation(topN) is resource cost, we only should take this when necessary, rather than a query replacement.

@lujiajing1126 I think you mixed the concepts again. Whether `serviceID` is used or not in the query stage, that is how to use `pre-aggregate` value(group).

wu-sheng:
We should have `tag` definition for pre-aggregation, then, the tag value should be specific or NULL. Then we have value-based groups and a NULL group(mostly for edge case only).
This is not for expanding wide query scope.

wu-sheng:
> @lujiajing1126 NULL group is a case of [fuzzy search](https://www.ibm.com/docs/en/informix-servers/12.10?topic=modifiers-fuzzy-searches), which is a term of search engine(ES, lucence ,and etc).

@hanahmily I think you missed a `NOT`? From what we discussed this morning.

lujiajing1126:
> @lujiajing1126 I think you mixed the concepts again. Whether `serviceID` is used or not in the query stage, that is how to use `pre-aggregate` value(group).

I cannot understand what you were saying. For example, `[ServiceID]` MUST be registered as `groupByTagNames` in the `TopNAggregation` schema if it is possible to occur in the query stage.

Then in this case, multiple groups (one for each `service_id`) will be generated in which a pre-aggregation has been done. NOTE: `NULL` group is virtual since `NULL` group does not exist in these groups.

I don't know what's problem here. I suppose I'm strictly following what we've discussed last week @hanahmily

wu-sheng:
> [ServiceID] MUST be registered as groupByTagNames in the TopNAggregation schema if it is possible to occur in the query stage.

This is correct.

> Then in this case, multiple groups (one for each service_id) will be generated in which pre-aggregation has been done. NOTE: NULL group is virtual since NULL group does not exist in these groups.

There is not problem about this too.

___
> I don't know what's problem here. I suppose I'm strictly following what we've discussed last week

The key is `{metadata_names}_{sortDirection}_{rankNo=*}_{tagValues=*}` misleading for many things. There is no `*` or `fuzzy-query`(you mentioned)

In the real case, there are only several cases
1. Value specified in the query. Use the specific group
2. No value specified in the query, Use all existing groups(including NULL group) to get the accurate final list
3. (No sure) Whether we accept `NULL` value as a query condition? If so, pick the NULL group.

lujiajing1126:
> The key is `{metadata_names}_{sortDirection}_{rankNo=*}_{tagValues=*}` misleading for many things. There is no `*` or `fuzzy-query`(you mentioned)

This is just internal mechanism. "Fuzzy" is actually key scanning with prefix. It is very fast. Let's move on.

banyand/measure/measure_write.go
```diff
@@ -131,9 +131,12 @@ func (s *measure) write(shardID common.ShardID, entity []byte, entityValues tsdb
...
 	if s.processorManager != nil {
-		s.processorManager.onMeasureWrite(&measurev1.WriteRequest{
-			Metadata:  s.GetMetadata(),
-			DataPoint: value,
+		s.processorManager.onMeasureWrite(&measurev1.InternalWriteRequest{
+			Request: &measurev1.WriteRequest{
+				Metadata:  s.GetMetadata(),
+				DataPoint: value,
+			},
+			EntityValues: entityValues[1:],
```
hanahmily, line 139: Why do you remove the first value? This value is from the measure's specification. The first is not a reserved entity.

banyand/measure/measure_topn.go
```diff
@@ -469,18 +498,20 @@ func (manager *topNProcessorManager) buildMapper(fieldName string, groupByNames
...
+				return stringify(dpWithEvs.GetTagFamilies()[locator.FamilyOffset].GetTags()[locator.TagOffset])
 			}), "|"),
 			// field value as v2
 			// TODO: we only support int64
-			dataPoint.GetFields()[fieldIdx].GetInt().GetValue(),
+			dpWithEvs.GetFields()[fieldIdx].GetInt().GetValue(),
 			// groupBy tag values as v3
 			transform(groupLocator, func(locator partition.TagLocator) *modelv1.TagValue {
-				return dataPoint.GetTagFamilies()[locator.FamilyOffset].GetTags()[locator.TagOffset]
+				return dpWithEvs.GetTagFamilies()[locator.FamilyOffset].GetTags()[locator.TagOffset]
```
hanahmily, line 514: If the tag is absent, a panic of `index out of range` might be raised. Suppose there are three tags here, and the group tag is the last one(tag3), [tag1,tag2] whose length is 2 is a valid data point.

test/cases/measure/data/testdata/service_instance_cpm_minute_data.json
```diff
@@ -0,0 +1,212 @@
...
+            }
+          },
+          {
+            "str": {
+              "value": "entity_1"
+            }
+          },
+          {
+            "str": {
+              "value": "svc_1"
```
hanahmily, line 18: Could you add an empty service id here to verify the `NULL` group?

test/cases/topn/data/input/asc.yaml
```diff
@@ -16,7 +16,7 @@
 # under the License.
 
 metadata:
-  name: "service_cpm_minute_top_bottom_100"
+  name: "service_instance_cpm_minute_top_bottom_100"
   group: "sw_metric"
 topN: 1
```
hanahmily, line 21: Please use a number more than one as we discussed before

hanahmily (changes requested):
We should align some notions:

`NULL` group: An abnormal data group that holds data points doesn't have tags defined by groupTagNames or have these tags with a `NULL` value.
Virtual `Global` group: this group doesn't get persisted in the intermediate result. If the query criteria don't set groupTagNames, data points in each existing group will be aggregated into the `Global` group, which is returned to the client.

banyand/measure/measure_write.go
lujiajing1126: > Why do you remove the first value? This value is from the measure's specification. The first is not a reserved entity.

https://github.com/apache/skywalking-banyandb/blob/e68249a385a720d075eab52b496d7c4b4679ba41/pkg/partition/entity.go#L58-L59

The first entity is just the source measure name

banyand/measure/measure_topn.go
lujiajing1126: Fixed

test/cases/measure/data/testdata/service_instance_cpm_minute_data.json
lujiajing1126: Added

test/cases/measure/data/testdata/service_instance_cpm_minute_data.json
```diff
@@ -155,7 +155,7 @@
           },
           {
             "str": {
-              "value": "svc_3"
```
hanahmily: You should have mocked a `Null` instead of an empty string. Could you cover two kinds of `Null`:

1. A null type by `{"null":0}`
2. A tag array with fewer elements by removing the last value from these three elements.

hanahmily (changes requested):
Please increase test coverage.

hanahmily:
@lujiajing1126 Would you please update `CHNAGES.md` as well? We introduce a new PR template similar to the main repo.

test/cases/measure/data/testdata/service_instance_cpm_minute_data.json
lujiajing1126: I'm not sure if we can do this in this case.

ServiceID is not only the groupBy tag but also the `SeriesID` of the source measure. It cannot be NULL?

hanahmily: It could be. groupBy tag could be any tags defined by a measure more than tags in the entity. ServiceID is a scenario from OAP that should not limit the scope of test cases.

lujiajing1126: Another case is added in order to verify the NULL group

hanahmily (approved):
LGTM

PR #260, opened 2023-03-12
[UI] Add YAML language filter

WuChuSheng1:
### <You can use YAML language to filter data>
- [ UI ] You can use yaml to filter the current page, quantity and time; At the same time, the yaml editor will cache your editing content, so you don't have to worry about editing again when switching.
image


hanahmily (approved):

hanahmily:
@WuChuSheng1 Please update `CHANGES.md` on submitting a PR.

hanahmily (approved):

PR #259, opened 2023-03-07
Add tsdb concept document

hanahmily:
Add a document explaining the process of TSDB ingesting and retrieving time series data.

This buffer is not implemented yet. It will be added in the 0.4.0.

- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking-banyandb/blob/main/CHANGES.md).
- [x] Upload images to website

PR #262, opened 2023-03-23
Add a sharded buffer to ingest data

hanahmily:
### <Feature description>

The `Buffer` serves as a concurrent, sharded, in-memory buffer. It is designed to handle high concurrency and efficiently store and retrieve time-series data. 

The `Buffer` is composed of multiple `bufferShardBucket` instances, each containing its own `skiplist` for key-value storage. This sharded architecture helps to distribute the load and minimize contention when accessing data.

I will integrate the `Buffer` with `block` in the following-up PR.

- [x] Update the documentation to include this new feature. #259 

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@ Release Notes.
 - Add TSDB concept document.
 - [UI] Add YAML editor for inputting query criteria.
 - Refactor TopN to support `NULL` group while keeping seriesID from the source measure.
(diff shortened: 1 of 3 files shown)


wu-sheng (dismissed):
Before 0.4 release, I think we need better and more detailed graphs to visualize/doc our storage concept in the docs.

wu-sheng (approved):

PR #265, opened 2023-04-16
reload topn

lujiajing1126:
In PR#239, I've moved `startSteamingManager` to the `service.Serve()` method which causes issue for TopN/measure reloading.


banyand/measure/metadata.go
```diff
@@ -115,6 +121,13 @@ func (sr *schemaRepo) OnAddOrUpdate(metadata schema.Metadata) {
 			})
 		}
 	case schema.KindTopNAggregation:
+		// createOrUpdate TopN schemas in advance
+		err := sr.createOrUpdateTopNMeasure(metadata.Spec.(*databasev1.TopNAggregation))
```
hanahmily, line 125: When the "TopnAggregation" is deleted, the corresponding measure should also be removed.

pkg/test/measure/etcd.go
```diff
@@ -66,13 +71,80 @@ func PreloadSchema(e schema.Registry) error {
...
 	if err := loadSchema(topNAggregationDir, &databasev1.TopNAggregation{}, func(topN *databasev1.TopNAggregation) error {
-		return e.CreateTopNAggregation(context.TODO(), topN)
+		return multierr.Append(e.CreateTopNAggregation(context.TODO(), topN), createOrUpdateTopNMeasure(e, topN))
 	}); err != nil {
 		return errors.WithStack(err)
 	}
 	return nil
 }
 
+func createOrUpdateTopNMeasure(reg schema.Registry, topNSchema *databasev1.TopNAggregation) error {
```
hanahmily, line 81: Could you tell me why you're copying codes here? You have already created measures for processing topNAggregation creation events.

If the issue involves resolving flaky scenarios, using the 'eventually' function may be helpful.

hanahmily (changes requested):

pkg/test/measure/etcd.go
lujiajing1126: > Could you tell me why you're copying codes here? You have already created measures for processing topNAggregation creation events.

If you remember, `preloadMeasure` and `preloadMeasure` services (specifically the `preload` stage) are started just before the `measure`/`stream` service. So no listener/callback is setup when these schemas are installed. This could ensure that every loaded schema will take effect as soon as the `measure` and `stream` starts. I guess it is your motivation to do so at the beginning.

> If the issue involves resolving flaky scenarios, using the 'eventually' function may be helpful.

No related to this scenario at all. But of course, we could move `preload*` after `measure` and `stream` modules. But we have to wait every actual storage modules are ready.

hanahmily: There is a bug in the system that needs to be addressed. If we create a topNAggregation without creating the corresponding measure, either due to a crash or normal shutdown, the "measure" will not be built after rebooting.

To fix this issue, we need to create these measures during the `PreRun` phase of the measure service. This means that there are two ways to create topNAggregation's measures:

1. By handling topNAggregation create or update events.
2. By booting up the measure service.

lujiajing1126: > To fix this issue, we need to create these measures during the `PreRun` phase of the measure service.

As we've discussed multiple times, it is not feasible. Since watcher is created in the `Serve()` stage, we cannot do this in this `preRun()` stage!

A possible solution is to do a sanity check once after all listeners are setup.

hanahmily (approved):
LGTM

PR #267, opened 2023-04-22
Use TopN in max revision

lujiajing1126:


lujiajing1126:
I've tested this commit locally. It works well with limited resource in OAP E2E test

hanahmily (approved):

PR #272, opened 2023-05-07
Fix resource corrupts caused by update indexRule operation

hailin0:


hanahmily (changes requested):
The patch is good for me. Can you please introduce a test case in `etcd_test.go` to cover the given scenario?

hanahmily (approved):
LGTM

PR #273, opened 2023-05-17
modified TestGoVersion to check Go CPU architectue

achintya-7:
### Fix <Added a check for Go CPU architecture in [TestGoVersion](https://github.com/apache/skywalking-banyandb/blob/main/scripts/ci/check/version_test.go)>
	According to the documentation, it is advisable to run `make check-req` before building BanyanDB from the source. It checks if the device has the necessary software version or not. I recently faced an issue where I had Go 1.20 Linux-386 where `make check-req` passed perfectly but the build failed with this error.

error
 
To resolve this, I had a check which parses through the output string of `go version` and checks if it has 64 in it or not.

### DIFF
--- a/scripts/ci/check/version_test.go
+++ b/scripts/ci/check/version_test.go
@@ -38,7 +38,7 @@ const (
 func TestGoVersion(t *testing.T) {
 	// the value of ptr will be 8 for 64 bit system and 4 for 32 bit system
 	ptr := 4 << (^uintptr(0) >> 63)
-	require.Equal(t, CPUType, ptr, "This CPU architectue is not supported, should be 64 bits Go version")
+	require.Equal(t, CPUType, ptr, "This CPU architectue is not supported, it should be a 64 bits Go version")
 
 	currentVersion := runtime.Version()
 	versionRegex := regexp.MustCompile(`go(\d+\.\d+\.\d)`)
--- a/scripts/ci/check/version_test.go
+++ b/scripts/ci/check/version_test.go
@@ -32,14 +32,13 @@ import (
 
 const (
 	GoVersion = "1.20"
-	CPUType   = "64"
+	CPUType   = 8
 )
 
 func TestGoVersion(t *testing.T) {
-	types := runtime.GOARCH
-
-	ok := strings.Contains(types, CPUType)
-	require.True(t, ok, "CPU type not supported, current[%s], want[%s bit Go release]", types, CPUType)
+	// the value of ptr will be 8 for 64 bit system and 4 for 32 bit system
+	ptr := 4 << (^uintptr(0) >> 63)
+	require.Equal(t, CPUType, ptr, "This CPU architectue is not supported, should be 64 bits Go version")
 
 	currentVersion := runtime.Version()
 	versionRegex := regexp.MustCompile(`go(\d+\.\d+\.\d)`)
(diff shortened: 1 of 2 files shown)


scripts/ci/check/version_test.go
```diff
@@ -28,13 +28,23 @@ import (
...
 )
 
-const GoVersion = "1.20"
+const (
+	GoVersion = "1.20"
+	CpuType   = "64"
+)
 
 func TestGoVersion(t *testing.T) {
 	goversion, err := exec.Command("go", "version").Output()
```
lujiajing1126, line 37: I think we can use `runtime.Version()` here

https://pkg.go.dev/runtime#Version

and use the const GOARCH for arch detection.

achintya-7: Oh yes, this seems to be a much better approach than running an `os.exec` command. I'll implement this on a different branch and can merge it later when we come to a consensus.

hanahmily:
@achintya-7 CI failed. Overall, The implements look good to me.

achintya-7:
@hanahmily It was a lint check issue, have modified the test accordingly.

scripts/ci/check/version_test.go
```diff
@@ -21,22 +21,35 @@ import (
...
+	GoVersion = "1.20"
+	CPUType   = "64"
+)
 
 func TestGoVersion(t *testing.T) {
-	goversion, err := exec.Command("go", "version").Output()
-	require.NoError(t, err)
+	types := runtime.GOARCH
+
+	ok := strings.Contains(types, CPUType)
```
lujiajing1126, line 41: According to https://gist.github.com/asukakenji/f15ba7e588ac42795f421b48b8aede63#goarch-values,

It is not accurate to simply check "64" string. Probably, we can use,

```go
const PtrSize = 4 << (^[uintptr](https://pkg.go.dev/builtin#uintptr)(0) >> 63)
```

From godoc,

> PtrSize is the size of a pointer in bytes - unsafe.Sizeof(uintptr(0)) but as an ideal constant. It is also the size of the machine's native word size (that is, 4 on 32-bit systems, 8 on 64-bit).

lujiajing1126 (changes requested):

scripts/ci/check/version_test.go
achintya-7: Ok. I have tried it on a 64-bit system and it is coming out as 8. I will try it on a 32-bit system or an i386 version of Go to check if it's coming as 4 and then push the new commit.

achintya-7: @lujiajing1126 I have added this snippet and also tested it with both a 64-bit and a 32-bits system. The output is coming out as desired when running the test.

lujiajing1126 (approved):
LGTM

hanahmily (changes requested):
@achintya-7 Could you update the `CHANGES.md` accordingly?

achintya-7:
@hanahmily Sure.

hanahmily (approved):
LGTM, thanks a lot.

HHoflittlefish777:
@achintya-7 hi, we have lots of work todo ,you can look issues about BanyanDB in https://github.com/apache/skywalking, you can choose some jobs that you enjoy, or put forward some of your opinions which BanyanDB can improve. We are currently working on distributed databases and look forward to your contribution.

PR #274, opened 2023-05-22
Add Measure Reader

WuChuSheng1:
### Add Measure Reader
image


ui/src/components2/Read/index.vue
```diff
@@ -185,14 +201,23 @@ function getTableData() {
...
     paramList.metadata = data.resourceData.metadata
-    getTableList(paramList)
+    getTableList(paramList, data.type)
         .then((res) => {
             if (res.status == 200) {
-                setTableData(res.data.elements)
+                if(data.type == 'stream') {
+                    setTableData(res.data.elements)
+                } else {
+                    setTableData(res.data.dataPoints)
```
hanahmily, line 218: The structure of the measure is different from stream. There is a “fields” field more than “tag_families”. https://github.com/apache/skywalking-banyandb/blob/23139b99ec2b0be194c3bf4bab140a125001fa0a/api/proto/banyandb/measure/v1/query.proto#L42

hanahmily (changes requested):

ui/src/components2/Read/index.vue
WuChuSheng1: I will improve it.

hanahmily (approved):

PR #276, opened 2023-05-30
Several improvements to address the issues discovered during load testing

hanahmily:
### <Feature description>
Improve the tsdb core features:

add units to memory flags for easier configuration, and implement TSTable for block structure customization.


wu-sheng:
The sparse index improvement would be separate, right? Do you have a plan on how OAP should be done?

hanahmily:
> The sparse index improvement would be separate, right? Do you have a plan on how OAP should be done?

According to https://github.com/apache/skywalking/issues/10871 OAP doesn't do anything.

wu-sheng (approved):

wu-sheng (approved):

PR #280, opened 2023-06-13
Add access logging and stress test cases

hanahmily:
- Add integration access log to the `liaison` module to record ingested data.
- Add the Istio scenario stress test based on the data generated by the integration access log.


wu-sheng:
How the test/* are going to run?

hanahmily:
> How the test/* are going to run?

You can use either "go test" or "ginkgo" to execute the cases. However, it is recommended to use the latter executor.

To indicate if a case will be executed in the PR CI, there is a label available called "slow". That means only fast(not slow) cases will be run by the PR checker. It's import to notice that the label is only available under the ginkgo.

wu-sheng:
So, it is a local benchmark. Usually, you should put a local result of the benchmark as a text in the root of this case to provide a baseline for other contributors.

hanahmily:
> So, it is a local benchmark. Usually, you should put a local result of the benchmark as a text in the root of this case to provide a baseline for other contributors.

Done.

wu-sheng (approved):

PR #284, opened 2023-06-19
Remove ID tag type

hanahmily:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/10945.


banyand/measure/measure_topn.go
```diff
@@ -190,8 +190,8 @@ func (t *topNStreamingProcessor) writeData(eventTime time.Time, timeBucket strin
 						Tags: append([]*modelv1.TagValue{
 							// MeasureID
 							{
-								Value: &modelv1.TagValue_Id{
```
hanahmily, line 193: The `measure_id`'s `indexRule` is removed. @lujiajing1126 Are there any potential side effects to consider?

lujiajing1126 (dismissed):
lgtm

wu-sheng (approved):

PR #293, opened 2023-07-03
build: drop check node version

tisonkun:
I'm locally using 20.3.0. Without specific reason, perhaps we can drop the node version check step.


hanahmily:
I apologize; we typically avoid breaking CI as a result of node version conflicts. I recommend installing a dedicated node specifically for running the toolchain. However, if you have any alternative solutions, please feel free to share them by raising another PR.

tisonkun:
@hanahmily if you pinned the node version as a feature, then there is no solution. But if you'd like to pin node version in CI environment, setup-node should already be able to take node version parameter.

What's your case here?

hanahmily:
That's a feat of building tool chain.

The checker promises that everyone will pass the building without any pain
because the CI has checked it under that version. We don’t expect to
receive any complaints about the building failure due to node version
mismatch.

The golang tool chain adopted the same convention.

PR #298, opened 2023-07-08
fix: BanyanDB ui unable to load icon

ButterBright:
### Fix <https://github.com/apache/skywalking/issues/11059>

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes https://github.com/apache/skywalking/issues/<issue number>.

res

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,10 @@ Release Notes.
 
 - List all properties in a group.
 
+### Bugs
+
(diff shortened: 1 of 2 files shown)


lujiajing1126 (changes requested):
Pls add changes

lujiajing1126 (approved):
lgtm

PR #305, opened 2023-07-19
Support multiple roles for banyand

hanahmily:
Banyand has been updated to run in multiple roles, but it cannot work in cluster mode. There are still several pending tasks to be completed.

As a reviewer, you should focus on the document that outlines the fundamental behavior of the server with multiple roles in Banyand.


docs/concept/clustering.md
```diff
@@ -35,6 +35,14 @@ BanyanDB integrates multiple roles into a single process in the standalone mode,
 
 In this mode, the single process performs the roles of the Liaison Node, Query Node, Data Node, and Meta Node. It receives requests, maintains metadata, processes queries, and handles data, all within a unified setup.
 
+### 1.6 Mix Mode in Storage Nodes
+
+Query nodes and data nodes are implemented by a same executable binary, Storage Node. With the flag "mode", the storage node can be started as a query node or a data node. The default mode is "mix", which means the storage node is both a query node and a data node.
+
+If the workload of query is high, you can start more storage nodes with the flag "mode" set to "query". If the workload of write is high, you can start more storage nodes with the flag "mode" set to "data".
```
HHoflittlefish777, line 42: I think not only workload, more important is the ability of scaling and saving resource.For example, when peak traffic, serval times resource may needed.If using mix mode, too much resource would be waste and need more time to scale.

docs/concept/clustering.md
```diff
@@ -35,6 +35,14 @@ BanyanDB integrates multiple roles into a single process in the standalone mode,
...
+
```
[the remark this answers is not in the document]
hanahmily: This part of the description also indicates that scalability based on the type of workload can help save resources: To add the dedicated nodes according to the workload type.

banyand/metadata/embeddedetcd/server.go
```diff
@@ -0,0 +1,159 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+// Package embeddedetcd implements an embedded etcd server.
+package embeddedetcd
```
hanahmily, line 19: @lujiajing1126 I've already separated the embed server from the schema registry. Could you update your local repo in case there are some conflicts?

wu-sheng:
So, with this spliting, banyandb still is able to run in stand alone mode? But as support to set an external etcd cluster?

hanahmily:
> So, with this spliting, banyandb still is able to run in stand alone mode? But as support to set an external etcd cluster?

It has the capacity to run as a standalone with embed etcd without external ones.

wu-sheng:
I mean, do we expect am external etcd in cluster? Or still running internally?

PR #306, opened 2023-07-27
Define file system operation interface

HHoflittlefish777:
related issue https://github.com/apache/skywalking/issues/10839


dist/LICENSE
```diff
@@ -456,6 +456,7 @@ MIT licenses
     fill-range 7.0.1 MIT
     follow-redirects 1.15.2 MIT
     form-data 4.0.0 MIT
+    fsevents 2.3.2 MIT
```
hanahmily, line 459: Where is it from? You've not introduced it in `go.mod` yet.

hanahmily (changes requested):
The interface is good for me. But I have no idea where is the added `fsevents` from.

dist/LICENSE
HHoflittlefish777: This is I run make license-dep and it generate automatically

hanahmily: Please remove your local `<banyandb_root_dir>/bin`. And rerun this command.

HHoflittlefish777: The result of removing and rerunning make license-dep is the same.

hanahmily (approved):

PR #310, opened 2023-07-31
Generate dependencies licenses in CI job

hanahmily:


wu-sheng:
What does this gen make for? Usually dependencies LICENSE should be hosted on git, and re-gen to compare diff for forcing update when dependency lib changed.

hanahmily:
> What does this gen make for? Usually dependencies LICENSE should be hosted on git, and re-gen to compare diff for forcing update when dependency lib changed.

It's for the same purpose, but it's absent in the check job.

PR #312, opened 2023-08-06
Add file system vector-IO operations interface.

HHoflittlefish777:
Add file system Vector-IO operations interface.


pkg/fs/file_system.go
```diff
@@ -44,10 +44,14 @@ type Dir interface {
 type File interface {
 	// Append mode, which adds new data to the end of a file.
 	AppendWriteFile(buffer []byte) error
+	// Vector Append mode, which supports appending consecutive buffers to the end of the file.
```
hanahmily, line 47: What's the different from

```go
for _, b range buffers {
        f.AppendWriteFile(b)
}
    
```

Do you implement any optimizations for this operation?

docs/concept/persistence-storage.md
```diff
@@ -100,7 +100,7 @@ return: File pointer, you can use it for various operations.
 
 ### Write
 BanyanDB provides two methods for writing files.
-Append mode, which adds new data to the end of a file. This mode is typically used for WAL.
+Append mode, which adds new data to the end of a file. This mode is typically used for WAL. And BanyanDB supports vector Append mode, which supports appending consecutive buffers to the end of the file.
```
hanahmily, line 103: All append operations are "vector" style. Can you elaborate why we should highlight it.

hanahmily (changes requested):

pkg/fs/file_system.go
HHoflittlefish777: Quite different, the difference between one system call and multiple system calls.

docs/concept/persistence-storage.md
HHoflittlefish777: A 'vector' refers to a continuous set of memory buffers. By using the writev system call, multiple buffers (i.e. vectors) can be written in one operation.
In writev, it is necessary to provide a vector composed of iovec structures, each describing the address and length of a buffer to be written. The content of this vector is equivalent to an array, hence it is called a "write vector".

HHoflittlefish777: That is what linux define.

pkg/fs/file_system.go
HHoflittlefish777: For example, we can use writev system call in Linux to implement it.

hanahmily: To ensure any changes made in a function affect the original two-dimensional slice, it's important to use `*[][]byte` as the parameter.

docs/concept/persistence-storage.md
```diff
@@ -148,6 +156,14 @@ buffer: The read length is the same as the buffer length.
 
 `File.ReadFile(offset int, buffer []byte) (error)`
 
+For vector reading:
+
+param:
+
+iov: Discontinuous buffers in memory.
+
+`File.ReadvFile(iov [][]byte) (error)`
```
hanahmily, line 165: As I mentioned, iov should be a pointer

hanahmily (approved):

wu-sheng:
I can't see where we link to `docs/concept/persistence-storage.md` doc, how a user could read this?

docs/concept/persistence-storage.md
```diff
@@ -112,7 +112,7 @@ return: File pointer, you can use it for various operations.
 
 ### Write
 BanyanDB provides two methods for writing files.
-Append mode, which adds new data to the end of a file. This mode is typically used for WAL.
+Append mode, which adds new data to the end of a file. This mode is typically used for WAL. And BanyanDB supports vector Append mode, which supports appending consecutive buffers to the end of the file.
```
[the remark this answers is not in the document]
HHoflittlefish777: Capitalize just because it's the first letter. So, what should the specification be？

hanahmily:
> I can't see where we link to `docs/concept/persistence-storage.md` doc, how a user could read this?

We have not implemented it. As WAL, once the implementation gets merged, we would publish this doc.

PR #313, opened 2023-08-07
Support for recovery measure/buffer using wal

hailin0:
### Support wal recovery

- [x] [design doc](https://github.com/apache/skywalking-banyandb/blob/main/docs/concept/wal.md).


banyand/tsdb/buffer_test.go
```diff
@@ -147,4 +151,124 @@ var _ = Describe("Buffer", func() {
 			}
 		})
 	})
+
+	Context("Write and Recover of wal correctly", func() {
+		writeConcurrency := 2
+		numShards := 2
+		flushSize := 1024
+		walPath := "wal"
```
hanahmily, line 159: To use `test.Space` to setup a tmp directory to test.

banyand/tsdb/buffer_test.go
```diff
@@ -147,4 +151,124 @@ var _ = Describe("Buffer", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+				currentShardWalFiles, err := os.ReadDir(shardWalDir)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(len(currentShardWalFiles) <= 2).To(BeTrue())
+				// Check wal delete
+				Expect(len(shardWalFileHistory[shardIndex]) > len(currentShardWalFiles)).To(BeTrue())
+			}
+		})
+
+		It("should recover wal file correctly", func() {
```
hanahmily, line 236:
```suggestion
		It("should recover buffer from wal file correctly", func() {
```

hanahmily (changes requested):

pkg/wal/wal.go
```diff
@@ -137,35 +138,56 @@ type segment struct {
...
-	seriesCount   uint32
 	batchLen      uint64
+	seriesCount   uint32
+	dataLen       int
+}
+
+type logSeriesID struct {
+	key     string
+	byteLen int
+}
```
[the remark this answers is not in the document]
hanahmily: Wrap the utility so that it can be used in other places.

hanahmily (approved):

PR #314, opened 2023-08-11
Implement node role registration

hanahmily:
Register the node role to the metadata registry.

There are two categories of nodes:

* Stateful nodes: standalone, data nodes, and mixed storage nodes. Their node id has to be passed through a flag "data-node-id", which is unique and sortable between each other.
* Stateless nodes: Liaison, query nodes. Their node id is generated by the metadata module automatically.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#10895.


wu-sheng:
Does this mean, nodes could be deployed on different roles now? If so, I think the `installation` doc could be updated.

https://skywalking.apache.org/docs/skywalking-banyandb/next/installation/

One is current standalone mode, the other should start from cluster mode deployment.

hanahmily:
> Does this mean, nodes could be deployed on different roles now? If so, I think the `installation` doc could be updated.
> 
> https://skywalking.apache.org/docs/skywalking-banyandb/next/installation/
> 
> One is current standalone mode, the other should start from cluster mode deployment.


Previous commits have already drafted the role-based deployment. 

https://github.com/apache/skywalking-banyandb/pull/314/files#diff-8b042e3f94ca5c59a7cd990b950aec0073ea84fe811e7b22be51158d7b180d56 amend more introduced in this change.

wu-sheng:
I can see the node ID is unique, and considering our query is applying to all data node, why the node IDs are not auto generated? 
What is the point to keep manually setting?

hanahmily:
> I can see the node ID is unique, and considering our query is applying to all data node, why the node IDs are not auto generated? What is the point to keep manually setting?

It is important for the data node id to be stateful because the shard allocation algorithm depends on it. This means that the node id should be either obtained from a flag or configuration, or generated initially and then stored locally. The first is easier and ensures straightforward maintenance and storage.

hanahmily:
> > I can see the node ID is unique, and considering our query is applying to all data node, why the node IDs are not auto generated? What is the point to keep manually setting?
> 
> It is important for the data node id to be stateful because the shard allocation algorithm depends on it. This means that the node id should be either obtained from a flag or configuration, or generated initially and then stored locally. The first is easier and ensures straightforward maintenance and storage.

Some databases can use the hostname as the node id, Would we prefer this pattern?

wu-sheng:
We just had a discussion. The node list should be removed, and no need to add node ID manually.
The data node and shard assignment should be done automatically, as our query solution is there and aggressively querying all nodes.

Further, etcd server is only recommended in a separate deployment. The bundled etcd should be for standalone only officially.

docs/installation/cluster.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+## Node Discovery
+
+The node discovery is based on the etcd cluster. The etcd cluster is required for the metadata module to provide the metadata service and nodes discovery service for the whole cluster.
+
+The host is registered to the etcd cluster by the `banyand-server` automatically based on `node-host-provider` :
+
+- `node-host-provider=hostname` : The OS's hostname is registered as the host part in the address.
+- `node-host-provider=ip` : The OS's the first public active IP address(IPv4) is registered as the host part in the address.
```
wu-sheng, line 44: How do you determine the `public`?

hanahmily: The IP parse algorithm will skip "loopback". Later on, I will change the term.

docs/installation/cluster.md
```diff
@@ -0,0 +1,45 @@
...
+$ ./banyand-server storage --etcd-endpoints=http://10.0.0.1:2379,http://10.0.0.2:2379,http://10.0.0.3:2379 <flags>
+$ ./banyand-server storage --etcd-endpoints=http://10.0.0.1:2379,http://10.0.0.2:2379,http://10.0.0.3:2379 <flags>
+$ ./banyand-server liaison --etcd-endpoints=http://10.0.0.1:2379,http://10.0.0.2:2379,http://10.0.0.3:2379 <flags>
+```
+
+## Node Discovery
+
+The node discovery is based on the etcd cluster. The etcd cluster is required for the metadata module to provide the metadata service and nodes discovery service for the whole cluster.
+
+The host is registered to the etcd cluster by the `banyand-server` automatically based on `node-host-provider` :
```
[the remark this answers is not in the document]
hanahmily: It defaults to "hostname".

wu-sheng (approved):

wu-sheng (approved):

PR #321, opened 2023-08-31
Refactor distributed query

hanahmily:
The new design merges liaison and query nodes to provide a more flexible and simple architecture.


wu-sheng:
So, Liaison is a query node now? There is no separate query node.
Basically, we have only liaison node and data node.

hanahmily:
> So, Liaison is a query node now? There is no separate query node. Basically, we have only liaison node and data node.

Exactly. There are two main roles in a cluster mode.

wu-sheng (approved):

PR #322, opened 2023-08-31
Add mod revision check to write requests(measure/stream)

hailin0:
### <Feature description>
- [x] If it's UI related, attach the screenshots below.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Close apache/skywalking#10956.


api/proto/banyandb/measure/v1/write.proto
```diff
@@ -45,8 +45,23 @@ message WriteRequest {
...
+  STATUS_INVALID_TIMESTAMP = 3;
+  STATUS_INVALID_METADATA = 4;
+  STATUS_EXPIRED_REVISION = 5;
+  STATUS_INTERNAL_ERROR = 6;
+}
+
 // WriteResponse is the response contract for write
-message WriteResponse {}
+message WriteResponse {
+  common.v1.Metadata metadata = 1 [(validate.rules).message.required = true];
```
hanahmily, line 61: Metadata should only be returned in case the status is STATUS_INVALID_METADATA.

api/proto/banyandb/measure/v1/write.proto
```diff
@@ -45,8 +45,23 @@ message WriteRequest {
...
+  STATUS_RECEIVE_ERROR = 2;
+  STATUS_INVALID_TIMESTAMP = 3;
+  STATUS_INVALID_METADATA = 4;
+  STATUS_EXPIRED_REVISION = 5;
+  STATUS_INTERNAL_ERROR = 6;
+}
+
 // WriteResponse is the response contract for write
-message WriteResponse {}
+message WriteResponse {
```
hanahmily, line 60: I would like to define request identification to both request and response.

You can add a unique messageId to both the WriteRequest and WriteResponse messages. With this messageId, we can easily identify which response corresponds to a particular request.

api/proto/banyandb/measure/v1/write.proto
hanahmily: The name should be "latestMetadata" which hints it's the most recent revision.

api/proto/banyandb/stream/v1/write.proto
```diff
@@ -39,13 +39,28 @@ message ElementValue {
...
+  STATUS_UNSPECIFIED = 0;
+  STATUS_SUCCEED = 1;
+  STATUS_RECEIVE_ERROR = 2;
+  STATUS_INVALID_TIMESTAMP = 3;
+  STATUS_INVALID_METADATA = 4;
+  STATUS_EXPIRED_REVISION = 5;
+  STATUS_INTERNAL_ERROR = 6;
+}
+
+message WriteResponse {
```
hanahmily, line 59: The same as the measure's suggestions.

api/proto/banyandb/stream/v1/write.proto
```diff
@@ -39,13 +39,28 @@ message ElementValue {
...
-  // the metadata is only required in the first write.
+  // the metadata is required.
   common.v1.Metadata metadata = 1 [(validate.rules).message.required = true];
   // the element is required.
   ElementValue element = 2 [(validate.rules).message.required = true];
 }
 
-message WriteResponse {}
+// Status is the response status for write
+enum Status {
```
hanahmily, line 49: Please move it and stream's Status into "model" package

banyand/liaison/grpc/discovery.go
```diff
@@ -223,22 +223,25 @@ var _ schema.EventHandler = (*entityRepo)(nil)
 
 type entityRepo struct {
 	log         *logger.Logger
-	entitiesMap map[identity]partition.EntityLocator
+	entitiesMap map[identity]*partition.EntityLocator
```
hanahmily, line 226:
```suggestion
	entitiesMap map[identity]partition.EntityLocator
```

The receiver of methods defined in EntityLocator is the value instead of a pointer.

hanahmily (changes requested):

banyand/liaison/grpc/discovery.go
hanahmily: Exactly. Please return an empty struct. This struct doesn't contain any large field, so instantiating it on the stack is not an overhead.

api/proto/banyandb/measure/v1/write.proto
```diff
@@ -43,10 +44,18 @@ message WriteRequest {
...
   // the data_point is required.
   DataPointValue data_point = 2 [(validate.rules).message.required = true];
+  // the message_id is required.
+  uint64 message_id = 3 [(validate.rules).uint64.gt = 0];
 }
 
 // WriteResponse is the response contract for write
-message WriteResponse {}
+message WriteResponse {
+  common.v1.Metadata last_metadata = 1 [(validate.rules).message.required = true];
```
hanahmily, line 53:
```suggestion
  common.v1.Metadata lastest_metadata = 3 [(validate.rules).message.required = true];
```

api/proto/banyandb/measure/v1/write.proto
```diff
@@ -43,10 +44,18 @@ message WriteRequest {
...
+  // status indicates the request processing result
+  model.v1.Status status = 2 [(validate.rules).enum.defined_only = true];
+  // the message_id from request.
+  uint64 message_id = 3 [(validate.rules).uint64.gt = 0];
```
hanahmily, line 57:
```suggestion
  uint64 message_id = 1 [(validate.rules).uint64.gt = 0];
```

api/proto/banyandb/model/v1/write.proto
```diff
@@ -0,0 +1,34 @@
...
+option go_package = "github.com/apache/skywalking-banyandb/api/proto/banyandb/model/v1";
+option java_package = "org.apache.skywalking.banyandb.model.v1";
+
+// Status is the response status for write
+enum Status {
+  STATUS_UNSPECIFIED = 0;
+  STATUS_SUCCEED = 1;
+  STATUS_RECEIVE_ERROR = 2;
+  STATUS_INVALID_TIMESTAMP = 3;
+  STATUS_INVALID_METADATA = 4;
```
hanahmily, line 31:
```suggestion
  STATUS_NOT_FOUND = 4;
```
This status indicates the resource referred to by the metadata doesn't exist.

api/proto/banyandb/model/v1/write.proto
```diff
@@ -0,0 +1,34 @@
...
+  STATUS_EXPIRED_REVISION = 5;
```
hanahmily, line 32:
```suggestion
  STATUS_EXPIRED_SCHEMA = 5;
```

banyand/liaison/grpc/stream.go
```diff
@@ -76,18 +78,29 @@ func (s *streamService) Write(stream streamv1.StreamService_WriteServer) error {
 		}
 		if err != nil {
 			s.sampled.Error().Stringer("written", writeEntity).Err(err).Msg("failed to receive message")
-			reply(stream, s.sampled)
+			reply(nil, modelv1.Status_STATUS_RECEIVE_ERROR, 0, stream, s.sampled)
```
hanahmily, line 81: Here is a bug introduced by me. When a non-EOF error comes, we should "return" instead of "continue".  The underlying issue causing the error is not addressed, and this could lead to further issues down the line.

Once you fixed this bug, "STATUS_RECEIVE_ERROR" is redundant as well.

pkg/query/logical/common.go
```diff
@@ -60,25 +59,30 @@ func ProjectItem(ec executor.ExecutionContext, item tsdb.Item, projectionFieldRe
...
+		} else {
+			tags = make([]*modelv1.Tag, len(refs))
+			for j, ref := range refs {
+				if len(parsedTagFamily.GetTags()) > ref.Spec.TagIdx {
+					tags[j] = parsedTagFamily.GetTags()[ref.Spec.TagIdx]
+				} else {
+					tags[j] = &modelv1.Tag{Key: ref.Tag.name, Value: nullTag}
+				}
 			}
 		}
```
hanahmily, line 85: There is a lot of repeated code and we can clean it up by simplifying the condition and the loop body.

```go
var tags []*modelv1.Tag
parsedTags := parsedTagFamily.GetTags()

// Determine the size of the tags slice
size := len(parsedTags)
if len(refs) > size {
	size = len(refs)
}

tags = make([]*modelv1.Tag, size)

// Populate the tags slice
for j, ref := range refs {
	if ref.Spec.TagIdx < len(parsedTags) {
		tags[j] = parsedTags[ref.Spec.TagIdx]
	} else {
		tags[j] = &modelv1.Tag{Key: ref.Tag.name, Value: nullTag}
	}
}
```

hanahmily (changes requested):

api/proto/banyandb/measure/v1/write.proto
hanahmily: It should be "latest" or "newest" instead of "last"

hanahmily (approved):

PR #327, opened 2023-09-10
Move clients and observebility docs to the bottom of menu

wu-sheng:

These two parts are needed when a user has chosen and deployed banyandb server.
So, they should have lower priority.


wu-sheng:
@hanahmily Please notice failing tests. It seems to have utc tz issues.

hanahmily (approved):

PR #332, opened 2023-09-17
Add TTL to the property

hanahmily:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#10893.

### DIFF
--- a/docs/concept/data-model.md
+++ b/docs/concept/data-model.md
@@ -183,7 +183,7 @@ Creating property.
 metadata:
   container:
     group: sw
-    name: ui_template
+    name: temp_data
   id: General-Service
 tags:
 - key: name
@@ -204,7 +204,7 @@ The property supports the TTL mechanism. You could set the `ttl` field to specif
 metadata:
   container:
     group: sw
-    name: ui_template
+    name: temp_data
   id: General-Service
 tags:
 - key: name
--- a/docs/concept/data-model.md
+++ b/docs/concept/data-model.md
@@ -198,6 +198,34 @@ tags:
 
 `Property` supports a three-level hierarchy, `group`/`name`/`id`, that is more flexible than schemaful data models.
 
+The property supports the TTL mechanism. You could set the `ttl` field to specify the time to live.
+
+```yaml
+metadata:
+  container:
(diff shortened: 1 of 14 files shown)


docs/concept/data-model.md
```diff
@@ -198,6 +198,34 @@ tags:
...
+ttl: "1h"
+```
+
+"General-Service" will be dropped after 1 hour. If you want to extend the TTL, you could use the "keepalive" operation. The "lease_id" is returned in the apply response. You can use get operation to get the property with the lease_id as well.
+
+```yaml
+lease_id: 1
+```
+
+"General-Service" lives another 1 hour.
```
wu-sheng, line 227: How to extend the lifecycle? Is there a kind of ttl from last read or last write?

hanahmily: Keepalive operation. https://github.com/apache/skywalking-banyandb/pull/332/files#diff-4b9d7be31734b204e1eecbc36ab19918c1a72267558df92b70e434a1bc9cef62R109

The lease_id is returned from the applying operation: https://github.com/apache/skywalking-banyandb/pull/332/files#diff-4b9d7be31734b204e1eecbc36ab19918c1a72267558df92b70e434a1bc9cef62R48

hanahmily: and If you want to get the lease_id separately, the get operation will help as the document mentioned.

docs/crud/property.md
```diff
@@ -111,6 +128,45 @@ List operation lists all properties in a group with a name.
...
+This functionality is supported by the lease mechanism. The readonly lease_id field is used to identify the lease of the property.
+
+### Examples of setting TTL
+
+```shell
+$ bydbctl property apply -f - <<EOF
+metadata:
+  container:
+    group: sw
+    name: ui_template
```
wu-sheng, line 144: Please replace all `ui_template` by `temp_data`. This could mislead people for TTL applied on the `ui_template`, which is not.

hanahmily: Done

wu-sheng (approved):

PR #324, opened 2023-09-06
Calculate topology based on registered data nodes

lujiajing1126:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#11238
- [ ] Update the [`CHANGES` log](https://github.com/apache/skywalking-banyandb/blob/main/CHANGES.md).

PR #334, opened 2023-09-20
Add node selector to support topology calculation

lujiajing1126:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#11238

A continuation of #324

### DIFF
--- a/pkg/node/maglev_test.go
+++ b/pkg/node/maglev_test.go
@@ -104,23 +104,15 @@ func TestMaglevSelector_DiffNode(t *testing.T) {
 	trialCount := 100_000
 	for j := 0; j < trialCount; j++ {
 		metricName := uuid.NewString()
-		fullDataNodeId, _ := fullSel.Pick("sw_metrics", metricName, 0)
-		brokenDataNodeId, _ := brokenSel.Pick("sw_metrics", metricName, 0)
-		if fullDataNodeId != brokenDataNodeId {
-			diff += 1
+		fullDataNodeID, _ := fullSel.Pick("sw_metrics", metricName, 0)
+		brokenDataNodeID, _ := brokenSel.Pick("sw_metrics", metricName, 0)
+		if fullDataNodeID != brokenDataNodeID {
+			diff++
 		}
 	}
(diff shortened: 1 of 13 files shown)


pkg/node/maglev_test.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	commonv1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/common/v1"
+	databasev1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/database/v1"
+)
+
+func TestMaglevSelector(t *testing.T) {
```
hanahmily, line 29: This case is for the consistency result for the same shard. That's great. But let's add more functionality and performance tests:

1. Distribution Testing: This test would validate the distribution of the Maglev hashing. If we query the data node for many different shards, the distribution of the results should be roughly even among all the data nodes.
2. Adding a data node: This test validates how the system behaves when adding a new data node. Ideally, only a small set of the shards should be remapped . *We should know the ratio and document it.*
3. Removing a data node: This test validates how the system behaves when removing a data node. Ideally, only the keys mapped to the removed node should be remapped.
4. Performance Testing: This test is a benchmark that would validate the performance of the Maglev hashing algorithm by checking the time it takes to get the data node for many shards.

hanahmily (changes requested):
The codes look good to me roughly, but additional test cases are necessary before merging.

pkg/node/maglev_test.go
lujiajing1126: For Maglev, 

1. Distribution is even, with epsilon < 3%.
2. Since it is a "stable" hashing algo, the ratio is approx. 1/targetNodesNum, for example, from 9 to 10 nodes, there will be ~10% data moving from an old node to a new one. I suppose only jumphash will generate very few migration.
3. Similarly, < 1/targetNodesNum
4. About 100 nanoseconds per op. which is very low.

hanahmily: Excellent work! Your performance is truly remarkable.

hanahmily (approved):

PR #336, opened 2023-09-27
Add health check command for bydbctl

ButterBright:
### <Feature description>

### DIFF
--- a/bydbctl/internal/cmd/health_check.go
+++ b/bydbctl/internal/cmd/health_check.go
@@ -18,24 +18,123 @@
 package cmd
 
 import (
+	"context"
+	"fmt"
+	"time"
 
-	"github.com/go-resty/resty/v2"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/health/grpc_health_v1"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
 
 	"github.com/apache/skywalking-banyandb/pkg/version"
 )
 
+type healthCheckClient struct {
+	conn *grpc.ClientConn
+}
+
+func newHealthCheckClient(ctx context.Context, addr string, opts []grpc.DialOption) (client *healthCheckClient, err error) {
+	conn, err := grpc.Dial(addr, opts...)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		if err != nil {
+			conn.Close()
+		}
+		go func() {
+			<-ctx.Done()
+			conn.Close()
+		}()
+	}()
+	return &healthCheckClient{conn: conn}, nil
+}
+
+func (g *healthCheckClient) Check(ctx context.Context, _ *grpc_health_v1.HealthCheckRequest, _ ...grpc.CallOption) (*grpc_health_v1.HealthCheckResponse, error) {
+	var resp *grpc_health_v1.HealthCheckResponse
+	if err := rpcRequest(ctx, 10*time.Second, func(rpcCtx context.Context) (err error) {
+		resp, err = grpc_health_v1.NewHealthClient(g.conn).Check(rpcCtx,
+			&grpc_health_v1.HealthCheckRequest{
+				Service: "",
+			})
+		return err
+	}); err != nil {
+		return nil, err
+	}
+	return resp, nil
+}
+
+func rpcRequest(ctx context.Context, rpcTimeout time.Duration, fn func(rpcCtx context.Context) error) error {
+	rpcCtx, rpcCancel := context.WithTimeout(ctx, rpcTimeout)
+	defer rpcCancel()
+	rpcCtx = metadata.NewOutgoingContext(rpcCtx, make(metadata.MD))
+
+	err := fn(rpcCtx)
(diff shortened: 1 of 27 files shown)


hanahmily (changes requested):
The command tool should be programmed to verify the status of the gRPC health service. This is necessary because the data node does not support HTTP services. If it did, we could rely on an HTTP health probe for status checks, which would negate the need for constructing this command tool.

bydbctl/internal/cmd/health_check.go
```diff
@@ -0,0 +1,138 @@
...
+	"google.golang.org/grpc/status"
+
+	"github.com/apache/skywalking-banyandb/pkg/version"
+)
+
+type healthCheckClient struct {
+	conn *grpc.ClientConn
+}
+
+func newHealthCheckClient(ctx context.Context, addr string, opts []grpc.DialOption) (client *healthCheckClient, err error) {
```
hanahmily, line 42: We have implemented it at https://github.com/apache/skywalking-banyandb/blob/main/pkg/test/helpers/grpc_health.go#L35 

Please feel free to reuse this feat.

hanahmily (changes requested):

pkg/test/helpers/grpc_health.go
```diff
@@ -32,15 +32,15 @@ import (
 var errServiceUnhealthy = errors.New("service is unhealthy")
 
 // HealthCheck returns a function for ginkgo "Eventually" poll it repeatedly to check whether a gRPC server is ready.
-func HealthCheck(addr string, connTimeout time.Duration, rpcTimeout time.Duration, opts ...grpc.DialOption) func() error {
+func HealthCheck(disableLogging bool, addr string, connTimeout time.Duration, rpcTimeout time.Duration, opts ...grpc.DialOption) func() error {
```
hanahmily, line 35: I believe we should not disable logging as it helps us identify why health checks fail.

bydbctl/internal/cmd/health_check.go
```diff
@@ -0,0 +1,63 @@
...
+			if grpcCert != "" {
+				creds, errTLS := credentials.NewClientTLSFromFile(grpcCert, "")
+				if err != nil {
+					return errTLS
+				}
+				opts = append(opts, grpc.WithTransportCredentials(creds))
+			} else {
+				opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
+			}
+			grpcAddr := viper.GetString("grpcAddr")
```
hanahmily, line 53: It would be best if you used pflag binding to pass the parameters.

pkg/test/helpers/grpc_health.go
hanahmily: We discussed and agreed that the help command info should be disabled when an error occurs. The regular logging messages should still be displayed.

bydbctl/internal/cmd/health_check.go
```diff
@@ -0,0 +1,63 @@
...
+	healthCheckCmd := &cobra.Command{
+		Use:           "health",
+		Version:       version.Build(),
+		Short:         "Health check",
+		SilenceErrors: true,
+		SilenceUsage:  true,
+		RunE: func(_ *cobra.Command, _ []string) (err error) {
+			opts := make([]grpc.DialOption, 0, 1)
+			grpcCert := viper.GetString("grpcCert")
+			if grpcCert != "" {
```
hanahmily, line 44: We need another flag "insecure" to skip server's cert, which is much helpful when this command is used as the health probe in k8s. When a pod is deployed, we might not know the host name of this workload. Ignore the cert might be suitable for this scenario. 

```go
tlsConfig := &tls.Config{
       InsecureSkipVerify: true,
    }
creds := credentials.NewTLS(tlsConfig)
```

bydbctl/internal/cmd/health_check.go
hanahmily: The flags bound to viper in the root command can be stored in the config file. But these flags you introduced won't be included.

hanahmily (changes requested):

bydbctl/internal/cmd/health_check.go
hanahmily: Furthermore, the HTTP-related requests also need this flag.

```go
client := resty.New()
client.SetTLSClientConfig(&tls.Config{ InsecureSkipVerify: true })
```

bydbctl/internal/cmd/health_check_test.go
```diff
@@ -0,0 +1,73 @@
...
+	It("health", func() {
+		rootCmd.SetArgs([]string{"health", "--grpc-addr", grpcAddr, "--grpc-cert", "../../../test/integration/standalone/other/testdata/server_cert.pem"})
+		out := capturer.CaptureStdout(func() {
+			err := rootCmd.Execute()
+			Expect(err).NotTo(HaveOccurred())
+		})
+		Expect(out).To(ContainSubstring("connected"))
+	})
+
+	It("health", func() {
```
hanahmily, line 50:
```suggestion
	It("should fail without the proper cert", func() {
```

hanahmily (changes requested):

hanahmily (approved):
@ButterBright Thank you for this contribution. 

Would you integrate the health command into the banyand docker image at `banyand/Dockerfile` in the follow-up PR? Then replace the healthcheck piece in `banyand`, `liaison` and `data` services in `test/docker/base-compose.yml` correspondingly.

ButterBright:
Sure.

PR #341, opened 2023-10-15
Implement local file system

HHoflittlefish777:
close https://github.com/apache/skywalking/issues/10839


pkg/fs/local_file_system.go
```diff
@@ -0,0 +1,408 @@
...
+			return rsize, &FileSystemError{
+				Code:    permissionError,
+				Message: fmt.Sprintf("There is not enough permission, file name: %s, error message: %s", iter.fileName, err),
+			}
+		} else {
+			return rsize, &FileSystemError{
+				Code:    readError,
+				Message: fmt.Sprintf("Read file error, file name: %s, read file size: %d, error message: %s", iter.fileName, rsize, err),
+			}
+		}
```
hanahmily, line 405: Duplicated code.

pkg/fs/local_file_system_test.go
```diff
@@ -0,0 +1,173 @@
...
+		dirName       string = "dir"
+		fileName      string = "dir/file"
+		flushFileName string = "dir/flushFile"
+	)
+
+	var fs FileSystem
+
+	ginkgo.Context("Local File", func() {
+		ginkgo.BeforeEach(func() {
+			fs = NewLocalFileSystem()
```
hanahmily, line 43: To prevent cluttering in the current directory, kindly execute the test within the `tmp` folder.

pkg/fs/local_file_system_test.go
```diff
@@ -0,0 +1,173 @@
...
+		})
+
+		ginkgo.AfterEach(func() {
+			err := os.RemoveAll(dirName)
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		})
+
+		ginkgo.It("File Operation", func() {
```
hanahmily, line 51: It will be better to create the case for each operation

pkg/fs/local_file_system.go
```diff
@@ -0,0 +1,285 @@
...
+}
+
+// NewLocalFileSystem is used to create the Local File system.
+func NewLocalFileSystem() FileSystem {
+	return &LocalFileSystem{
+		logger: logger.GetLogger(moduleName),
+	}
+}
+
+func readErrorHandle(err error, name string, size int) (int, error) {
```
hanahmily, line 47: I believe this method can handle all errors, but it would be to add a parameter indicating the operation where the error occurred.

hanahmily (changes requested):

hanahmily (approved):

PR #343, opened 2023-10-16
re: Unify topn&query plan

zesiar0:
related issue https://github.com/apache/skywalking/issues/10562.


banyand/measure/measure_query.go
```diff
@@ -54,10 +54,25 @@ type Measure interface {
 	GetSchema() *databasev1.Measure
 	GetIndexRules() []*databasev1.IndexRule
 	GetInterval() time.Duration
+	BuildSchema(other Measure) *databasev1.Measure
 }
 
 var _ Measure = (*measure)(nil)
 
+func (s *measure) BuildSchema(other Measure) *databasev1.Measure {
```
lujiajing1126, line 62: According to our last discussion, we should build topN schema with entity names,

https://github.com/apache/skywalking-banyandb/blob/d722db6bcd52554ef429961600d617d28d561ae5/banyand/measure/metadata.go#L204-L220

The actual entity names written for the TopN data point can be found,

https://github.com/apache/skywalking-banyandb/blob/d722db6bcd52554ef429961600d617d28d561ae5/banyand/measure/measure_topn.go#L233-L247

lujiajing1126 (changes requested):

banyand/measure/metadata.go
```diff
@@ -217,6 +217,7 @@ func createOrUpdateTopNMeasure(ctx context.Context, measureSchemaRegistry schema
 			},
 		},
 		Fields: []*databasev1.FieldSpec{TopNValueFieldSpec},
+		Entity: sourceMeasureSchema.GetEntity(),
```
lujiajing1126, line 220: Source measure entity has a offset of 3

pkg/query/logical/measure/topn_analyzer.go
```diff
@@ -0,0 +1,95 @@
...
+	return ms, nil
+}
+
+// TopNAnalyze converts logical expressions to executable operation tree represented by Plan.
+func TopNAnalyze(_ context.Context, criteria *measurev1.TopNRequest, schema *databasev1.Measure,
+	topNAggregation *databasev1.TopNAggregation, s logical.Schema,
+) (logical.Plan, error) {
+	groupByProjectionTags := schema.GetEntity().GetTagNames()
+	groupByTags := make([][]*logical.Tag, len(schema.GetTagFamilies()))
+	for i, tagFamily := range schema.GetTagFamilies() {
```
lujiajing1126, line 58: This is strange, why we need to assemble groupByTags for all tag families?

pkg/query/logical/measure/topn_plan_localscan.go
```diff
@@ -0,0 +1,254 @@
...
+		projectionFieldsRefs: projFieldRefs,
+		metadata:             uls.metadata,
+		entity:               entity,
+		l:                    logger.GetLogger("topn", "measure", uls.metadata.Group, uls.metadata.Name, "local-index"),
+	}, nil
+}
+
+func (uls *unresolvedLocalScan) locateEntity(entityList []string) (tsdb.Entity, error) {
+	entityMap := make(map[string]int)
+	entity := make([]tsdb.Entry, 1+1+len(entityList))
```
lujiajing1126, line 92: Then (continued...)

```go
entity := make([]tsdb.Entry, len(entityList))
for idx, tagName := range entityList {
    if tagName == "rank" {}
    else if tagName == "sort" {}
    else {
        entityMap[tagName] = idx
        // allow to make fuzzy search with partial conditions
        entity[idx] = tsdb.AnyEntry
    }
}
```

lujiajing1126 (approved):
LGTM

PR #351, opened 2023-11-05
feat: auto refresh time range

zesiar0:
- [x] Fixes apache/skywalking#11312.

https://github.com/apache/skywalking-banyandb/assets/59257677/61b6a41c-7fde-41df-967c-63942a3ede29

### DIFF
--- a/ui/src/components/Read/index.vue
+++ b/ui/src/components/Read/index.vue
@@ -32,6 +32,16 @@ const route = useRoute()
 
 const yamlRef = ref()
 
+const isDatePicker = ref(false)
+
+const last15Minutes = ref(900 * 1000)
+
+const lastWeek = ref(3600 * 1000 * 24 * 7)
+
+const lastMonth = ref(3600 * 1000 * 24 * 30)
+
+const last3Months = ref(3600 * 1000 * 24 * 90)
+
 // Loading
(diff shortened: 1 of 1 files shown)


zesiar0:
> Is it possible to add screen recording demonstration operation and console output
> 
> reference [apache/skywalking#11312](https://github.com/apache/skywalking/issues/11312)

zesiar0:
> Is it possible to add screen recording demonstration operation and console output
> 
> reference [apache/skywalking#11312](https://github.com/apache/skywalking/issues/11312)

Done.

zesiar0:
> Should display 'Last xxx' in `selected` instead of the specific time?
> 
> image

`el-date-picker` is a component in element-plus. I check the [official docs](https://element-plus.org/en-US/component/date-picker.html#date-formats) about `Date Picker`, and It seem that there is no avaliable [date format](https://day.js.org/docs/en/display/format#list-of-all-available-formats) which can satisfy your needs.

https://github.com/apache/skywalking-banyandb/blob/4e92c42997ee64e449bb86fd8478420eff20772b/ui/src/components/Read/index.vue#L405-L408

wu-sheng:
@hailin0 Once the time could be refreshed, and shown correctly, it should be good enough.

This would be only a bug fix, rather than some enhancement.

PR #352, opened 2023-11-06
ui: name search improvement on the measure page

LinuxSuRen:
Before:
image

After:
image

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#11502.

A changelog item might not be necessary due to too minor. But please let me know if you still belive it's required.

### DIFF
--- a/ui/src/components/Aside/index.vue
+++ b/ui/src/components/Aside/index.vue
@@ -248,13 +248,20 @@ function searchGroup() {
     let groupLists = []
     data.groupListsCopy.forEach(item => {
         let itemCache = JSON.parse(JSON.stringify(item))
+        let matched = false
         if (Array.isArray(itemCache.children)) {
             itemCache.children = itemCache.children.filter(child => {
                 return child.metadata.name.indexOf(data.search) > -1
             })
             if (itemCache.children.length > 0) {
                 groupLists.push(itemCache)
+                matched = true
             }
+        }
+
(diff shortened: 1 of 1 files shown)


ui/src/components/Aside/index.vue
```diff
@@ -248,7 +248,9 @@ function searchGroup() {
     let groupLists = []
     data.groupListsCopy.forEach(item => {
         let itemCache = JSON.parse(JSON.stringify(item))
-        if (Array.isArray(itemCache.children)) {
+        if (itemCache.metadata.name.indexOf(data.search) > -1) {
+            groupLists.push(itemCache)
```
hanahmily, line 252: It's better to display any matching group or item. Currently, the solution skips matched terms if the group is matched.

hanahmily (changes requested):
The change log is still required.

ui/src/components/Aside/index.vue
LinuxSuRen: Could be more specific? I might not get your point.

hanahmily: If any items under the group match the keyword, it should also be shown.

groupA:
  - itemA
  - itemB

When you issue "A", only groupA and itemA should show up instead of all of them.

hanahmily (approved):

PR #364, opened 2024-01-03
Implement measure merger

hanahmily:
The final task of measure column-based engine 

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#11472.


wu-sheng (approved):
Data merging and file changes should be documented as well.
Could you start an issue to list all documents to be changed later? I think we need to track and polish many docs for 0.6. @hanahmily

PR #371, opened 2024-01-13
Introduce compaction algorithm

lujiajing1126:


banyand/measure/merger_algo.go
```diff
@@ -0,0 +1,118 @@
...
+	"sort"
+)
+
+var (
+	_ MergePolicy = (*noopMergePolicy)(nil)
+	_ MergePolicy = (*lowestWriteAmplificationMergePolicy)(nil)
+)
+
+// MergePolicy is to select parts to be merged.
+type MergePolicy interface {
```
hanahmily, line 30: Why would you use an interface here? It's an internal component of the measure.

banyand/measure/merger_algo.go
```diff
@@ -0,0 +1,118 @@
...
+
+func (s *noopMergePolicy) GetPartsToMerge(src []*partWrapper, _ uint64) []*partWrapper {
+	return src
+}
+
+// lowestWriteAmplificationMergePolicy aims to choose an optimal combination
+// that has the lowest write amplification.
+// This policy can be referred to https://github.com/VictoriaMetrics/VictoriaMetrics/blob/v0.4.2-victorialogs/lib/logstorage/datadb.go
+// But parameters can be varied.
+type lowestWriteAmplificationMergePolicy struct {
```
hanahmily, line 45: The primary objective of all merge policies is to minimize write amplification. However, there may be better practices than naming the policy based on this objective.

banyand/measure/merger.go
```diff
@@ -179,15 +183,21 @@ var reservedDiskSpace uint64
 
 func (tst *tsTable) getPartsToMerge(snapshot *snapshot, maxFanOut uint64) ([]*partWrapper, map[uint64]struct{}) {
 	var parts []*partWrapper
-	toBeMerged := make(map[uint64]struct{})
+
+	maxBytesPerPart := maxFanOut / 2
```
hanahmily, line 187: We should parameterize it, taking free disk space into account if not set by the end user.

banyand/measure/merger_algo.go
```diff
@@ -0,0 +1,118 @@
...
+	maxParts           int
+	minMergeMultiplier float64
```
hanahmily, line 47: It seems a bit tricky to set it.

banyand/measure/merger_algo.go
```diff
@@ -0,0 +1,118 @@
...
+	_ MergePolicy = (*lowestWriteAmplificationMergePolicy)(nil)
+)
+
+// MergePolicy is to select parts to be merged.
+type MergePolicy interface {
+	GetPartsToMerge(src []*partWrapper, maxFanOut uint64) []*partWrapper
+}
+
+// noopMergePolicy aims to merge all parts.
+type noopMergePolicy struct{}
```
hanahmily, line 35: Are you planning to continue with this policy? It appears to be ineffective in merging parts in production env.

banyand/measure/merger_algo.go
lujiajing1126: I want the parameters to be adjusted both for testing and for users since due to our experience with VictoriaLogs, different workload/env need various parameter set.

banyand/measure/merger_algo.go
```diff
@@ -0,0 +1,133 @@
...
+import (
+	"math"
+	"sort"
+)
+
+// MergePolicy aims to choose an optimal combination
+// that has the lowest write amplification.
+// This policy can be referred to https://github.com/VictoriaMetrics/VictoriaMetrics/blob/v0.4.2-victorialogs/lib/logstorage/datadb.go
+// But parameters can be varied.
+type MergePolicy struct {
```
hanahmily, line 29:
```suggestion
// MergePolicy aims to choose an optimal combination
// that has the lowest write amplification.
type mergePolicy struct {
```

banyand/measure/merger_algo.go
```diff
@@ -0,0 +1,133 @@
...
+	maxParts           int
+	minMergeMultiplier float64
+	maxFanOutSize      uint64
+}
+
+// NewDefaultMergePolicy create a MergePolicy with default parameters.
+func NewDefaultMergePolicy() *MergePolicy {
```
hanahmily, line 36:
```suggestion
func newDefaultMergePolicy() *MergePolicy {
```

banyand/measure/merger_algo.go
```diff
@@ -0,0 +1,133 @@
...
+	return NewMergePolicy(15, 1.7, math.MaxUint64)
+}
+
+// NewMergePolicy creates a MergePolicy with given parameters.
+func NewMergePolicy(maxParts int, minMergeMul float64, maxFanOutSize uint64) *MergePolicy {
```
hanahmily, line 41:
```suggestion
func newMergePolicy(maxParts int, minMergeMul float64, maxFanOutSize uint64) *MergePolicy {
```

banyand/measure/measure.go
```diff
@@ -46,7 +46,8 @@ const (
 )
 
 type option struct {
-	flushTimeout time.Duration
+	flushTimeout  time.Duration
+	maxFanOutSize uint64
```
hanahmily, line 50:
```suggestion
	maxFanOutSize run.Bytes
```

banyand/measure/merger.go
```diff
@@ -179,15 +196,23 @@ var reservedDiskSpace uint64
...
+
 	for _, pw := range snapshot.parts {
-		if pw.mp != nil || pw.p.partMetadata.TotalCount < 1 || pw.p.partMetadata.CompressedSizeBytes > maxFanOut {
+		if pw.mp != nil || pw.p.partMetadata.TotalCount < 1 {
 			continue
 		}
 		parts = append(parts, pw)
+	}
+
+	tst.pwsChunk = mergerPolicy.getPartsToMerge(tst.pwsChunk[:0], parts, maxFanOut)
```
hanahmily, line 207: Could you use a local var instead of a field?

banyand/measure/service.go
```diff
@@ -98,6 +100,8 @@ func (s *service) Role() databasev1.Role {
 
 func (s *service) PreRun(_ context.Context) error {
 	s.l = logger.GetLogger(s.Name())
+	// set merge policy
+	customizeMergePolicy(withMaxFanOut(s.option.maxFanOutSize))
```
hanahmily, line 104: You could get the `maxFanoutSize` from the `tst.option.maxFanOutSize`. You don't have to set it to a global var.

banyand/measure/merger.go
```diff
@@ -30,6 +30,23 @@ import (
 	"github.com/apache/skywalking-banyandb/pkg/watcher"
 )
 
+// mergePolicy is a global policy for all merge works.
+var mergerPolicy = NewDefaultMergePolicy()
```
hanahmily, line 34: Please remove the global variable `mergerPolicy` and instead add it as a field of the `tstable`.

banyand/measure/merger_algo.go
```diff
@@ -0,0 +1,133 @@
+// Licensed to Apache Software Foundation (ASF) under one or more contributor
```
hanahmily, line 1: How about renaming the file to `merge_policy.go` since there is only a struct with the same name in it?

hanahmily (changes requested):

banyand/measure/merger.go
lujiajing1126: I think this could help reduce memory alloc

banyand/measure/merger.go
lujiajing1126: I cannot understand. The policy should be a global one right? We wouldn't have different policies for different tables?

banyand/measure/merger.go
```diff
@@ -179,15 +196,23 @@ var reservedDiskSpace uint64
...
-		if pw.mp != nil || pw.p.partMetadata.TotalCount < 1 || pw.p.partMetadata.CompressedSizeBytes > maxFanOut {
+		if pw.mp != nil || pw.p.partMetadata.TotalCount < 1 {
 			continue
 		}
 		parts = append(parts, pw)
+	}
+
+	tst.pwsChunk = mergerPolicy.getPartsToMerge(tst.pwsChunk[:0], parts, maxFanOut)
+	if len(tst.pwsChunk) == 0 {
+		tst.pwsChunk = append(tst.pwsChunk[:0], parts...)
```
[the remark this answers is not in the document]
lujiajing1126: Removed

banyand/measure/merger.go
hanahmily: The given slice consists of pointers. It is feasible to allocate it on the stack as the process is quick enough.

banyand/measure/tstable.go
```diff
@@ -114,9 +115,12 @@ type tsTable struct {
 	loopCloser    *run.Closer
 	p             common.Position
 	root          string
-	gc            garbageCleaner
-	curPartID     uint64
-	option        option
+	// pwsChunk is the memory buffer for parts to be merged
+	pwsChunk  []*partWrapper
```
hanahmily, line 119: As I mentioned previously, the field won't help reduce the memory.

banyand/measure/service.go
```diff
@@ -78,6 +79,8 @@ func (s *service) FlagSet() *run.FlagSet {
 	flagS := run.NewFlagSet("storage")
 	flagS.StringVar(&s.root, "measure-root-path", "/tmp", "the root path of database")
 	flagS.DurationVar(&s.option.flushTimeout, "measure-flush-timeout", defaultFlushTimeout, "the memory data timeout of measure")
+	s.option.maxFanOutSize = math.MaxInt64
```
hanahmily, line 82: Move the `maxFanoutSize` into `s`. Initialize the `mergePolicy` in `PreRun`. Finally, assign the policy to the `s.option`

banyand/measure/tstable.go
```diff
@@ -53,6 +53,7 @@ func newTSTable(fileSystem fs.FileSystem, rootPath string, p common.Position,
 		option:     option,
 		l:          l,
 		p:          p,
+		merger:     newDefaultMergePolicy(),
```
hanahmily, line 56: After the initialization process, you can get the mergePolicy from the option.

hanahmily (changes requested):

banyand/measure/service.go
lujiajing1126: Fixed. Pls check the new impl

hanahmily (approved):

PR #377, opened 2024-01-22
Stream column-based storage

ButterBright:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#11474.


wu-sheng:
Please fix the CI.

banyand/stream/iter_builder.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+var (
+	errUnspecifiedIndexType = errors.New("Unspecified index type")
+	rangeOpts               = index.RangeOpts{}
+)
+
+type filterFn func(item item) bool
+
+type seekerBuilder struct {
```
hanahmily, line 39: Since seekerBuilder currently only contains a single method, would you be willing to change it to a global method? This change would allow most of the relevant objects to be allocated on the stack, improving performance and making the code easier to manage.

banyand/stream/stream.go
```diff
@@ -21,25 +21,70 @@ package stream
...
+
+// Stream allows inspecting elements' details.
+type Stream interface {
+	io.Closer
+	GetSchema() *databasev1.Stream
+	GetIndexRules() []*databasev1.IndexRule
+	Query(ctx context.Context, opts pbv1.StreamQueryOptions) (pbv1.StreamQueryResult, error)
+	Sort(ctx context.Context, opts pbv1.StreamSortOptions) (pbv1.StreamSortResult, error)
+	Filter(ctx context.Context, opts pbv1.StreamFilterOptions) (pbv1.StreamFilterResult, error)
+}
```
hanahmily, line 75: Please move the methods related to queries to the `query.go` file.

hanahmily (changes requested):

hanahmily (approved):

PR #378, opened 2024-01-24
Fix stream startup warning

ButterBright:


banyand/stream/tstable.go
```diff
@@ -180,6 +180,9 @@ func newTSTable(fileSystem fs.FileSystem, rootPath string, p common.Position,
 	var needToDelete []string
 	for i := range ee {
 		if ee[i].IsDir() {
+			if ee[i].Name() == "element_idx" {
```
hanahmily, line 183: You should exclude directories that do not start with "part-". If any directories do not meet this criterion, the condition will not be satisfied.

hanahmily (changes requested):

hanahmily (approved):

wu-sheng:
CI fails. Please recheck.

hanahmily (approved):

PR #379, opened 2024-01-24
Introduce merge algorithm for stream

ButterBright:


hanahmily (approved):
LGTM if the CI passed

PR #381, opened 2024-01-24
Remove maxBlockLength in stream

ButterBright:


banyand/stream/part.go
```diff
@@ -325,8 +325,7 @@ func (mp *memPart) mustInitFromElements(es *elements) {
 			sidPrev = sid
 		}
 
-		if uncompressedBlockSizeBytes >= maxUncompressedBlockSize ||
-			(i-indexPrev) > maxBlockLength || sid != sidPrev {
+		if uncompressedBlockSizeBytes >= maxUncompressedBlockSize || sid != sidPrev {
```
hanahmily, line 328: You should remove "indexPrev" as well

banyand/stream/merger.go
```diff
@@ -300,20 +300,14 @@ func mergeBlocks(closeCh <-chan struct{}, bw *blockWriter, br *blockReader) (*pa
...
 			pendingBlock, tmpBlock = tmpBlock, pendingBlock
 			continue
 		}
 
-		if len(tmpBlock.timestamps) <= maxBlockLength {
-			bw.mustWriteBlock(tmpBlock.bm.seriesID, &tmpBlock.block)
-			releaseDecoder()
-			continue
-		}
-		tmpBlock.idx = maxBlockLength
```
hanahmily, line 316: You can't simply remove it.

When the tmpBlock's size is larger than maxUncompressedBlockSize, it should be written. The rest of the pieces in the loop will be unnecessary.

hanahmily (changes requested):

hanahmily (changes requested):

banyand/stream/part_metadata.go
```diff
@@ -32,8 +32,7 @@ type partMetadata struct {
 	BlocksCount           uint64 `json:"blocksCount"`
 	MinTimestamp          int64  `json:"minTimestamp"`
 	MaxTimestamp          int64  `json:"maxTimestamp"`
-	// TODO: remove this?
-	ID uint64 `json:"-"`
+	ID                    uint64 `json:"-"`
```
hanahmily, line 35: You should raise another PR to remove it and the relevant "partID" field since the stream doesn't have to duplicate data.

hanahmily (approved):

PR #384, opened 2024-01-27
feat: remove entity info from ".tf"&".tfm" file

zesiar0:
- [x] remove entity from ".tf"&".tfm" file, and find entity in series index

### DIFF
--- a/banyand/measure/query.go
+++ b/banyand/measure/query.go
@@ -420,7 +420,7 @@ func (qr *queryResult) Pull() *pbv1.MeasureResult {
 				qr.data = append(qr.data[:i], qr.data[i+1:]...)
 				i--
 			}
-			if qr.schema.GetEntity() == nil || len(qr.entities ) == 0 {
+			if qr.schema.GetEntity() == nil || len(qr.entities) == 0 {
 				continue
 			}
 			sidIndex := qr.sidToIndex[qr.data[i].bm.seriesID]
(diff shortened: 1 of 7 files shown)


banyand/measure/query.go
```diff
@@ -141,13 +175,59 @@ func (s *measure) Query(ctx context.Context, mqo pbv1.MeasureQueryOptions) (pbv1
...
+			}
+			values = append(values, value)
+		}
+	case databasev1.TagType_TAG_TYPE_STRING_ARRAY:
+		if tagValue.GetStrArray() == nil {
+			return values
+		}
+		for i := 0; i < num; i++ {
+			var value string
+			for j := range tagValue.GetStrArray().Value {
```
hanahmily, line 220: The encoding method is incorrect.

According to `block.mustInitFromTagsAndFields`, the encoding process should contain two steps:

1. Convert `tagValue` to `nameValue` using `encodeTagValue` laid in `write.go`
2. `nameValue.marshal` could serialize the value to a byte slice.

banyand/measure/query.go
```diff
@@ -128,6 +140,28 @@ func (s *measure) Query(ctx context.Context, mqo pbv1.MeasureQueryOptions) (pbv1
 	if tstIter.Error() != nil {
 		return nil, fmt.Errorf("cannot iterate tstIter: %w", tstIter.Error())
 	}
+
+	entities := make([]string, 0)
+	if s.GetSchema().GetTagFamilies()[0].GetName() == topNTagFamily {
+		for _, topNAggregation := range s.topNAggregations {
+			if topNAggregation.GetMetadata().GetName() == s.schema.GetMetadata().GetName() {
+				entities = append(entities, "sortDirection", "rankNumber")
```
hanahmily, line 148: The entity used to define the topN result measure lacks tags, causing a mismatch between writing and reading. https://github.com/apache/skywalking-banyandb/blob/main/banyand/measure/metadata.go#L214 is relevant. 

Please correct this issue, please prepend "sortDirection" and "rankNumber" to the above Entity field.

hanahmily (changes requested):

hanahmily (approved):
LGTM

PR #386, opened 2024-02-03
get entity value from series index when query stream data

zesiar0:
- [x] correct encode tag value method
- [x] when query stream data, get entity value from series index

### DIFF
--- a/banyand/measure/query.go
+++ b/banyand/measure/query.go
@@ -18,6 +18,7 @@
 package measure
 
 import (
+	"bytes"
 	"container/heap"
 	"context"
 	"fmt"
@@ -33,6 +34,7 @@ import (
 	"github.com/apache/skywalking-banyandb/banyand/internal/storage"
 	"github.com/apache/skywalking-banyandb/pkg/convert"
 	"github.com/apache/skywalking-banyandb/pkg/logger"
+	"github.com/apache/skywalking-banyandb/pkg/partition"
 	pbv1 "github.com/apache/skywalking-banyandb/pkg/pb/v1"
 	resourceSchema "github.com/apache/skywalking-banyandb/pkg/schema"
 )
@@ -78,11 +80,21 @@ func (s *measure) Query(ctx context.Context, mqo pbv1.MeasureQueryOptions) (pbv1
 			tabWrappers[i].DecRef()
 		}
 	}()
+
+	tagNameIndex := make(map[string]partition.TagLocator)
+	tagFamilySpecs := s.schema.GetTagFamilies()
+	for i, tagFamilySpec := range tagFamilySpecs {
+		for j, tagSpec := range tagFamilySpec.GetTags() {
+			tagNameIndex[tagSpec.GetName()] = partition.TagLocator{
+				FamilyOffset: i,
+				TagOffset:    j,
+			}
+		}
+	}
 	sl, err := tsdb.IndexDB().Search(ctx, &pbv1.Series{Subject: mqo.Name, EntityValues: mqo.Entity}, mqo.Filter, mqo.Order)
 	if err != nil {
 		return nil, err
 	}
-
 	var result queryResult
 	if len(sl) < 1 {
 		return &result, nil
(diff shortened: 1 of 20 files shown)


banyand/measure/query.go
```diff
@@ -393,6 +393,9 @@ func (qr *queryResult) Pull() *pbv1.MeasureResult {
 					offset := qr.tagNameIndex[tagProj]
 					tagFamilySpec := qr.schema.GetTagFamilies()[offset.FamilyOffset]
 					tagSpec := tagFamilySpec.GetTags()[offset.TagOffset]
+					if tagSpec.IndexedOnly {
+						continue
+					}
```
hanahmily, line 398: It would be better for you to update write.go instead of query.go. "IndexOnly" means the tag is not stored in the file.

banyand/measure/write.go
```diff
@@ -148,6 +149,14 @@ func (w *writeCallback) handle(dst map[string]*dataPointsInGroup, writeEvent *me
 			} else {
 				tagValue = tagFamily.Tags[j]
 			}
+			tagFamiliesForIndexWrite[i].values = append(tagFamiliesForIndexWrite[i].values, encodeTagValue(
```
hanahmily, line 152: It is better to merge the encoding operation in lines 152 and 160 and encode the tag value only once.

banyand/stream/query.go
```diff
@@ -41,6 +43,15 @@ type queryOptions struct {
 	maxTimestamp int64
 }
 
+func mustEncodeTagValue(name string, tagType databasev1.TagType, tagValue *modelv1.TagValue, num int) [][]byte {
+	values := make([][]byte, num)
+	nv := encodeTagValue(name, tagType, tagValue)
+	for i := 0; i < num; i++ {
+		values[i] = nv.marshal()
```
hanahmily, line 50: Moving the marshaling process outside the loop will help improve performance and efficiency.

banyand/stream/query.go
```diff
@@ -342,6 +437,33 @@ func (s *stream) Sort(ctx context.Context, sso pbv1.StreamSortOptions) (ssr pbv1
...
+				FamilyOffset: i,
+				TagOffset:    j,
+			}
+		}
+	}
+	sidToIndex := make(map[common.SeriesID]int)
+	for idx, series := range seriesList {
+		sidToIndex[series.ID] = idx
+	}
+
```
hanahmily, line 466: Duplicated codes. It's better to abstract a function to generate them.

banyand/stream/query.go
```diff
@@ -161,6 +175,46 @@ func (qr *queryResult) Pull() *pbv1.StreamResult {
 				qr.data = append(qr.data[:i], qr.data[i+1:]...)
 				i--
 			}
+			if qr.schema.GetEntity() == nil || len(qr.schema.GetEntity().GetTagNames()) == 0 {
+				continue
+			}
+			sidIndex := qr.sidToIndex[qr.data[i].bm.seriesID]
+			series := qr.seriesList[sidIndex]
+			entityMap := make(map[string]int)
```
hanahmily, line 183: It's better to move it in queryResult and fill it in Query function.

banyand/stream/write.go
```diff
@@ -134,7 +139,7 @@ func (w *writeCallback) handle(dst map[string]*elementsInGroup, writeEvent *stre
 				tagFamilySpec.Tags[j].Type,
 				tagValue,
 			))
-			if tagFamilySpec.Tags[j].IndexedOnly {
+			if tagFamilySpec.Tags[j].IndexedOnly || entityMap[tagFamilySpec.Tags[j].Name] {
 				continue
 			}
 			tagFamilies[i].values = append(tagFamilies[i].values, encodeTagValue(
```
hanahmily, line 145: As I mentioned in the measure's "write.go", merging encoding tag operations is better.

hanahmily (changes requested):
Some nits

hanahmily (approved):
Nice work. This effort will reduce most of volumn overhead of the stream.

PR #389, opened 2024-02-05
Add lock and version file for process control and compatibility

ButterBright:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#11493.

### DIFF
--- a/banyand/internal/storage/segment.go
+++ b/banyand/internal/storage/segment.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"path"
+	"path/filepath"
 	"sort"
 	"strconv"
 	"sync"
@@ -30,8 +31,14 @@ import (
 
 	"github.com/apache/skywalking-banyandb/api/common"
 	"github.com/apache/skywalking-banyandb/banyand/tsdb/bucket"
+	"github.com/apache/skywalking-banyandb/pkg/fs"
 	"github.com/apache/skywalking-banyandb/pkg/logger"
 	"github.com/apache/skywalking-banyandb/pkg/timestamp"
+)
+
+const (
+	versionFilename = "version"
+	currentVersion  = "0.1"
 )
 
 var errEndOfSegment = errors.New("reached the end of the segment")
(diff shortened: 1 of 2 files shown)


banyand/internal/storage/tsdb.go
```diff
@@ -78,6 +86,12 @@ func (d *database[T, O]) Close() error {
 	for _, s := range d.sLst {
 		s.close()
 	}
+	if d.lock != nil {
```
hanahmily, line 89: lock file can't be nil

banyand/internal/storage/tsdb.go
```diff
@@ -105,6 +119,13 @@ func OpenTSDB[T TSTable, O any](ctx context.Context, opts TSDBOpts[T, O]) (TSDB[
 		p:        p,
 	}
 	db.logger.Info().Str("path", opts.Location).Msg("initialized")
+	db.fileSystem = fs.NewLocalFileSystem()
```
hanahmily, line 122: Please use `lfs` which is a global var

banyand/internal/storage/segment.go
```diff
@@ -30,10 +31,16 @@ import (
...
 	"github.com/apache/skywalking-banyandb/banyand/tsdb/bucket"
+	"github.com/apache/skywalking-banyandb/pkg/fs"
 	"github.com/apache/skywalking-banyandb/pkg/logger"
 	"github.com/apache/skywalking-banyandb/pkg/timestamp"
 )
 
+const (
+	versionFilename = "version"
+	currentVersion  = "0.1"
+)
```
hanahmily, line 42: Please move them to a separate file called "version.go"

banyand/internal/storage/segment.go
```diff
@@ -160,6 +167,20 @@ func (sc *segmentController[T, O]) selectTSTables(timeRange timestamp.TimeRange)
 }
 
 func (sc *segmentController[T, O]) createTSTable(ts time.Time) (TSTableWrapper[T], error) {
+	data := []byte(currentVersion)
+	versionPath := filepath.Join(sc.location, versionFilename)
+	fileSystem := fs.NewLocalFileSystem()
```
hanahmily, line 172: Please use "lfs" instead of a local var.

banyand/internal/storage/segment.go
```diff
@@ -160,6 +167,20 @@ func (sc *segmentController[T, O]) selectTSTables(timeRange timestamp.TimeRange)
...
+	lf, err := fileSystem.CreateLockFile(versionPath, filePermission)
```
hanahmily, line 173: If the version file exists, the segment controller should check whether the version is compatible.

hanahmily (changes requested):

banyand/version/version.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+// ErrVersionIncompatible is returned when versions are not compatible.
+var ErrVersionIncompatible = errors.New("version not compatible")
+
+//go:embed versions.yml
+var versionFS embed.FS
+
+// ReadCompatibleVersions reads compatible versions from versions.yml.
+func ReadCompatibleVersions() (map[string][]float64, error) {
```
hanahmily, line 46: I want to move versioned files into storage and make them private.

hanahmily (changes requested):

banyand/version/version.go
```diff
@@ -0,0 +1,60 @@
...
+	"errors"
+
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	// VersionFilename represents the name of the version file.
+	VersionFilename = "version"
+	// CurrentVersion represents the current version number.
+	CurrentVersion = "0.1"
```
hanahmily, line 33: Please use the SemVer (Major.Minor.Patch) pattern, starting with "1.0.0".

hanahmily (changes requested):

hanahmily (approved):

PR #396, opened 2024-03-04
Fix 11273 remove exception log is output in console

sksDonni:
### Fix [banyandb web-ui console error logs](https://github.com/apache/skywalking/issues/11273)
Experiencing exception logs in the console of the banyandb dashboard when we switch between two read tabs. 

- [X] Explain briefly why the bug exists and how to fix it.
The issue was caused due to the use of `JSON.parse(JSON.stringify())` to clone the route object. This method of cloning objects cannot handle circular references, leading to an exception in the console.

- [X] If it's UI related, attach the screenshots below.

https://github.com/apache/skywalking-banyandb/assets/53624363/b62bba6e-2678-4e8a-9b3e-4436cd0f5649

- [X] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#11273.

### DIFF
--- a/ui/src/components/Read/index.vue
+++ b/ui/src/components/Read/index.vue
@@ -191,7 +191,8 @@ watch(() => data.code, () => {
         data.codeStorage[index].params.code = data.code
     } else {
         route.params.code = data.code
-        data.codeStorage.push(JSON.parse(JSON.stringify(route)))
+        const clonedObject = clonedObject(route)
+        data.codeStorage.push(clonedObject)
     }
 })
 function initCode() {
@@ -216,6 +217,28 @@ orderBy:
   sort: SORT_UNSPECIFIED
 `)
     }
+}
+function cloneObject(obj){
+    if (obj === null || typeof obj !== 'object') {
+        return obj;
+    }
+
+    if (Array.isArray(obj)) {
+        const arrCopy = [];
+        for (let i = 0; i < obj.length; i++) {
+        arrCopy[i] = cloneObject(obj[i]);
+        }
+        return arrCopy;
+    }
+
+    const objCopy = {};
+    for (const key in obj) {
+        if (obj.hasOwnProperty !== undefined && obj.hasOwnProperty(key)) {
+            objCopy[key] = cloneObject(obj[key]);
+        }
+    }
+
+    return objCopy;
 }
 function changeCode(name, value) {
     let code = yamlToJson(data.code).data


ui/src/components/Read/index.vue
```diff
@@ -184,8 +184,10 @@ watch(() => data.code, () => {
     if (index >= 0) {
         data.codeStorage[index].params.code = data.code
     } else {
+
         route.params.code = data.code
-        data.codeStorage.push(JSON.parse(JSON.stringify(route)))
+        const clonedObject = Object.assign({}, route);
```
hanahmily, line 189: After researching, I found that this code only creates a shallow copy, not a deep copy, which could lead to potential issues. Can you take a look, @Fine0830?

hanahmily: https://github.com/WebReflection/flatted might help with this issue.

Fine0830: The `Object. assign` can only deeply copy the first layer of json.

wfxSuchith: sure @hanahmily and @Fine0830, will look into this and update the PR accordingly.

sksDonni: Hi @Fine0830 @hanahmily, tried using structuredClone(). Am encountering some issue similar to this [here](https://stackoverflow.com/questions/72632173/unable-to-use-structuredclone-on-value-of-ref-variable). Will try to manually loop and clone each child recursively.

Fine0830: Could you use `toRow` before using `structuredClone`?

hanahmily: Did you try using "toRaw" as suggested in the link you shared? This solution may help resolve the problem you encountered.

Fine0830: Probably, it is enough to use `toRaw` here. Transforming a responsive object generated with reactive into a non responsive regular object, then assign a value to the new variable. All operations on this object will not cause page updates.

sksDonni: @Fine0830 @hanahmily, Using toRaw is still causing the issue to persist. Will see what else can be done. Will revert to a recursive solution if nothing works out.

Fine0830: How do you use `toRaw`, Only toRaw(route)? If so, it is not enough. All data in the `route` need to be converted.

sksDonni: Yeah, tried using toRaw(route), toRaw(route.data) and toRaw(route.value). Nothing seemed to work for me

hanahmily: You could try JSON.parse (JSON. stringify()). If it doesn't work, writing a custom function might be the only solution.

sksDonni:
Had closed this pull request accidentally. Requesting second review.

Fine0830:
While debugging this code, I noticed that it wasn't executing in development mode and encountered errors in production mode, without affecting the page interaction. As a result, I concluded that this segment of code was invalid and proceeded to delete it.

wu-sheng:
@Fine0830 Will you use another PR to remove the codes?

Fine0830:
I have removed the codes in https://github.com/apache/skywalking-banyandb/pull/408

PR #397, opened 2024-03-05
[UI] Default Time Range for Querying Data to Last 15 Minutes. 

Sylvie-Wxr:
### <Feature description>
- [x] If it's UI related, attach the screenshots below.
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#11973.

### Feature Description: 
- When selecting a stream or measure, the time range with be auto set to last 15 minutes.  
image

- If the user did not maunally select any time range and clicked search, the time range will be auto reset to last 15 minutes 
image

- If the user selected shortcut timeranges (last xxx) and clicked search, the time range will be auto reset to last xxx. 
image

- If the user manually handpicked time range  and clicked search, the time range will not be auto reset 
image

### Video Demo
https://github.com/apache/skywalking-banyandb/assets/129717259/ffe2cf8e-2bbb-4131-99c9-8e8deff3efd3

### Implementation Explannation:
1. Added ` :key="$route.path"` in `RouterView` to trigger an router-view reinitialization everytime an read editor is selected.  
2. Fix code bug in `initCode` function where the startTime is set wrongly to -15 microseconds. 
3. Remove redundant interval logics.


ui/src/components/Read/index.vue
```diff
@@ -347,27 +344,25 @@ function handleCodeData() {
     getTableData()
 }
 function autoRefreshTimeRange() {
-    if (isDatePicker.value) {
```
hanahmily, line 350: This variable is still useful, but the original version still has some flaws. Let me clarify the logic.

### "last 15 minutes" is selected. 
1. If user click "🔍" , both the start and end dates have been updated to reflect the current time.
2. If user click "↻", the dates remain unchanged.

### A fixed time range is selected 
1. If user click "🔍" , the dates remain unchanged.
2. If user click "↻", the dates remain unchanged.

Based on the four cases mentioned above, it appears that "autoRefreshTimeRange" feature has limited functionality and works only in one scenario.

ui/src/components/Read/index.vue
```diff
@@ -197,11 +195,10 @@ function initCode() {
     } else {
         let timeRange = {
             timeRange: {
-                begin: new Date(new Date() - 15),
+                begin: new Date(new Date() - 15 * 60 * 1000),
```
hanahmily, line 198: This is a good find.

Since we have a constant variable named 'last15Minutes' defined at line 35, would it be possible to reuse it instead of generating it repeatedly? I understand that this constant is a "ref", but it would be better to use a numeric constant instead and refer to it in the two places it is used.

hanahmily (changes requested):

ui/src/components/Read/index.vue
Sylvie-Wxr: The "ref" here means reactive reference. When the value of the ref changes, any Vue components that depend on this value will automatically re-render to reflect the new value. 

I think we should not use "ref" here for these time variables (last15Minutes, lastMonth, ...).  
We should just change to `const last15Minutes = 900 * 1000`.

Sylvie-Wxr: Changed accordingly.

ui/src/components/Read/index.vue
hanahmily: If a fixed time range is selected, the start and end should remain unchanged whenever you click "🔍" or "↻". That's why you have to use a variable to mark this status.

The fixed time range allows users to select specific data within a certain time range without having to change the time when clicking searching or refreshing buttons.

ui/src/components/Read/index.vue
```diff
@@ -347,27 +344,25 @@ function handleCodeData() {
...
-        json.data.timeRange.end = end.toISOString()
-        data.code = jsonToYaml(json.data).data
+    if (interval === last15Minutes) {
+        begin.setTime(begin.getTime() - last15Minutes)
+    } else if (interval === lastWeek) {
+        begin.setTime(begin.getTime() - lastWeek)
+    } else if (interval === lastMonth) {
+        begin.setTime(begin.getTime() - lastMonth)
+    } else if (interval === last3Months) {
+        begin.setTime(begin.getTime() - last3Months)
```
hanahmily, line 361: There is a bug here

The "last xx" dynamic time range type cannot be determined by the interval. If a user selects a fixed time range such as "last 15 minutes", the component will not update the time range upon clicking the search button.

hanahmily (changes requested):

ui/src/components/Read/index.vue
Sylvie-Wxr: Changed accordingly and upated the description. Please verify the new video demo.

ui/src/components/Read/index.vue
Sylvie-Wxr: Removed these code.

PR #406, opened 2024-03-09
chore: imporve wal and remove some comment

caicancai:


caicancai:
I think you can add comments to Options instead of stating how many KB/MB it is in each test. This is not necessary.

hanahmily:
@caicancai The "wal", "banyand/kv" and "banyand/tsdb" will be deprecated in version 0.6.0. Please stop improving them.

PR #409, opened 2024-03-11
Implement GC for SeriesDatabase of TSDB

zesiar0:
https://github.com/apache/skywalking/issues/11494


wu-sheng:
Please fix the CI.

banyand/internal/storage/index.go
```diff
@@ -224,3 +229,84 @@ func appendSeriesList(dest, src pbv1.SeriesList, filter posting.List) pbv1.Serie
...
+	}
+	sir.hot = sir.standby
+	location := filepath.Clean(sir.opts.Location)
+	path := fmt.Sprintf("idx-%v", fastrand.Uint32())
+	sir.standby, err = newSeriesIndex(sir.ctx, filepath.Join(location, path), sir.opts.SeriesIndexFlushTimeoutSeconds)
+	if err != nil {
+		l.Error().Msg("fail to create series index")
+		return false
+	}
+	return true
```
hanahmily, line 295: The current execution order is incorrect. The correct order should be as follows:

1. Swap the hot and standby, and ensure that an RWLock is used to protect the swap.
2. Wait for an hour, then shut down the new standby (which was the previous hot).
3. If the current time is less than an hour away from the next TTL, create a new standby.

banyand/internal/storage/index.go
```diff
@@ -224,3 +229,84 @@ func appendSeriesList(dest, src pbv1.SeriesList, filter posting.List) pbv1.Serie
...
+}
+
+func (sir *seriesIndexController[T, O]) Write(docs index.Documents) error {
+	return sir.hot.Write(docs)
+}
+
+func (sir *seriesIndexController[T, O]) searchPrimary(ctx context.Context, series *pbv1.Series) (pbv1.SeriesList, error) {
+	return sir.hot.searchPrimary(ctx, series)
```
hanahmily, line 303: If the hot's search result is empty and the standby is not closed, searchPrimary might also try to search in standby.

banyand/internal/storage/index.go
```diff
@@ -224,3 +229,84 @@ func appendSeriesList(dest, src pbv1.SeriesList, filter posting.List) pbv1.Serie
...
+}
+
+func (sir *seriesIndexController[T, O]) Search(ctx context.Context, series *pbv1.Series, filter index.Filter, order *pbv1.OrderBy) (pbv1.SeriesList, error) {
+	return sir.hot.Search(ctx, series, filter, order)
```
hanahmily, line 307: Same as "primarySearch"

banyand/internal/storage/index.go
```diff
@@ -224,3 +228,115 @@ func appendSeriesList(dest, src pbv1.SeriesList, filter posting.List) pbv1.Serie
...
+	opts TSDBOpts[T, O],
+) (*seriesIndexController[T, O], error) {
+	location := filepath.Clean(opts.Location)
+	path := fmt.Sprintf("idx-%v", time.Now().UnixNano())
+	h, err := newSeriesIndex(ctx, filepath.Join(location, path), opts.SeriesIndexFlushTimeoutSeconds)
+	if err != nil {
+		return nil, err
+	}
+	path = fmt.Sprintf("idx-%v", time.Now().UnixNano())
+	sb, err := newSeriesIndex(ctx, filepath.Join(location, path), opts.SeriesIndexFlushTimeoutSeconds)
```
hanahmily, line 263: standby will be created when the time is 1 hour before the deadline.

banyand/internal/storage/index.go
```diff
@@ -224,3 +228,115 @@ func appendSeriesList(dest, src pbv1.SeriesList, filter posting.List) pbv1.Serie
...
+	panic("invalid interval unit")
+}
+
+func newSeriesIndexController[T TSTable, O any](
+	ctx context.Context,
+	opts TSDBOpts[T, O],
+) (*seriesIndexController[T, O], error) {
+	location := filepath.Clean(opts.Location)
+	path := fmt.Sprintf("idx-%v", time.Now().UnixNano())
+	h, err := newSeriesIndex(ctx, filepath.Join(location, path), opts.SeriesIndexFlushTimeoutSeconds)
```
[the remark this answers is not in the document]
hanahmily: The query-on-disk test case failed due to this reason.

banyand/internal/storage/index.go
```diff
@@ -224,3 +228,115 @@ func appendSeriesList(dest, src pbv1.SeriesList, filter posting.List) pbv1.Serie
...
+	}
+	panic("invalid interval unit")
+}
+
+func newSeriesIndexController[T TSTable, O any](
+	ctx context.Context,
+	opts TSDBOpts[T, O],
+) (*seriesIndexController[T, O], error) {
+	location := filepath.Clean(opts.Location)
+	path := fmt.Sprintf("idx-%v", time.Now().UnixNano())
```
hanahmily, line 257: Please use hex representation by "fmt.Sprintf("%016x", time.Now().UnixNano())"

banyand/internal/storage/index.go
```diff
@@ -224,3 +228,115 @@ func appendSeriesList(dest, src pbv1.SeriesList, filter posting.List) pbv1.Serie
 func (s *seriesIndex) Close() error {
 	return s.store.Close()
 }
+
+type seriesIndexController[T TSTable, O any] struct {
+	ctx     context.Context
+	hot     *seriesIndex
+	standby *seriesIndex
+	timestamp.TimeRange
```
hanahmily, line 236: What's the time range used for?

banyand/internal/storage/index.go
```diff
@@ -224,3 +228,115 @@ func appendSeriesList(dest, src pbv1.SeriesList, filter posting.List) pbv1.Serie
...
+	sir.Lock()
+	defer sir.Unlock()
+	return sir.hot.Write(docs)
+}
+
+func (sir *seriesIndexController[T, O]) searchPrimary(ctx context.Context, series *pbv1.Series) (pbv1.SeriesList, error) {
+	sir.RLock()
+	defer sir.RUnlock()
+
+	sl, _ := sir.hot.searchPrimary(ctx, series)
```
hanahmily, line 318: Please check the error

banyand/internal/storage/index.go
```diff
@@ -224,3 +228,115 @@ func appendSeriesList(dest, src pbv1.SeriesList, filter posting.List) pbv1.Serie
...
+	if sl != nil {
+		return sl, nil
+	}
+
+	return sir.standby.searchPrimary(ctx, series)
```
hanahmily, line 323: the standby might be nil. Please check whether it's nil before querying.

banyand/internal/storage/index.go
```diff
@@ -224,3 +228,115 @@ func appendSeriesList(dest, src pbv1.SeriesList, filter posting.List) pbv1.Serie
...
+
+func (sir *seriesIndexController[T, O]) Search(ctx context.Context, series *pbv1.Series, filter index.Filter, order *pbv1.OrderBy) (pbv1.SeriesList, error) {
+	sir.RLock()
+	defer sir.RUnlock()
+
+	sl, _ := sir.hot.Search(ctx, series, filter, order)
+	if sl != nil {
+		return sl, nil
+	}
+	return sir.hot.Search(ctx, series, filter, order)
```
hanahmily, line 334: It should be standby

hanahmily (changes requested):

banyand/internal/storage/index.go
zesiar0: it's used to mark the creation time of series index and closing time which reached the TTL.  When triggering the scheduler, it will compare      the end time of seriesIndexController and deadline. And if the end time is before deadline, it will switch standby and hot index, otherwise create a new series index

hanahmily: You can remove the time range and leave only the end time.

banyand/internal/storage/index.go
```diff
@@ -224,3 +228,115 @@ func appendSeriesList(dest, src pbv1.SeriesList, filter posting.List) pbv1.Serie
...
+
+		startTime := standard(time.Now(), sir.opts.TTL.Unit)
+		endTime := startTime.Add(sir.opts.TTL.estimatedDuration())
+		sir.TimeRange = timestamp.NewSectionTimeRange(startTime, endTime)
+	}
+	if sir.End.Sub(deadline) < time.Hour {
+		location := filepath.Clean(sir.opts.Location)
+		path := fmt.Sprintf("idx-%v", time.Now().UnixNano())
+		sir.standby, err = newSeriesIndex(sir.ctx, filepath.Join(location, path), sir.opts.SeriesIndexFlushTimeoutSeconds)
+	}
```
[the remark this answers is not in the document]
hanahmily: You can remove the time range and leave only the end time.

PR #414, opened 2024-03-14
Feature to view and download value of property from UI

sksDonni:
Fix https://github.com/apache/skywalking/issues/11974
Large amount of text to display in property value 

- [X] Explain briefly why the bug exists and how to fix it.
The issue was caused due to the use of `property` value being very long in some cases resulting in bad UX. The fix ellipsizes the content of the value beyond 20 characters and provides a new popup to view and download the value of a property.

- [X] If it's UI related, attach the screenshots below.

https://github.com/apache/skywalking-banyandb/assets/53624363/dff66b02-b8a6-4c93-95b9-3f794a028601

### DIFF
--- a/ui/src/components/Property/PropertyValueReader.vue
+++ b/ui/src/components/Property/PropertyValueReader.vue
@@ -1,0 +1,87 @@
+<!--
+  ~ Licensed to Apache Software Foundation (ASF) under one or more contributor
+  ~ license agreements. See the NOTICE file distributed with
+  ~ this work for additional information regarding copyright
+  ~ ownership. Apache Software Foundation (ASF) licenses this file to you under
+  ~ the Apache License, Version 2.0 (the "License"); you may
+  ~ not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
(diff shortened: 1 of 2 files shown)


ui/src/components/Property/PropertyValueReader.vue
```diff
@@ -0,0 +1,87 @@
...
+const downloadValue = () => {
+    const dataBlob = new Blob([valueData.formattedData], { type: 'text/JSON' })
+    var a = document.createElement('a');
+
+    a.download = 'value.txt';
+    a.href = URL.createObjectURL(dataBlob);
+    document.body.appendChild(a);
+
+    a.click();
+    document.body.removeChild(a);
```
Fine0830, line 53: The code style should be consistent.

Fine0830 (dismissed):
LGTM. Thanks @sksDonni

ui/src/components/Property/PropertyValueReader.vue
wu-sheng: Let's fix this before merging.

wu-sheng:
@hanahmily I noticed, the approval is not dismissed when new changes are applied.
You should update .asf.yaml

Fine0830 (approved):

PR #422, opened 2024-03-23
Add trace stress test suit

hanahmily:
Add tracee stress test suite and fix several flaws found by the suite: 

1. Create a deep copy of the tag family map data.
2. Decrease snapshot references in the get element query function.


wu-sheng:
Please fix CI.

wu-sheng (approved):

PR #423, opened 2024-03-29
chore: add quick_start docs

caicancai:
I don’t think my change is reasonable, but I think the readme should add a quick_start so that users can know how to use skywalking-banyandb faster. I didn't find it at first

Or I should add a quick_start document. I haven’t found the corresponding information in the document so far.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ The database research community usually uses [RUM Conjecture](http://daslab.seas
 - [Dev version doc](https://skywalking.apache.org/docs/skywalking-banyandb/next/readme/)
 - [Latest release doc](https://skywalking.apache.org/docs/skywalking-banyandb/latest/readme/)
 - [Java Client SDK doc](https://skywalking.apache.org/docs/#BanyanDBJavaClient)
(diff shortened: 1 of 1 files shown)


wu-sheng:
Next is not a stable doc. Latest is.

wu-sheng:
0.6 will be the first recommend to use version.
It is better you could wait for 0.6 doc update(TBD) after we have stable benchmark and code base.

caicancai:
Thank you for your reply. closing

caicancai:
@wu-sheng Strictly speaking, I am not a user. I am mainly interested in this project. I feel that there are many areas for improvement in this project. This is a good learning opportunity for me. I think I should focus on the code. But seeing that some documents are not perfect, I might still want to improve them, thank you.

wu-sheng:
BanyanDB is at very early stage, it is true a lot of things are not well documented. Because majorly, we are not focusing on that.
As I mentioned, 0.6 is first version for public, main concepts are first time fully implemented.

If you want to join codes development, @hanahmily in the slack, you could talk with him.

PR #436, opened 2024-04-24
Fix flaws found by stream load test

hanahmily:
Fix several issues:

1. If no index data is generated, memory leaks can occur, causing the index writer to block the entire writing process.
2. Approximately 50% of the space can be saved by removing the term value from the inverted index.
3. The stream previously did not scan all block metadata when retrieving a single element from part. This has been fixed.
4. In the stream query process, instead of scanning each series separately, a single query sorts data.
5. Empty shards have been removed.


wu-sheng (approved):

wu-sheng:
e2e fails, please fix it.

wu-sheng (approved):

PR #437, opened 2024-04-24
Polish block docs

wu-sheng:

### DIFF
--- a/docs/concept/tsdb.md
+++ b/docs/concept/tsdb.md
@@ -33,7 +33,8 @@ Notably, for data of the `Stream` type, since there are no field columns, the `f
 
 ## Block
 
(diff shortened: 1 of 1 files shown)


docs/concept/tsdb.md
```diff
@@ -33,7 +33,9 @@ Notably, for data of the `Stream` type, since there are no field columns, the `f
 
 ## Block
 
-The diagram below shows the detailed fields within each block. The block is the minimal unit of tsdb, which contains several rows of data. Due to the column-based design, each block is spread over several files.
+Each block holds data with the same series ID. 
+The max size of the measure block is controlled by data volume and the number of rows. Meanwhile, the max size of the stream block is controlled by data volume.
```
Superskyyy, line 37: Just curious is the data volume and number of rows adjustable through config?

wu-sheng: I think they are not able to do for now.

hanahmily: No, I do not want to expose it to end users who do not have experience configuring it. However, dedicated users or contributors can access it by reading the code.

PR #439, opened 2024-04-27
Fix the wrong server name

hanahmily:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#12090.


wu-sheng (approved):

wu-sheng:
There seems to have a deadlink to the doc. Please recheck

wu-sheng (approved):

PR #440, opened 2024-04-28
Improve filtering performance of Stream

ButterBright:
The specification of the test environment is:
OS: Ubuntu 22.04.3 LTS x86_64
CPU: AMD EPYC 7282 (4) @ 2.794GHz
Memory: 7951MiB

The parameters for the benchmark test are:
{batchCount: 2, timestampCount: 500, seriesCount: 100, tagCardinality: 10, startTimestamp: 1, endTimestamp: 1000, scenario: "large-scale"}
{batchCount: 2, timestampCount: 500, seriesCount: 100, tagCardinality: 10, startTimestamp: 900, endTimestamp: 1000, scenario: "latest"}
{batchCount: 2, timestampCount: 500, seriesCount: 100, tagCardinality: 10, startTimestamp: 300, endTimestamp: 400, scenario: "historical"}

Query benchmark results before optimization:
BenchmarkFilter/filter-4                   1    4685496299 ns/op    2221001152 B/op  9874080 allocs/op
BenchmarkFilter/filter#01-4         1000000000           0.4689 ns/op          0 B/op          0 allocs/op
BenchmarkFilter/filter#02-4                1    1107566728 ns/op    221140904 B/op    990019 allocs/op

Query benchmark results after optimization:
BenchmarkFilter/filter-large-scale-4            1000000000           0.06766 ns/op         0 B/op          0 allocs/op
BenchmarkFilter/filter-latest-4                 1000000000           0.01630 ns/op         0 B/op          0 allocs/op
BenchmarkFilter/filter-historical-4             1000000000           0.01633 ns/op         0 B/op          0 allocs/op

Memory allocation details of Filter before optimization:
image
Memory allocation details of Filter after optimization:
image
Memory allocation details of Pull after optimization:
image

The improvement is significant and evident.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#12160.


wu-sheng:
Are we going to include this into 0.6?

wu-sheng:
@ButterBright Please fix e2e.

wu-sheng:
I will leave this @hanahmily whether we include this into 0.6 or we should cut 0.6 first and do more tests for this commit.

hanahmily:
> I will leave this @hanahmily whether we include this into 0.6 or we should cut 0.6 first and do more tests for this commit.

I prefer to hold this commitment for a while. Since this improvement is a performance issue, we should introduce benchmark testing in addition to traditional UTs and E2E testing.

banyand/stream/block.go
```diff
@@ -575,6 +585,64 @@ func (bc *blockCursor) loadData(tmpBlock *block) bool {
 	return true
 }
 
+func (bc *blockCursor) searchData(tmpBlock *block) bool {
```
hanahmily, line 588: Please merge this function into `loadData`.

banyand/stream/filter.go
```diff
@@ -0,0 +1,352 @@
...
+)
+
+type filterOptions struct {
+	elementRefMap map[common.SeriesID][]int64
+	pbv1.StreamFilterOptions
+	minTimestamp int64
+	maxTimestamp int64
+}
+
+type filterResult struct {
```
hanahmily, line 42: It may be a good idea to merge this type with `queryResult` since they share the most logic.

banyand/stream/filter.go
```diff
@@ -0,0 +1,352 @@
...
+	db := s.databaseSupplier.SupplyTSDB()
+	var result filterResult
+	if db == nil {
+		return srp, nil
+	}
+	tsdb := db.(storage.TSDB[*tsTable, option])
+	tabWrappers := tsdb.SelectTSTables(*sfo.TimeRange)
+	sort.Slice(tabWrappers, func(i, j int) bool {
+		return tabWrappers[i].GetTimeRange().Start.Before(tabWrappers[j].GetTimeRange().Start)
+	})
```
hanahmily, line 232: Why did you sort the tables here? The table list is already in ascending order by default.

banyand/stream/filter.go
```diff
@@ -0,0 +1,352 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if len(seriesList) == 0 {
+		return srp, nil
+	}
+
+	var sids []common.SeriesID
+	for i := range seriesList {
+		sids = append(sids, seriesList[i].ID)
```
hanahmily, line 256: sids should contain `index.Search` result instead of `tsdb.Lookup`

hanahmily (changes requested):

wu-sheng:
@ButterBright @hanahmily Let's think how we could build tests to verify this? @hanahmily Is your local benchmark suitable for this case?

hanahmily:
> @ButterBright @hanahmily Let's think how we could build tests to verify this? @hanahmily Is your local benchmark suitable for this case?

Existing stress tests could also verify it, but we need to use a lightweight benchmark suite to do this job. I have discussed with @ButterBright to add some benchmarks to the package.

banyand/measure/query.go
```diff
@@ -378,20 +379,32 @@ func (qr *queryResult) Pull() *pbv1.MeasureResult {
 		if len(qr.data) == 0 {
 			return nil
 		}
-		// TODO:// Parallel load
-		tmpBlock := generateBlock()
-		defer releaseBlock(tmpBlock)
+		blankCursorList := []int{}
+		var mu sync.Mutex
+		var wg sync.WaitGroup
```
hanahmily, line 384: You could use a single channel to implement it:

```go
resultsChan := make(chan int, len(qr.data))
...
if qr.loadData(i, tmpBlock) {
		if qr.orderByTimestampDesc(i) {
			qr.data[i].idx = len(qr.data[i].timestamps) - 1
		}
		resultsChan <- -1 // Indicate success
	} else {
		resultsChan <- i // Indicate failure with the index
	}
...
var blankCursorList []int
completed := 0

// Process results from all goroutines.
for completed < len(qr.data) {
result := <-resultsChan
if result != -1 {
	blankCursorList = append(blankCursorList, result)
}
completed++
}

close(resultsChan) // Close the channel as we're done with it.
```

banyand/measure/query.go
```diff
@@ -378,20 +379,32 @@ func (qr *queryResult) Pull() *pbv1.MeasureResult {
...
-				i--
-			}
-			if i < 0 {
-				continue
-			}
-			if qr.orderByTimestampDesc() {
-				qr.data[i].idx = len(qr.data[i].timestamps) - 1
-			}
+			wg.Add(1)
+			go func(i int) {
```
hanahmily, line 387: Please limit the maximum number of goroutines to twice the number of CPU cores specified by the GOMAXPROCS. You can find more information about GOMAXPROCS [here](https://pkg.go.dev/runtime#GOMAXPROCS).

banyand/stream/block.go
```diff
@@ -543,12 +548,30 @@ func (bc *blockCursor) loadData(tmpBlock *block) bool {
 	bc.bm.tagFamilies = tf
 	tmpBlock.mustReadFrom(&bc.tagValuesDecoder, bc.p, bc.bm)
 
-	start, end, ok := timestamp.FindRange(tmpBlock.timestamps, bc.minTimestamp, bc.maxTimestamp)
-	if !ok {
-		return false
+	idxList := make([]int, 0)
+	var start, end int
+	if applyFilter {
```
hanahmily, line 553: The "applyFilter" is unnecessary. Instead, you can simply check if "expectedTimestamps" is equal to nil.

banyand/stream/query.go
```diff
@@ -151,62 +153,77 @@ type queryResult struct {
...
 	if !qr.loaded {
 		if len(qr.data) == 0 {
 			return nil
 		}
-		// TODO:// Parallel load
-		tmpBlock := generateBlock()
-		defer releaseBlock(tmpBlock)
+		blankCursorList := []int{}
+		var mu sync.Mutex
+		var wg sync.WaitGroup
```
hanahmily, line 163: The same as the measure's query.

hanahmily (changes requested):

hanahmily:
@ButterBright The benchmark result shows significant improvement. Can you analyze the benchmark to understand why the allocation is 0? This could provide valuable insights.

ButterBright:
Sure.

hanahmily (approved):

PR #445, opened 2024-05-14
Fix topn query nil and add more context to err

hanahmily:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#12219.


wu-sheng (approved):

lujiajing1126 (approved):
LGTM

lujiajing1126:
Will we cherry-pick this commit to v0.6.x?

wu-sheng:
> Will we cherry-pick this commit to v0.6.x?

It depends on why this happens. If this is blocking the general use, we should have 0.6.1. Otherwise, it is not in urgent. 
e2e covers the feature, but it doesn't complain anything.

@lujiajing1126 Do you face anything in testing?

hanahmily:
> Will we cherry-pick this commit to v0.6.x?

I prefer waiting for 24 hours for future feedback on the error, and regardless of the result, a new patch is necessary.

Almot77:
Okay, i'l test it now.

We use php application in docker + sw php libay for trace collecting + grafana dashboards (made from sw examples for grafana)

PR #455, opened 2024-05-29
Fix several issues of topN pre-calculation

hanahmily:


wu-sheng (approved):
I renamed the milestone name to match your changelogs. And created new 0.7 milestone.

PR #460, opened 2024-06-01
[GsoC][BanyanDB] Self Observability: Fix Metrics Creation 

Sylvie-Wxr:

Fix Explanation:
Metrics such as Gauge, Counter and Histogram should only be created once. 
Fix the code to use Register to create metrics. Use getter to get already registered metrics.


hanahmily (approved):
LGTM

wu-sheng:
This PR seems breaking the CIs somehow. Please verify and recheck.

hanahmily (approved):

PR #464, opened 2024-06-06
Fix the filtering of stream in descending order by timestamp

ButterBright:
When filtering according to the indexRule while sorting in descending order by timestamp,  the elements remaining after applying the limit are incorrect.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,7 @@ Release Notes.
 - Fix duplicated items in the query aggregation top-n list.
 - Fix non-"value" field in topN pre-calculation result measure is lack of data.
 - Encode escaped characters to int64 bytes to fix the malformed data.
+- Fix the filtering of stream in descending order by timestamp.
 
 ## 0.6.0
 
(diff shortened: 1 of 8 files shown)


wu-sheng:
Could you describe what is the effect of this bug?

ButterBright:
Sure.

hanahmily (approved):

CHANGES.md
```diff
@@ -19,6 +19,7 @@ Release Notes.
 - Fix duplicated items in the query aggregation top-n list.
 - Fix non-"value" field in topN pre-calculation result measure is lack of data.
 - Encode escaped characters to int64 bytes to fix the malformed data.
+- Fix the filtering of stream in descending order by timestamp.
```
hanahmily, line 22: Move it to 0.7.0

hanahmily (changes requested):

ButterBright:
When sorting in descending order by timestamp, the previous method always filtered out the elements with the smallest timestamps, rather than the largest ones.

hanahmily (approved):

PR #465, opened 2024-06-06
Return the tagValue of sorted tag

ButterBright:

### DIFF
--- a/api/proto/banyandb/database/v1/schema.proto
+++ b/api/proto/banyandb/database/v1/schema.proto
@@ -171,7 +171,7 @@ message IndexRule {
   }
   // analyzer analyzes tag value to support the full-text searching for TYPE_INVERTED indices.
   Analyzer analyzer = 5;
-  // sorted indicates whether the index is for sorting.
+  // no_sort indicates whether the index is not for sorting.
   bool no_sort = 6;
 }
 
--- a/api/proto/banyandb/database/v1/schema.proto
+++ b/api/proto/banyandb/database/v1/schema.proto
@@ -156,8 +156,10 @@ message IndexRule {
   }
   // type is the IndexType of this IndexObject.
   Type type = 3 [(validate.rules).enum.defined_only = true];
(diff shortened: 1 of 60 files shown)


api/proto/banyandb/database/v1/schema.proto
```diff
@@ -156,8 +156,10 @@ message IndexRule {
   }
   // type is the IndexType of this IndexObject.
   Type type = 3 [(validate.rules).enum.defined_only = true];
+  // sorted indicates whether the index is for sorting.
+  bool sorted = 4;
```
wu-sheng, line 160: Adding the flag(4) would make incompatible change.

The good practice should be the largest index for a new field.

wu-sheng: You should move `sorted` to 7, and accept the value absent with a default value, which could keep the compatibility.

wu-sheng:
Which scenario and how would the client uses this new return time?

ButterBright:
When performing stream sorting, the tagValue will be used for merge sort, ensuring that not too many elements are read out.

api/proto/banyandb/database/v1/schema.proto
hanahmily: Please set the default value to 'true' for compatibility.

hanahmily (changes requested):

wu-sheng:
BTW @hanahmily @ButterBright As we are adding more features into server side, and we plan to keep the compatibility as much as possible with OAP, we should run a matrix e2e testing.

OAP should be set for a version matrix. Such as in this case, we should support the version is 10.0.1 and latest(not latest commit, just another commit ID with `latest`name). Then we will know the version compatible list for the latest BanyanDB server.

hanahmily (changes requested):
You must also check whether the index rule supports sorting in the stream query analyzer.

hanahmily (approved):

PR #466, opened 2024-06-11
Check Unregistered Nodes in Background

hanahmily:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,12 @@
 
 Release Notes.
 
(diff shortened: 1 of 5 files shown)


wu-sheng (approved):

wu-sheng:
@ButterBright When you do merging, please keep the meaningful change logs rather than every commits.

image

That is why we setup squash merge. The commit logs in the main stream should be clear.

ButterBright:
Okay, I understand.

PR #467, opened 2024-06-12
[GsoC][BanyanDB] Self-Observability: Write Metric Data to Measure In Standalone Mode

Sylvie-Wxr:
### <Feature description>
- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#10634.

image

Implementation Detail 
- build IWR from value and labelValues
- Set up map to store metrics, batch flush metrics to measure every 5s.


pkg/meter/native/instruments.go
```diff
@@ -18,10 +18,126 @@
...
+}
+
+// Counter Only Methods.
+func (n *nativeInstrument) Inc(_ float64, _ ...string) {}
+
+// Gauge Only Methods.
+func (n *nativeInstrument) Set(value float64, labelValues ...string) {
+	n.mutex.Lock()
+	defer n.mutex.Unlock()
+	message := bus.NewBatchMessageWithNode(bus.MessageID(time.Now().UnixNano()), "", n.buildIWR(value, labelValues...))
```
hanahmily, line 79: It would be better to store the value in an internal field of 'nativeInstrument' instead. The flushing process will then generate such messages.

pkg/meter/native/instruments.go
```diff
@@ -18,10 +18,126 @@
...
+		log.Error().Err(err).Msg("Failed to convert tagValues to Entity")
+	}
+	writeRequest := &measurev1.WriteRequest{
+		MessageId: uint64(time.Now().UnixNano()),
+		Metadata: &commonv1.Metadata{
+			Group: NativeObservabilityGroupName,
+			Name:  n.measureName,
+		},
+		DataPoint: &measurev1.DataPointValue{
+			Timestamp: timestamppb.New(time.Now().Truncate(time.Millisecond)),
```
hanahmily, line 104: The "second" unit is sufficient; we don't require millisecond precision.

hanahmily (changes requested):

pkg/meter/native/vec.go
```diff
@@ -0,0 +1,146 @@
...
+func newMetricVec(measureName string, pipeline queue.Client, scope meter.Scope) *metricVec {
+	clock, _ := timestamp.GetClock(context.TODO())
+	n := &metricVec{
+		measureName: measureName,
+		pipeline:    pipeline,
+		scope:       scope,
+		scheduler:   timestamp.NewScheduler(log, clock),
+		metrics:     map[uint64]metricWithLabelValues{},
+	}
+	err := n.scheduler.Register("flush metrics", cron.Descriptor, "@every 5s", func(_ time.Time, _ *logger.Logger) bool {
```
hanahmily, line 70: Please move this to "service.go". I prefer to keep a single scheduler for flushing all metrics.

pkg/meter/native/vec.go
```diff
@@ -0,0 +1,146 @@
...
+		messages = append(messages, bus.NewBatchMessageWithNode(bus.MessageID(time.Now().UnixNano()), "", iwr))
+	}
+	_, err := publisher.Publish(data.TopicMeasureWrite, messages...)
+	if err != nil {
+		log.Error().Err(err).Msg("Failed to publish messasges")
+	}
+}
+
+func generateKey(labelValues []string) uint64 {
+	return convert.HashStr(strings.Join(labelValues, ""))
```
hanahmily, line 145: There is a flaw when the label values are:

* "ab", "cd"
* "abc", "d"

Your solution will mess them up.

pkg/meter/native/instruments.go
```diff
@@ -18,10 +18,76 @@
...
+func (c *counter) Inc(delta float64, labelValues ...string) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	key := generateKey(labelValues)
+	v, exist := c.metrics[key]
+	if !exist {
+		v = metricWithLabelValues{
+			labelValues: labelValues,
+		}
+	}
```
hanahmily, line 45: Could you extract these lines to a method of metricVec, since all relevant methods will use it?

pkg/meter/native/vec.go
```diff
@@ -0,0 +1,146 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+	return &measurev1.InternalWriteRequest{
+		Request:      writeRequest,
+		ShardId:      uint32(0),
+		SeriesHash:   pbv1.HashEntity(entities),
```
hanahmily, line 120: You can use this value as the map key to avoid the collision mentioned in the "generateKey" function.

hanahmily (changes requested):

hanahmily (approved):
LGTM

PR #468, opened 2024-06-12
Add version to timestamp file

hanahmily:
### Fix querying old data points 

A version column is introduced to each data point and stored in the timestamp file.

Even the file format is changed, the banyandb server has the capability to operate the previous files.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#12302.


wu-sheng (approved):

hanahmily:
Noop

PR #473, opened 2024-06-21
Fix several bugs

hanahmily:


banyand/queue/pub/client.go
```diff
@@ -115,8 +115,8 @@ func (p *pub) OnAddOrUpdate(md schema.Metadata) {
 			}
 			p.log.Info().Str("status", p.dump()).Str("node", name).Msg("node is removed from active queue by the new gRPC address updated event")
 		}
-		p.registered[name] = node
 	}
+	p.registered[name] = node
```
hanahmily, line 119: @wu-sheng That's the reason why liaison lost the data nodes

wu-sheng (approved):

wu-sheng:
Let's update the demo and see how this is going.

PR #477, opened 2024-06-26
[UI] Fix offset update issues

Sylvie-Wxr:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#12376.

image

image

**Fix detail:**
1. Offset default value set to 0
2. Allow offset and limit to be updated with values user changes to
3. When there is no offset or limit value, the default is 0

### DIFF
--- a/ui/src/components/Read/index.vue
+++ b/ui/src/components/Read/index.vue
@@ -194,7 +194,7 @@ function initCode() {
         }
         timeRange = jsonToYaml(timeRange).data
         data.code = ref(
-            `${timeRange}offset: 1
+            `${timeRange}offset: 0
 limit: 10
 orderBy:
   indexRuleName: ""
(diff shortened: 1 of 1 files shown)


ui/src/components/Read/index.vue
```diff
@@ -315,8 +315,8 @@ function changeTagFamilies() {
 }
 function handleCodeData() {
     const json = yamlToJson(data.code).data
-    param.offset = json.offset ? json.offset : param.offset
-    param.limit = json.limit ? json.limit : param.limit
+    param.offset = (json.offset !== undefined) ? json.offset : 0;
+    param.limit = (json.limit !== undefined) ? json.limit : 0;
```
hanahmily, line 319: limit defaults to 10

hanahmily (changes requested):

hanahmily (approved):

PR #478, opened 2024-06-28
[OSPP] Submitted the delta and cumulative API design plan,

sumingfirst:
which has passed the  make -C api lint check

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@ Release Notes.
 - Check unregistered nodes in background.
 - Improve sorting performance of stream.
 - Add the measure query trace.
+- Assign a separate lookup table to each group in the maglev selector.
 
 ### Bugs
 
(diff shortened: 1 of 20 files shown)


wu-sheng:
Please fix the check.

wu-sheng:
I think this should be for 0.8 rather than the current iteration? @hanahmily

hanahmily:
> I think this should be for 0.8 rather than the current iteration? @hanahmily

This change is part of a long-term plan, so we are not assigning it to any specific milestone.

wu-sheng:
But we will need to merge it at some points. I want to confirm whether it impact the compatibility of client and DB server internal.

hanahmily (approved):

hanahmily:
> But we will need to merge it at some points. I want to confirm whether it impact the compatibility of client and DB server internal.

No it won't break anything. We can merge it safely.

PR #479, opened 2024-06-28
[ospp] Adds MeasureAggregateFunctionService.Support API

StLeoX:
### <Feature description>
- [x] If this is non-trivial feature, paste the links/URLs to the design doc.

[link](https://github.com/apache/skywalking/issues/8720#issuecomment-2196389729)


api/proto/banyandb/database/v1/schema.proto
```diff
@@ -95,6 +96,8 @@ message FieldSpec {
   EncodingMethod encoding_method = 3 [(validate.rules).enum.defined_only = true];
   // compression_method indicates how to compress data during writing
   CompressionMethod compression_method = 4 [(validate.rules).enum.defined_only = true];
+  // aggregate_function indicates how to aggregate data
+  model.v1.AggregationFunction aggregate_function = 5;
```
wu-sheng, line 100: @hanahmily I think this is not the function I mentioned. This AggregationFunction is targeting for query aggregation, this API changes should be stream flowing process aggregation, right?

hanahmily: Our plan is to use the same functions to handle these scenarios. These functions are pure computing logic without any context related to the processes they are in.

api/proto/banyandb/model/v1/common.proto
```diff
@@ -77,3 +77,17 @@ enum AggregationFunction {
...
+  MEASURE_AGGREGATE_UNSPECIFIED = 0;
+  MEASURE_AGGREGATE_MIN = 1;
+  MEASURE_AGGREGATE_MAX = 2;
+  MEASURE_AGGREGATE_COUNT = 3;
+  MEASURE_AGGREGATE_SUM = 4;
+  MEASURE_AGGREGATE_AVG = 5;
+  MEASURE_AGGREGATE_PERCENT = 6;
+  MEASURE_AGGREGATE_RATE = 7;
+  MEASURE_AGGREGATE_HISTOGRAM = 8;
+  MEASURE_AGGREGATE_PERCENTILE2 = 9;
```
hanahmily, line 91: @wu-sheng Would you mind if we call it 'percentile2', which aligns with the name defined in OAP?

wu-sheng: I think it is OK to align. Do you have percentile? I think we need that too.

StLeoX: Ok, I'll add that.

docs/api-reference.md
```diff
@@ -690,6 +695,28 @@ Trace is the top level message of a trace.
...
 
+
+<a name="banyandb-model-v1-MeasureAggregate"></a>
+
+### MeasureAggregate
+
+
+| Name | Number | Description |
+| ---- | ------ | ----------- |
+| MEASURE_AGGREGATE_UNSPECIFIED | 0 |  |
```
wu-sheng, line 706: I think comments are required to generate docs here, to describe meanings of each function.

StLeoX: Sure

hanahmily (approved):

PR #485, opened 2024-07-13
Support duplicated data for Stream

hanahmily:
### Fix a bug that the Stream module didn't support duplicated in index-based filtering and sorting

In the index, the document id holds the timestamp before these changes. This can lead to several incorrect reads when using the index to filter or sort the data, especially when there are multiple duplicated timestamps in a single time series.

In this pull request, I replaced the timestamp with the elementID, which is unique in a time series. The sorted elementID is now represented by the hashed uint64 value instead of the original string. The file system version is also updated due to this change.

### Fix missing data when a tag filter is applied

When sorting data by an index, if the final result size exceeds the offset+limit and a tag filter is applied, some elements may be lost in the query result.

In this pull request, the `limit` and `tag filter` executor can drain the parent source until no more data is returned or until the offset+limit is reached. This should help resolve the issue of missing data.

A new `distributed limit`, which is the same as the previous `limit`, is applied to the distributed executor. This is because the liaison couldn't drain data nodes several times; it should interact with each data node only once in a query request.

### Reverted Index Enhancement

The "Write" operation has been removed from the reverted index. Now, we can use the `Batch` operation to write index data.

The timestamp has been moved from the document id to a field with a date-time type, and its value is stored in the index.

@ButterBright Please check the query process to verify the change is expected.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#12371,apache/skywalking#12375


wu-sheng (dismissed):

pkg/index/inverted/inverted.go
```diff
@@ -174,7 +183,7 @@ func (s *store) Iterator(fieldKey index.FieldKey, termRange index.RangeOpts, ord
 	if err != nil {
 		return nil, err
 	}
-	fk := fieldKey.MarshalIndexRule()
+	fk := fieldKey.Marshal()
```
ButterBright, line 186: Please remove the line bluge.blugeMatchAllquery() from this function, as it's redundant.

hanahmily: done

ButterBright (approved):

PR #490, opened 2024-07-18
[GSOC] [Self-Observability] Suppress Error When Getting Path Stat

Sylvie-Wxr:
### Fixes apache/skywalking#12462
- [ ] Explain briefly why the bug exists and how to fix it.
- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#12462.


banyand/observability/metrics_system.go
```diff
@@ -171,8 +171,7 @@ func collectDisk() {
 	for path := range getPath() {
 		usage, err := disk.Usage(path)
 		if err != nil {
-			// skip logging the case where the data has not been created
-			if err.Error() == "no such file or directory" {
+			if err.Error() != "no such file or directory" {
```
hanahmily, line 174: No, checking the error message is not reliable. 
```suggestion
if _, err := os.Stat(path); err != nil {
			if !os.IsNotExist(err) {
				log.Error().Err(err).Msgf("failed to get stat for path: %s", path)
			}
			return
		}
```

hanahmily (changes requested):

hanahmily (approved):

PR #491, opened 2024-07-22
Optimize query performance of series index

ButterBright:
- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#12239.


wu-sheng:
Is there benchmark or some reports to indicate how much this will improve?

ButterBright:
Not yet. Let me add one.

banyand/internal/storage/index.go
```diff
@@ -201,12 +200,11 @@ func (s *seriesIndex) Search(ctx context.Context, series []*pbv1.Series, filter
 	}
 
 	pl := seriesList.ToList()
-	if filter != nil && filter != logical.ENode {
+	if blugeQuery != nil {
 		var plFilter posting.List
 		func() {
 			if tracer != nil {
 				span, _ := tracer.StartSpan(ctx, "filter")
```
hanahmily, line 207: Please add `String` to blugeQuery. The query tracing needs to debug the relevant issues.

hanahmily (changes requested):
As we discussed before, the bluge's structs lack the `String` needed to build the tracing tree. Furthermore, the bulge, as a third-party library, should not spread out the inverted package. Considering these two reasons, you would have to design a series of structs to encapsulate bluge queries.

pkg/index/inverted/inverted.go
```diff
@@ -304,6 +305,26 @@ func (s *store) Range(fieldKey index.FieldKey, opts index.RangeOpts) (list posti
 	return
 }
 
+func (s *store) Execute(query bluge.Query) (posting.List, error) {
+	reader, err := s.writer.Reader()
```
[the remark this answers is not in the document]
ButterBright: I assume that the reader will be closed when `iter.Close` is called.

banyand/internal/storage/storage.go
```diff
@@ -67,7 +68,7 @@ type SupplyTSDB[T TSTable] func() T
 // IndexDB is the interface of index database.
 type IndexDB interface {
 	Write(docs index.Documents) error
-	Search(ctx context.Context, series []*pbv1.Series, filter index.Filter, order *model.OrderBy, preloadSize int) (pbv1.SeriesList, error)
+	Search(ctx context.Context, series []*pbv1.Series, query *inverted.BlugeQuery, filter index.Filter, order *model.OrderBy, preloadSize int) (pbv1.SeriesList, error)
```
hanahmily, line 71: Is the `filter` still necessary?

pkg/index/index.go
```diff
@@ -188,6 +190,8 @@ type SeriesStore interface {
 	Store
 	// Search returns a list of series that match the given matchers.
 	Search(context.Context, []SeriesMatcher) ([]Series, error)
+	// Execute returns a posting list that matches the given query.
+	Execute(bluge.Query) (posting.List, error)
```
hanahmily, line 194: Please encapture the `bluge` in the inverted pkg.

pkg/index/inverted/index_filter.go
```diff
@@ -15,38 +15,129 @@
...
 type GlobalIndexError struct {
 	IndexRule *databasev1.IndexRule
-	Expr      LiteralExpr
+	Expr      logical.LiteralExpr
 }
 
 func (g GlobalIndexError) Error() string { return g.IndexRule.String() }
 
+// BlugeQuery is a wrapper for bluge.Query.
+type BlugeQuery struct {
```
hanahmily, line 56:
```suggestion
type Query struct {
```

pkg/index/inverted/index_filter.go
```diff
@@ -15,38 +15,129 @@
...
+			return &BlugeQuery{query}, entities, false, nil
+		}
+	}
+	return nil, nil, false, logical.ErrInvalidCriteriaType
+}
+
 // BuildLocalFilter returns a new index.Filter for local indices.
-func BuildLocalFilter(criteria *modelv1.Criteria, schema Schema, entityDict map[string]int,
-	entity []*modelv1.TagValue, mandatoryIndexRule bool,
+func BuildLocalFilter(criteria *modelv1.Criteria, schema logical.Schema,
```
hanahmily, line 139: Why do you move the filter here? Its implementation doesn't relate to bulge.

It should be in the `pkg/query/logical/stream`

pkg/index/inverted/index_filter.go
hanahmily, line 1: Rename the file name to `query.go`

pkg/index/inverted/index_filter.go
hanahmily, line 56: Add `ToString` to this struct.

hanahmily (changes requested):

pkg/index/inverted/inverted.go
```diff
@@ -62,26 +62,28 @@ var (
...
+// StoreOpts wraps options to create an inverted index repository.
 type StoreOpts struct {
 	Logger       *logger.Logger
 	Path         string
 	BatchWaitSec int64
 }
 
-type store struct {
+// Store is an inverted index repository.
+type Store struct {
```
hanahmily, line 86: You should not expose this type since we defined several interfaces to enclose it. If you want to use its methods, define this function in an interface firstly.

hanahmily (changes requested):

hanahmily (approved):

PR #505, opened 2024-08-04
Merge the query process of series index

ButterBright:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#12239.


wu-sheng:
Pleaae update e2e tool version to fix CI

banyand/internal/storage/index.go
```diff
@@ -245,31 +223,49 @@ func (s *seriesIndex) Search(ctx context.Context, series []*pbv1.Series, opts In
...
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+	var query index.Query
+	if opts.Query != nil {
+		query, err = s.store.Query(seriesMatchers, opts.Query)
+	} else {
+		query, err = s.store.Query(seriesMatchers, nil)
+	}
```
hanahmily, line 238:
```suggestion
	query, err = s.store.Query(seriesMatchers, opts.Query)
```

pkg/index/index.go
```diff
@@ -156,7 +157,8 @@ type Writer interface {
 
 // FieldIterable allows building a FieldIterator.
 type FieldIterable interface {
-	Iterator(fieldKey FieldKey, termRange RangeOpts, order modelv1.Sort, preLoadSize int) (iter FieldIterator[*DocumentResult], err error)
+	Query(seriesMatchers []SeriesMatcher, secondaryQuery Query) (Query, error)
```
hanahmily, line 160: May we call it `BuildQuery`?

pkg/index/index.go
```diff
@@ -204,8 +206,7 @@ type SeriesDocument struct {
 type SeriesStore interface {
 	Store
 	// Search returns a list of series that match the given matchers.
-	Search(context.Context, []SeriesMatcher, []FieldKey) ([]SeriesDocument, error)
-	Execute(context.Context, Query) (posting.List, error)
+	Search(context.Context, []SeriesMatcher, []FieldKey, Query) ([]SeriesDocument, error)
```
hanahmily, line 209: I might merge `SeriesMatcher` and `Query` into `Query`. The storage.index could build the query first, then pass it to the `Search`.

pkg/index/index.go
```diff
@@ -156,7 +157,8 @@ type Writer interface {
 
 // FieldIterable allows building a FieldIterator.
 type FieldIterable interface {
-	Iterator(fieldKey FieldKey, termRange RangeOpts, order modelv1.Sort, preLoadSize int) (iter FieldIterator[*DocumentResult], err error)
+	Query(seriesMatchers []SeriesMatcher, secondaryQuery Query) (Query, error)
+	Iterator(fieldKey FieldKey, termRange RangeOpts, order modelv1.Sort, preLoadSize int, query Query, fields []string) (iter FieldIterator[*DocumentResult], err error)
```
hanahmily, line 161:
```suggestion
	Iterator(fieldKey FieldKey, termRange RangeOpts, order modelv1.Sort, preLoadSize int, query Query, fields []FieldKey) (iter FieldIterator[*DocumentResult], err error)
```

Please ensure that the semantics are consistent with `SeriesStore.Search`.

hanahmily (changes requested):

pkg/index/inverted/inverted.go
```diff
@@ -210,21 +210,31 @@ func (s *store) Iterator(fieldKey index.FieldKey, termRange index.RangeOpts,
...
 	}
+	query := bluge.NewBooleanQuery().AddMust(rangeQuery)
+	if indexQuery != nil && indexQuery.(*Query).query != nil {
+		query.AddMust(indexQuery.(*Query).query)
+	}
+	fields := make([]string, 0, len(fieldKeys))
+	for i := range fieldKeys {
+		fields = append(fields, fieldKeys[i].Marshal())
+	}
+	// TODO: add trace information
```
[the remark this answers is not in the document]
hanahmily: I agree. We should trace the query before merging it.

pkg/index/inverted/query.go
```diff
@@ -49,30 +49,22 @@ type GlobalIndexError struct {
 
 func (g GlobalIndexError) Error() string { return g.IndexRule.String() }
 
+var _ index.Query = (*Query)(nil)
+
 // Query is a wrapper for bluge.Query.
 type Query struct {
```
hanahmily, line 55: Do you need to expose this type since its operations are wrapped by the `index.Query`?

hanahmily (changes requested):

hanahmily (approved):
Nice job

wu-sheng:
What is the new query performance benchmark after this?

ButterBright:
Sorry for the late reply. I'm adding the relevant benchmarks. Theoretically, the topology queries should see a significant improvement.

PR #514, opened 2024-08-18
Support ARM64 build binary and images

StLeoX:
### Support ARM64 build binary and images
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#12308 .


.github/workflows/publish-docker.yml
```diff
@@ -79,6 +79,10 @@ jobs:
         run: |
           GOOS=windows GOARCH=amd64 make -C banyand banyand-server-static
           GOOS=windows GOARCH=amd64 make -C bydbctl build
+      - name: Build OSX binaries
+        run: |
+          GOOS=darwin GOARCH=arm64 make -C banyand banyand-server-static
+          GOOS=darwin GOARCH=arm64 make -C bydbctl build
```
wu-sheng, line 85: Is this AARM or ARM? The CI name seems a little tricky.

StLeoX:
Currently, I just add `darwin-arm64` for  docker  publishing.
  
I wonder if we really need something like `banyand-server-static-darwin-arm64` in the tar release. And the binaries list  may  be semilar  with `bydnctl` ?

wu-sheng:
> Currently, I just add `darwin-arm64` for  docker  publishing.
> 
>   
> 
> I wonder if we really need something like `banyand-server-static-darwin-arm64` in the tar release. And the binaries list  may  be semilar  with `bydnctl` ?

darwin(aarm), we don't need this for official release. But we need Linux ARM as that, same for ctl.

The issue is referring these two things.

lujiajing1126:
Shall we consider [goreleaser](https://goreleaser.com/) to build binanries and docker images?

wu-sheng:
We could consider this in a separate thread?

wu-sheng:
@hanahmily Please confirm whether this is the scope of this PR. Because in the end, we will need 3 images(linux intel, linux and, dawin) in GitHub image repo and first two in the DockerHub.

hanahmily:
The build system needs to be improved to support building multi-platform binaries, which is more complicated than I initially thought. I will address this issue. @StLeoX, I will leave you to your ospp project.

StLeoX:
Fine

PR #515, opened 2024-08-19
[GSoC][UI] Fix and show message when group is not exist and native is turned off

Sylvie-Wxr:
### Fix the error message 500 when native mode is not turned on and the group is not created

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#10634.

**Problem:**
Previously when the native mode is not turned on, the dashboard showed error since the group is not created:
image
<br>
**Solution:**
Now, this bug is fixed and add a message to user to remind them of the info and to turn on the native mode:
Screenshot 2024-08-18 at 10 06 25 PM


ui/src/views/Dashboard/index.vue
```diff
@@ -360,7 +387,11 @@ watchEffect(() => {
                 </el-select>
             </span>
         </div>
-
+        <div class="error-alert">
+            <!-- Conditionally display the alert if hasMonitoring is false -->
+            <el-alert v-if="!hasMonitoring" title="Self-monitoring not available, please turn on native flag."
```
hanahmily, line 392:
```suggestion
            <el-alert v-if="!hasMonitoring" title="Self-monitoring not available, please turn it on by setting \"--observability-modes=native\"."
```

hanahmily (changes requested):

hanahmily (approved):

PR #520, opened 2024-08-25
[Feature] Support duplicate data verification

zzzk1:
### <Feature description>

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#12253.


wu-sheng:
I think you followed the wrong context.
What we discussed is about adding data verification in the e2e tool for OAP. Not for database itself.
We don't want this performance cost in the database.

PR #526, opened 2024-09-03
Add slow query logging

hanahmily:
This PR introduces slow query logging to BanyanDB. Slow query logging records the slow query before its execution latency times out. This will remind users to analyze the query by using query tracing. When the query takes a long time to execute, the query tracing result may not provide enough information to analyze the root cause. With this feature, users will have a better chance of identifying and troubleshooting slow queries.


wu-sheng (approved):
Good. I like this flag.

PR #528, opened 2024-09-05
Dev measure aggregate function

StLeoX:
# Add several functions

Here are new added functions:
- max
- count
- sum
- percent
- rate


wu-sheng:
Please rebases the commits to avoid so long commit logs.

banyand/measure/aggregate/aggregate_function.go
```diff
@@ -76,6 +86,20 @@ func zeroValue[R Output]() R {
 	return r
 }
 
+func minValue[R Output]() (r R) {
+	switch a := any(&r).(type) {
+	case *int64:
+		*a = math.MinInt64
+	case *float64:
+		*a = -math.MaxFloat64
+	case *string:
```
hanahmily, line 95: It should panic

hanahmily (changes requested):

banyand/measure/aggregate/avg.go
```diff
@@ -24,14 +24,14 @@ type Avg[A, B Input, R Output] struct {
 }
 
 // Combine takes elements to do the aggregation.
-// Avg uses type parameter A and B.
-func (a *Avg[A, B, R]) Combine(arguments Arguments[A, B]) error {
+// Avg uses type parameter A.
+func (f *Avg[A, B, R]) Combine(arguments Arguments[A, B]) error {
 	for _, arg0 := range arguments.arg0 {
 		switch arg0 := any(arg0).(type) {
```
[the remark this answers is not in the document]
StLeoX: `sum[A Input, R Output]` will dismatch with `Function[A, B Input, R Output]`, then can't call `Combine` on it.
I've removed type assertion, by aliasing int64 to  Void.

banyand/measure/aggregate/aggregate_function.go
```diff
@@ -26,20 +26,20 @@ import (
...
-type Void struct{}
+// It's implemented as int64, but it won't be used as an int64.
+type Void int64
 
-// Input covers possible types of Function's arguments. It synchronizes with `FieldType` in schema.
+// Input covers possible types of Function's arguments. It synchronizes with `FieldType`.
+// It also covers Void type.
 type Input interface {
-	Void | ~int64 | ~float64
+	~int64 | ~float64
```
hanahmily, line 35: Do you need to separate 'Input' and 'Output'?

banyand/measure/aggregate/avg.go
```diff
@@ -27,23 +27,11 @@ type Avg[A, B Input, R Output] struct {
...
 	for _, arg0 := range arguments.arg0 {
-		switch arg0 := any(arg0).(type) {
-		case int64:
-			f.summation += R(arg0)
-		case float64:
-			f.summation += R(arg0)
-		default:
-			return errFieldValueType
-		}
+		f.summation += R(arg0)
```
hanahmily, line 30: We could enhance the performance by following the guidelines in https://sourcegraph.com/blog/slow-to-simd. You can implement the suggested improvement methods outlined in the article. However, I wouldn't advise using the SIMD path as it's overly complex. Kindly provide the performance comparison results by submitting benchmarks for each function.

hanahmily (changes requested):

banyand/measure/aggregate/aggregate_function.go
StLeoX: Laterly we may introduce `HashMap` to Output.

banyand/measure/aggregate/aggregate_function.go
```diff
@@ -54,6 +54,12 @@ type Function[A, B Input, R Output] interface {
 
 	// Result gives the result for the aggregation.
 	Result() R
+
+	// FirstCumulation gives the first cumulative state for the aggregation.
+	FirstCumulation() A
+
+	// SecondCumulation gives the second cumulative state for the aggregation.
+	SecondCumulation() B
```
wu-sheng, line 62: I think `Cumulation` is strange. Where do you find the naming?

StLeoX: Maybe like `cumulative` ?
I used this to represent the intermediate state hold by the struct, like Min.minimum.

wu-sheng: I am not sure about the point. But Culculation is not a popular word(I am even not sure, is it a real word).
Meanwhile, in a statistic, these things are called factor or parameter of this aggregates.

So, my questions are about the source of this naming?

@hanahmily Could you recheck this?

hanahmily: Why not return them through "Result() (A, B, R)"?

hanahmily: So "Result() (A, B, R)" is enough. We don't have any chance to get them independently.

hanahmily: > I am quite sure, this is a wrong naming.

Yes this naming needs to be corrected. The NewRelic's link only tries to guess the source of "cumulative".

StLeoX: Fine, I'll change the interface to `Result() (A, B, R)`, and then once calling Result  returns all we need.

StLeoX:
# Benchmark loop-unrolling

The benchmark code is [here](https://github.com/StLeoX/skywalking-banyandb/commit/86c6db0a4501688ce09887f38c588ad7317efe86#diff-dc8eb834dc7b7327dade49c5c3d6940f0dcd3ee7995dcffbb3085103cc383f00R52).

On my machine, the result is:
image

We can see that, with the round increasing, the efficiency difference is decreasing.

hanahmily (approved):

PR #530, opened 2024-09-08
Add benchmarks for topological queries

ButterBright:
The specification of the test environment is:
OS: Ubuntu 22.04.3 LTS x86_64
CPU: AMD EPYC 7282 (4) @ 2.794GHz
Memory: 7951MiB

Topological query benchmark results before series index optimization:
Metric Name, Min, Max, Mean, Median, P95
topology-0, 0.038543, 0.362319, 0.099407, 0.084507, 0.234849

Topological query benchmark results after series index optimization:
Metric Name, Min, Max, Mean, Median, P95
topology-0, 0.038886, 0.236007, 0.082450, 0.075451, 0.133140

Index query trace before optimization:

Index query trace after optimization:

As we can see, the topological query performance has improved, especially for p95. The index optimization proves effective, as expected.


hanahmily:
@ButterBright Could you share the query tracing result between them? I want to ensure the scanning strategy is improved as we expected.

ButterBright:
Updated.

wu-sheng (approved):

PR #550, opened 2024-10-08
Fix metrics system typo

sollhui:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#<issue number>.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,12 @@
 
 Release Notes.
 
+## 0.8.0
+
+### Chores
+
+- Fix metrics system typo.
+
 ## 0.7.0
 
 ### File System Changes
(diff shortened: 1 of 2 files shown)


wu-sheng:
I think you should update the dashboard setup(Grafana) accordingly?

wu-sheng:
Also, changelog should be updated for 0.8.

FYI 0.7 is on the vote, this typo fix could be in 0.8-dev directly.

sollhui:
> dashboard

This typo only involves the code level and does not require any modifications to the dashboard.

wu-sheng:
OK, please update the changelog.

wu-sheng (approved):

PR #562, opened 2024-12-04
Index: Use numeric index type to support Int and Float

hanahmily:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#12797.


wu-sheng:
Does this require reset a new database? The previous version data is not supported anymore.

ButterBright (approved):

PR #577, opened 2024-12-19
Parse string and int array in the query result table

Fine0830:
- [x] Closes apache/skywalking#12857.

View string and int array in the query result table. And fix style.


wu-sheng:
Could you post a demo video?

Fine0830:
> Could you post a demo video?

There is no array data in the demo env. Need @hanahmily to add array data.

wu-sheng:
I think segment(stream) should have?

Fine0830:
> I think segment(stream) should have?

It seems that there is no array data either.

wu-sheng:
@hanahmily could you mock some data?

wu-sheng (approved):

PR #578, opened 2024-12-21
Index `measure_name` and `tags` in `entity` to improve the query performance.

hanahmily:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#12877.


wu-sheng:
This should improve the `mode=index` measure query performance, right?

wu-sheng:
And it doesn't impact the other normal measure data.

hanahmily:
> This should improve the `mode=index` measure query performance, right?

Yep. It doesn't impact other measures

wu-sheng (approved):
Does this improve enough to show benefits on demo.s.a.o? If so, we should consider redeploying?

PR #579, opened 2024-12-23
Fix querying the index on a entity tag

hanahmily:
Fix issues found at https://github.com/apache/skywalking/pull/12888


hanahmily:
I will remove ebpf relevant test cases temparely. It will be added in the 0.9.0

api/proto/banyandb/measure/v1/rpc.proto
```diff
@@ -38,5 +38,10 @@ service MeasureService {
   }
 
   rpc Write(stream WriteRequest) returns (stream WriteResponse);
-  rpc TopN(TopNRequest) returns (TopNResponse);
+  rpc TopN(TopNRequest) returns (TopNResponse) {
+    option (google.api.http) = {
```
hanahmily, line 42: It is related to apache/skywalking#12848.

wu-sheng (approved):

PR #582, opened 2024-12-26
feat(UI): Implement `TopNAggregation` data query page

Fine0830:
- [x] Closes apache/skywalking#12848.

Video

https://github.com/user-attachments/assets/3b520be9-2994-4f15-9097-9cb08af67a5a


wu-sheng:
@hanahmily The current `read-*` tab is a kind of query APIs, which is good. But meanwhile, we should provide raw data view. If I am as a developer, I want to check topN list(all existing ones) of each time buckets, e.g. T1->list 0...1000, T2->list 0...1000

If possible, I want to add another raw data view.

hanahmily (approved):

hanahmily:
> @hanahmily The current `read-*` tab is a kind of query APIs, which is good. But meanwhile, we should provide raw data view. If I am as a developer, I want to check topN list(all existing ones) of each time buckets, e.g. T1->list 0...1000, T2->list 0...1000
> 
> 
> 
> If possible, I want to add another raw data view.

We could restrict the time range to a single time bucket to get the top-n list within that bucket.

PR #587, opened 2024-12-30
Change Property's storage layer

hanahmily:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#12664.


api/proto/banyandb/property/v1/rpc.proto
```diff
@@ -96,17 +98,10 @@ service PropertyService {
     option (google.api.http) = {delete: "/v1/property/{metadata.container.group}/{metadata.container.name}/{metadata.id}/{tags}"};
   }
 
-  rpc Get(GetRequest) returns (GetResponse) {
-    option (google.api.http) = {get: "/v1/property/{metadata.container.group}/{metadata.container.name}/{metadata.id}/{tags}"};
-  }
-
-  rpc List(ListRequest) returns (ListResponse) {
+  rpc Query(QueryRequest) returns (QueryResponse) {
```
hanahmily, line 101: @wu-sheng @wankai123 The new `Query` operation filters the properties. `List` and `Get` will be removed.

wu-sheng: It is fine.  We will adopt those changes. We prefer querying.

hanahmily:
@wankai123 Once this is merged, I will publish the client 0.8.0-rc1 to adopt this chage.

wu-sheng (approved):

wu-sheng:
> @wankai123 Once this is merged, I will publish the client 0.8.0-rc1 to adopt this chage.

@hanahmily In oap implementation, you should only set shard for property to 1 shard, and we will do replication when we have that.

PR #602, opened 2025-02-06
Fix NPE on sorting by index with an empty result

hanahmily:
This bug is brought by #597


wankai123 (approved):

wu-sheng:
CI fails. Please fix them.

wu-sheng (approved):

PR #605, opened 2025-02-11
Add Backup CLI tool

hanahmily:
This's a part of apache/skywalking#12876

### DIFF
--- a/docs/operation/backup.md
+++ b/docs/operation/backup.md
@@ -23,7 +23,8 @@ Before running the backup tool, ensure you have:
 - A running BanyanDB data node exposing a gRPC service (by default at `127.0.0.1:17912`).
 - Network connectivity to the data node.
 - A valid destination URL specified using the `file:///` scheme (or any supported scheme in future releases).
-- Necessary access rights for reading snapshot data and writing to the destination.
+- Necessary access rights for writing to the destination.
+- Sufficient permissions to access the snapshots directories for **Stream**, **Measure**, and **Property** catalogs on the data node.
 
 ## Command-Line Usage
 
--- a/docs/operation/backup.md
+++ b/docs/operation/backup.md
@@ -1,0 +1,73 @@
+# Backup Data Using BanyanDB Backup Tool
+
+The backup command-line tool is used to backup BanyanDB snapshots to remote storage. The tool allows both one-time and scheduled backups.
+
+## Overview
+
+The backup tool performs the following operations:
+
+- Connects to a BanyanDB data node via gRPC to request the latest snapshots.
+- Determines the local snapshot directories based on the catalog type:
+  - **Stream Catalog:** Uses the `stream-root-path`.
+  - **Measure Catalog:** Uses the `measure-root-path`.
+  - **Property Catalog:** Uses the `property-root-path`.
+- Computes a time-based directory name (formatted as daily or hourly).
+- Uploads files that are not found in the remote storage to the specified destination.
+- Deletes orphaned files in the remote destination that no longer exist locally.
+- Optionally schedules periodic backups using cron-style expressions.
+
+## Prerequisites
+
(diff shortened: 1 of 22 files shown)


docs/operation/backup.md
```diff
@@ -0,0 +1,73 @@
+# Backup Data Using BanyanDB Backup Tool
+
+The backup command-line tool is used to backup BanyanDB snapshots to remote storage. The tool allows both one-time and scheduled backups.
+
+## Overview
+
+The backup tool performs the following operations:
+
+- Connects to a BanyanDB data node via gRPC to request the latest snapshots.
+- Determines the local snapshot directories based on the catalog type:
```
wu-sheng, line 10: So, this tool will be deployed as a sidecar(in k8s) of a data node and request through grpc locally?

docs/operation/backup.md
```diff
@@ -0,0 +1,73 @@
+# Backup Data Using BanyanDB Backup Tool
```
wu-sheng, line 1: This doc requires a menu item to show up in the website.

docs/operation/backup.md
```diff
@@ -0,0 +1,73 @@
...
+The backup tool performs the following operations:
+
+- Connects to a BanyanDB data node via gRPC to request the latest snapshots.
+- Determines the local snapshot directories based on the catalog type:
+  - **Stream Catalog:** Uses the `stream-root-path`.
+  - **Measure Catalog:** Uses the `measure-root-path`.
+  - **Property Catalog:** Uses the `property-root-path`.
+- Computes a time-based directory name (formatted as daily or hourly).
+- Uploads files that are not found in the remote storage to the specified destination.
+- Deletes orphaned files in the remote destination that no longer exist locally.
```
wu-sheng, line 16: As the directory name includes time, what causes the orphaned files to exist? The backup happened twice in one hour(or other time style)?

docs/operation/backup.md
wu-sheng: If so, I think we should add some docs about this tool should be used in the same place of the data node.

wu-sheng: > So, this tool will be deployed as a sidecar(in k8s) of a data node and request through grpc locally?

Or, it will be included into DB server binary folder and just run locally?

hanahmily: Added in the `Prerequisites` section

docs/operation/backup.md
hanahmily: Done

docs/operation/backup.md
hanahmily: The immutable files are deleted locally. To my knowledge, only the old and sufficiently large files can remain until the next backup time, which is one hour later. Most small files will be cleared up in the remote.

docs/operation/backup.md
hanahmily: > > So, this tool will be deployed as a sidecar(in k8s) of a data node and request through grpc locally?
> 
> Or, it will be included into DB server binary folder and just run locally

My preference is to build a separate binary as `bydbbackup` which contains two commands: `backup` and `restore`. We can assemble the binary into the current image.

docs/operation/backup.md
wu-sheng: I saw your doc about time-style, and backup period, which could be per hour. So, to delete the old files, mostly because the time-style is in day, and we run backup multiple times per day, right?

wu-sheng (approved):

docs/operation/backup.md
hanahmily: Absolutely. By setting the `schedule` to `hourly` and the `time-style` to `daily`, the remote destination will be uploaded 24 times a day. This is my preferred strategy, and I plan to use it as the default setting for the Banyandb Helm chart.

PR #621, opened 2025-03-17
Introduce the schema style to the property

hanahmily:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13072.

I will tweak the documents in the following-up PRs.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,6 +32,7 @@ Release Notes.
 - Improve the write performance.
 - Add node labels to classify the nodes.
 - Add lifecycle management for the node.
+- Property: Introduce the schema style to the property.
 
 ### Bug Fixes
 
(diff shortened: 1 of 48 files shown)


test/docker/base-compose.yml
```diff
@@ -59,7 +59,7 @@ services:
       - sw_agent:/skywalking-java-agent
 
   oap:
-    image: "ghcr.io/apache/skywalking/oap:${SW_OAP_COMMIT}"
+    image: "docker.io/hanahmily/oap:ca528d984506faee1276232f4ebd31eaba9db40c"
```
wu-sheng, line 62: TODO, will update this after the new driver RC released, and oap updated.

wu-sheng (approved):

PR #630, opened 2025-03-23
Add time range parameters to stream index filter

ButterBright:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#12444

### DIFF
--- a/pkg/index/index.go
+++ b/pkg/index/index.go
@@ -167,14 +167,14 @@ func (f FloatTermValue) String() string {
 	return strconv.FormatInt(numeric.Float64ToInt64(f.Value), 10)
 }
 
-// IntTermValue represents a int term value.
-type IntTermValue struct {
+// TimestampValue represents a int term value.
+type TimestampValue struct {
 	Value int64
 }
 
-func (IntTermValue) isTermValue() {}
+func (TimestampValue) isTermValue() {}
 
-func (i IntTermValue) String() string {
+func (i TimestampValue) String() string {
 	return strconv.FormatInt(i.Value, 10)
 }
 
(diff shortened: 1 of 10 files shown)


wu-sheng:
I think the original issue is in 0.9?
We are going to cut release very soon for 0.8

pkg/index/index.go
```diff
@@ -194,12 +205,22 @@ func (r RangeOpts) Valid() bool {
...
+			return false
+		}
 	default:
 		return false
 	}
 	return true
 }
 
+// IsTimeRangeQuery returns true if the range is for a time range query.
+func (r RangeOpts) IsTimeRangeQuery() bool {
```
hanahmily, line 219: I don't like it. Would you please create a "TimeRangeOpts" ?

pkg/index/index.go
```diff
@@ -167,6 +167,17 @@ func (f FloatTermValue) String() string {
 	return strconv.FormatInt(numeric.Float64ToInt64(f.Value), 10)
 }
 
+// IntTermValue represents a int term value.
```
hanahmily, line 170: It's not a general IntTermValue; it is only used by the timestamp. How about renaming it to "TimeStampValue"?

pkg/index/inverted/query.go
```diff
@@ -464,11 +464,12 @@ func (m *matchAllNode) String() string {
...
-	min          string
-	max          string
-	minInclusive bool
-	maxInclusive bool
+	indexRule        *databasev1.IndexRule
+	min              string
+	max              string
+	minInclusive     bool
+	maxInclusive     bool
+	isTimeRangeQuery bool
```
hanahmily, line 472: It always seems "false".

hanahmily (changes requested):

hanahmily (approved):

PR #642, opened 2025-04-06
Enable Dynamic Reloading of TLS Certificates and Keys

onkar717:
### Feature Description

BanyanDB currently supports TLS for encrypted communication between clients and the server. However, it lacks the capability to automatically reload updated TLS certificates and keys without requiring a server restart. This feature implements dynamic reloading of TLS materials, enhancing operational flexibility and security.

### Testing

rm -rf test-certs && mkdir test-certs && cd test-certs

#### Step 1: Generate initial cert/key pair
openssl req -x509 -newkey rsa:2048 -keyout key1.pem -out cert1.pem -days 365 -nodes -subj "/CN=localhost"
cp cert1.pem cert.pem && cp key1.pem key.pem 

#### Step 2: Start the server with TLS
./banyand/build/bin/dev/banyand-server standalone --tls=true --cert-file=test-certs/cert.pem --key-file=test-certs/key.pem --http-tls=true --http-cert-file=test-certs/cert.pem --http-key-file=test-certs/key.pem --http-grpc-cert-file=test-certs/cert.pem

#### Step 3: Verify the initial certificate
echo "Port 17912:" && echo | openssl s_client -connect localhost:17912 2>/dev/null | openssl x509 -noout -subject
echo "Port 17913:" && echo | openssl s_client -connect localhost:17913 2>/dev/null | openssl x509 -noout -subject

#### Step 4: Update cert/key pair
openssl req -x509 -newkey rsa:2048 -keyout key2.pem -out cert2.pem -days 365 -nodes -subj "/CN=localhost-new"
cp cert2.pem cert.pem && cp key2.pem key.pem

#### Step 5: Check if the new certificate is used without restarting
echo "Port 17912:" && echo | openssl s_client -connect localhost:17912 2>/dev/null | openssl x509 -noout -subject
echo "Port 17913:" && echo | openssl s_client -connect localhost:17913 2>/dev/null | openssl x509 -noout -subject
 

### Checklist

- [x]  Implemented dynamic reloading of TLS certificates and keys.
- [x]   Added integration tests to verify reload behavior.
- [x]   Manually tested via openssl s_client.


banyand/liaison/grpc/server.go
```diff
@@ -323,6 +341,186 @@ func (s *server) GracefulStop() {
...
+// Stop gracefully stops the TLS reloader.
+func (tr *TLSReloader) Stop() {
+	close(tr.stopCh)
+	if err := tr.watcher.Close(); err != nil {
+		tr.log.Error().Err(err).Msg("Failed to close fsnotify watcher")
+	}
+}
+
+// DynamicTLSCredentials implements credentials.TransportCredentials for dynamic reloading.
+type DynamicTLSCredentials struct {
```
hanahmily, line 476: I prefer using `tls.NewTLS` to initialize "TransportCredentials" instead of a custom one. You can dynamically load certificates using "GetCertificate" in `tls.Config`.

banyand/liaison/grpc/server.go
```diff
@@ -323,6 +341,186 @@ func (s *server) GracefulStop() {
...
+func (tr *TLSReloader) watchFiles() {
+	tr.log.Info().Msg("TLS file watcher loop started for gRPC")
+	for {
+		select {
+		case event, ok := <-tr.watcher.Events:
+			if !ok {
+				tr.log.Warn().Msg("Watcher events channel closed unexpectedly")
+				return
+			}
+			if event.Op&fsnotify.Remove == fsnotify.Remove {
```
hanahmily, line 414: Why do you watch the "remove" operation?

banyand/liaison/grpc/server.go
```diff
@@ -323,6 +341,186 @@ func (s *server) GracefulStop() {
 	}
 }
 
+// TLSReloader manages dynamic reloading of TLS certificates and keys for the gRPC server.
+type TLSReloader struct {
+	watcher  *fsnotify.Watcher // 8 bytes
+	cert     *tls.Certificate  // 8 bytes
+	log      *logger.Logger    // 8 bytes
+	stopCh   chan struct{}     // 8 bytes
```
hanahmily, line 349: stopCh seems redundant since you can exit the loop by closing the watcher.

banyand/liaison/http/server.go
```diff
@@ -265,3 +307,170 @@ func serveFileContents(file string, files http.FileSystem) http.HandlerFunc {
 		http.ServeContent(w, r, fi.Name(), fi.ModTime(), index)
 	}
 }
+
+// TLSReloader manages dynamic reloading of TLS certificates and keys for the HTTP server.
+type TLSReloader struct {
```
hanahmily, line 312: It duplicates the one in gRPC. Please move it to a pkg package to share between gRPC and HTTP.

onkar717: ok Thanks

banyand/liaison/grpc/server.go
```diff
@@ -232,18 +242,18 @@ func (s *server) Validate() error {
...
 func (s *server) Serve() run.StopNotify {
 	var opts []grpclib.ServerOption
 	if s.tls {
-		opts = []grpclib.ServerOption{grpclib.Creds(s.creds)}
+		if err := s.tlsReloader.Start(); err != nil {
+			s.log.Error().Err(err).Msg("Failed to start TLSReloader for gRPC")
+			close(s.stopCh)
+			return s.stopCh
+		}
+		opts = append(opts, grpclib.Creds(s.tlsReloader.GetGRPCTransportCredentials()))
```
hanahmily, line 256: The gRPC server cannot load the updated credentials.\

You can try below:
```
        tlsConfig := &tls.Config{
		GetCertificate: s.tlsReloader.GetGRPCTransportCredentials(),
		MinVersion:     tls.VersionTLS12,
	}
	
	creds := credentials.NewTLS(tlsConfig)
```

pkg/tls/reloader.go
```diff
@@ -0,0 +1,210 @@
...
+		case event, ok := <-r.watcher.Events:
+			if !ok {
+				r.log.Warn().Msg("Watcher events channel closed unexpectedly")
+				return
+			}
+
+			// Log the event for debugging
+			r.log.Debug().Str("file", event.Name).Str("op", event.Op.String()).Msg("Detected file event")
+
+			// We need to watch for Remove operations because certificate rotation tools often
```
hanahmily, line 117: If so, please add a unit test to verify this behaviour. And remove the comment from this file. I prefer to use code to explain the behaviour.

test/docker/base-compose.yml
```diff
@@ -59,15 +59,18 @@ services:
       - sw_agent:/skywalking-java-agent
 
   oap:
-    # TODO: Update the image to the latest version
-    image: "docker.io/hanahmily/oap:ca528d984506faee1276232f4ebd31eaba9db40c"
+    # Updated image to ensure compatibility with swctl tool
+    image: "apache/skywalking-oap-server:${SW_OAP_COMMIT:-9.6.0}"
```
hanahmily, line 63: Rollback it. 9.6.0 doesn't work with banyandb at all.

hanahmily (changes requested):
Please revert all your changes to the test. The current test can run successfully without your modifications. If it fails, the only reason is your changes.

pkg/tls/reloader.go
onkar717: ok

hanahmily (changes requested):

pkg/tls/reloader.go
```diff
@@ -0,0 +1,200 @@
...
+
+// getCertificate returns the current TLS certificate for TLS Config's GetCertificate callback.
+func (r *Reloader) getCertificate(*tls.ClientHelloInfo) (*tls.Certificate, error) {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	return r.cert, nil
+}
+
+// GetCertificateForTLS returns the current TLS certificate.
+func (r *Reloader) GetCertificateForTLS() *tls.Certificate {
```
hanahmily, line 174: No references

pkg/tls/reloader.go
```diff
@@ -0,0 +1,200 @@
...
+func (r *Reloader) GetTLSConfig() *tls.Config {
+	return &tls.Config{
+		GetCertificate: r.getCertificate,
+		MinVersion:     tls.VersionTLS12,
+		NextProtos:     []string{"h2"},
+	}
+}
+
+// GetGRPCTransportCredentials returns transport credentials for gRPC using this reloader.
+func (r *Reloader) GetGRPCTransportCredentials() credentials.TransportCredentials {
```
hanahmily, line 198: No references as well

pkg/tls/reloader.go
```diff
@@ -0,0 +1,200 @@
...
+
+	"github.com/fsnotify/fsnotify"
+	"github.com/pkg/errors"
+	"google.golang.org/grpc/credentials"
+
+	"github.com/apache/skywalking-banyandb/pkg/logger"
+)
+
+// Config contains TLS configuration options.
+type Config struct {
```
hanahmily, line 33: Where is the Config referred to?

pkg/tls/reloader.go
```diff
@@ -0,0 +1,200 @@
...
+	return nil
+}
+
+func (r *Reloader) watchFiles() {
+	r.log.Info().Msg("TLS file watcher loop started")
+	for {
+		select {
+		case event, ok := <-r.watcher.Events:
+			if !ok {
+				r.log.Warn().Msg("Watcher events channel closed unexpectedly")
```
hanahmily, line 107:
```suggestion
				r.log.Info().Msg("Watcher events channel closed")
```

pkg/tls/reloader_test.go
```diff
@@ -0,0 +1,246 @@
...
+	certPEM, keyPEM := generateSelfSignedCert(t, "initial.local")
+	require.NoError(t, os.WriteFile(certFile, certPEM, 0o600))
+	require.NoError(t, os.WriteFile(keyFile, keyPEM, 0o600))
+
+	// Create reloader
+	log := logger.GetLogger("tls-test")
+	reloader, err := NewReloader(certFile, keyFile, log)
+	require.NoError(t, err)
+	require.NoError(t, reloader.Start())
+	defer reloader.Stop()
```
hanahmily, line 137: This section should be included in each `t.Run` to prevent accidental changes by subcases.

pkg/tls/reloader_test.go
```diff
@@ -0,0 +1,246 @@
...
+		time.Sleep(500 * time.Millisecond)
+
+		// Create new files with different content
+		certPEM, keyPEM = generateSelfSignedCert(t, "recreated.local")
+		require.NoError(t, os.WriteFile(certFile, certPEM, 0o600))
+		require.NoError(t, os.WriteFile(keyFile, keyPEM, 0o600))
+		time.Sleep(500 * time.Millisecond)
+
+		// Force reload after file changes
+		require.NoError(t, reloader.reloadCertificate())
```
hanahmily, line 153: Reloader doesn't work fine if you have to reload the certs manually.

pkg/tls/reloader.go
```diff
@@ -0,0 +1,200 @@
...
+			if !ok {
+				r.log.Warn().Msg("Watcher errors channel closed unexpectedly")
+				return
+			}
+			r.log.Error().Err(err).Msg("Error in file watcher")
+		}
+	}
+}
+
+func (r *Reloader) reloadCertificate() error {
```
hanahmily, line 151: Why do you need to reload certs manually?

pkg/tls/reloader_test.go
```diff
@@ -0,0 +1,246 @@
...
+
+		cert, err := reloader.getCertificate(nil)
+		require.NoError(t, err)
+		leafCert, err := x509.ParseCertificate(cert.Certificate[0])
+		require.NoError(t, err)
+		assert.Equal(t, "recreated.local", leafCert.Subject.CommonName)
+	})
+
+	// Test 2: Create files after the reloader starts
+	t.Run("Create files after reloader starts", func(t *testing.T) {
```
hanahmily, line 163: The loader is created in the parent test. You need to refactor the test case entirely..

hanahmily (changes requested):
The test cases are mostly incorrect. You cannot load the certificates manually to assist the reloader, which does not verify the watcher.

pkg/tls/reloader_test.go
```diff
@@ -0,0 +1,953 @@
...
+	require.NoError(t, err)
+	leafCert, err := x509.ParseCertificate(cert.Certificate[0])
+	require.NoError(t, err)
+	require.Equal(t, "test1.local", leafCert.Subject.CommonName)
+
+	// Create a new directory and move files there to trigger watcher
+	newDir := filepath.Join(tempDir, "new")
+	require.NoError(t, os.Mkdir(newDir, 0o755))
+
+	// Move files to new location
```
hanahmily, line 113: Could you please explain why you're moving the files? It's not a user case we discussed before.

pkg/tls/reloader_test.go
```diff
@@ -0,0 +1,953 @@
...
+	t.Run("invalid certificate files", func(t *testing.T) {
+		log := logger.GetLogger("tls-test")
+		reloader, err := NewReloader("nonexistent.pem", "nonexistent.pem", log)
+		assert.Error(t, err)
+		assert.Nil(t, reloader)
+		assert.Contains(t, err.Error(), "failed to load initial TLS certificate")
+	})
+}
+
+func TestReloader_InvalidCertificates(t *testing.T) {
```
hanahmily, line 364: The test appears to be a duplicate of "Create Invalid Files."

hanahmily (changes requested):
.The number of test cases is too large to review. Please limit them to less than five functions to cover the core scenarios I mentioned earlier.

pkg/tls/reloader_test.go
```diff
@@ -0,0 +1,376 @@
...
+		certFile := filepath.Join(tempDir, "cert.pem")
+		keyFile := filepath.Join(tempDir, "key.pem")
+
+		// Create valid certificate files
+		certPEM, keyPEM := generateSelfSignedCert(t, "test.local")
+		require.NoError(t, os.WriteFile(certFile, certPEM, 0o600))
+		require.NoError(t, os.WriteFile(keyFile, keyPEM, 0o600))
+
+		log := logger.GetLogger("tls-test")
+		reloader, err := NewReloader(certFile, keyFile, log)
```
hanahmily, line 89: You didn't start the reloader

pkg/tls/reloader_test.go
```diff
@@ -0,0 +1,376 @@
...
+
+		// Start reloader
+		err = reloader.Start()
+		require.NoError(t, err)
+
+		// Wait for initial certificate to be loaded
+		time.Sleep(100 * time.Millisecond)
+
+		// Remove the files
+		require.NoError(t, os.Remove(certFile))
```
hanahmily, line 196: You should verify the original cert before removing them. Like 

```
tlsConfig := reloader.GetTLSConfig()
		cert, err := tlsConfig.GetCertificate(nil)
		require.NoError(t, err)
		leafCert, err := x509.ParseCertificate(cert.Certificate[0])
		require.NoError(t, err)
		require.Equal(t, "test1.local", leafCert.Subject.CommonName)
```

pkg/tls/reloader_test.go
```diff
@@ -0,0 +1,376 @@
...
+		require.NoError(t, os.Remove(keyFile))
+		time.Sleep(100 * time.Millisecond)
+
+		// Create new files with different content
+		certPEM2, keyPEM2 := generateSelfSignedCert(t, "test3.local")
+		require.NoError(t, os.WriteFile(certFile, certPEM2, 0o600))
+		require.NoError(t, os.WriteFile(keyFile, keyPEM2, 0o600))
+		time.Sleep(500 * time.Millisecond)
```
hanahmily, line 204: You can use https://pkg.go.dev/github.com/stretchr/testify/assert#Eventually to wait until the certificates load. Please use "flags.EventuallyTimeout" to set the timeout.

pkg/tls/reloader_test.go
```diff
@@ -0,0 +1,376 @@
...
+		tlsConfig := reloader.GetTLSConfig()
+		cert, err := tlsConfig.GetCertificate(nil)
+		require.NoError(t, err)
+		leafCert, err := x509.ParseCertificate(cert.Certificate[0])
+		require.NoError(t, err)
+		require.Equal(t, "test3.local", leafCert.Subject.CommonName)
+	})
+
+	// Test creating files after reloader starts
+	t.Run("creating files after reloader starts", func(t *testing.T) {
```
hanahmily, line 216: The reloader can't wait for the absent files. I think that's fine for the use case. Let's delete this test case.

pkg/tls/reloader_test.go
```diff
@@ -0,0 +1,376 @@
...
+		require.NoError(t, err)
+		leafCert, err = x509.ParseCertificate(cert.Certificate[0])
+		require.NoError(t, err)
+		assert.Equal(t, "remove-test.local", leafCert.Subject.CommonName)
+	})
+}
+
+// TestReloaderInvalidCertificate tests handling of invalid certificates.
+func TestReloaderInvalidCertificate(t *testing.T) {
+	t.Run("handling invalid certificates", func(t *testing.T) {
```
hanahmily, line 296: Please remove the t.Run

hanahmily (changes requested):
some nits.

hanahmily:
@onkar717 

The reloader seems good to me, although there are a few minor issues to address. Let's move on to the next test:

Update https://github.com/apache/skywalking-banyandb/blob/main/test/integration/standalone/other/tls_test.go#L76.

Add the following: `g.It("queries an updated TLS server")`. In the test function, please do the following:

1. Create new certificate and key files, similar to the existing ones.
2. Establish a new connection using these new files.
3. Finally, run the verification function as shown below:

```go
gm.Eventually(func(innerGm gm.Gomega) {
    casesMeasureData.VerifyFn(innerGm, helpers.SharedContext{
        Connection: conn,
        BaseTime:   baseTime,
    }, helpers.Args{Input: "all", Duration: 25 * time.Minute, Offset: -20 * time.Minute})
}, flags.EventuallyTimeout).Should(gm.Succeed())
```

test/integration/standalone/other/tls_test.go
```diff
@@ -74,4 +77,79 @@ var _ = g.Describe("Query service_cpm_minute", func() {
 			}, helpers.Args{Input: "all", Duration: 25 * time.Minute, Offset: -20 * time.Minute})
 		}, flags.EventuallyTimeout).Should(gm.Succeed())
 	})
+
+	g.It("queries an updated TLS server", func() {
```
hanahmily, line 81: Please add a new test titled "Queries an Updated HTTP Server":

1. In the line "addr, _, deferFn = setup.StandaloneWithTLS", the second returned value (which is ignored) represents the HTTP address.
2. You can utilize the [HTTP health checker](https://github.com/apache/skywalking-banyandb/blob/main/pkg/test/helpers/http_health.go) to verify the functionality of the HTTP server in this scenario.
3. The procedure should be the same as for the "Queries an Updated TLS Server."

hanahmily (changes requested):
The last step should test the HTTP server.

banyand/liaison/http/server.go
```diff
@@ -69,6 +70,7 @@ type Server interface {
 
 type server struct {
```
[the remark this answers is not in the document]
onkar717: Yes right 👍

test/integration/standalone/other/tls_test.go
onkar717: Ok

banyand/liaison/http/server.go
```diff
@@ -134,23 +146,148 @@ func (p *server) GetPort() *uint32 {
...
+		host, _, err := net.SplitHostPort(p.grpcAddr)
+		if err != nil {
+			p.l.Error().Err(err).Msg("Failed to split gRPC address")
+			return err
+		}
+		if host == "" || host == "0.0.0.0" || host == "[::]" {
+			host = "localhost"
+		}
+
+		p.creds = credentials.NewTLS(&tls.Config{
```
hanahmily, line 196: The new credentials are not applied to the client because they are created statically.

Your test passed, but it cannot verify that the credentials are updated. The gRPC connection between the HTTP server and the gRPC server is long-lived. Even if the server's credentials are updated, the connection does not close and recreate itself, which is why your test does not fail.

banyand/liaison/http/server.go
```diff
@@ -134,23 +146,148 @@ func (p *server) GetPort() *uint32 {
...
+		if err != nil {
+			p.l.Error().Err(err).Msg("Failed to initialize TLSReloader for HTTP")
+			return err
+		}
+	} else {
+		p.l.Warn().Msg("HTTP TLS is disabled, skipping TLSReloader initialization")
+	}
+
+	// Initialize gRPC client with cert file
+	if p.grpcCert != "" {
```
hanahmily, line 169: Could you use the reloader for reloading? The reloader can act as either a server or a client. If only credentials are passed to the reloader, it can generate the client-side TLS configuration.

banyand/liaison/http/server.go
```diff
@@ -134,23 +146,148 @@ func (p *server) GetPort() *uint32 {
...
+
+		// Start a goroutine to watch for certificate changes
+		go func() {
+			for {
+				select {
+				case event, ok := <-p.certWatcher.Events:
+					if !ok {
+						return
+					}
+					if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) {
```
hanahmily, line 225: The "Remove" event is ignored. This is another reason to use the reloader, which already supports handling this event.

test/integration/standalone/other/tls_test.go
```diff
@@ -74,4 +81,203 @@ var _ = g.Describe("Query service_cpm_minute", func() {
...
+			},
+		}
+		// Add the new cert to the pool
+		newCertPool := newHTTPClient.Transport.(*http.Transport).TLSClientConfig.RootCAs
+		newCertPool.AppendCertsFromPEM(certPEM)
+
+		// Verify HTTP server works with the new certificates
+		gm.Eventually(func() error {
+			var respErr2 error
+			resp, respErr2 := newHTTPClient.Get(healthCheckURL)
```
hanahmily, line 257: Since the health check is a long-lived connection, please use another endpoint https://github.com/apache/skywalking-banyandb/blob/main/api/proto/banyandb/database/v1/rpc.proto#L104 to verify the HTTP server.

test/integration/standalone/other/tls_test.go
```diff
@@ -74,4 +81,203 @@ var _ = g.Describe("Query service_cpm_minute", func() {
...
+		// Verify using the connection with new certificates
+		gm.Eventually(func(innerGm gm.Gomega) {
+			casesMeasureData.VerifyFn(innerGm, helpers.SharedContext{
+				Connection: newConn,
+				BaseTime:   testBaseTime,
+			}, helpers.Args{Input: "all", Duration: 25 * time.Minute, Offset: -20 * time.Minute})
+		}, flags.EventuallyTimeout).Should(gm.Succeed())
+	})
+
+	g.It("queries an updated HTTP server", func() {
```
hanahmily, line 160: You can replace this test with the previous one since it verifies both HTTP and gRPC.

hanahmily (changes requested):

hanahmily (approved):

PR #645, opened 2025-04-12
Fix a bug and update API

hanahmily:
The details are laid out in the change log.

The Java client and UI should be updated accordingly.

### DIFF
--- a/api/proto/banyandb/common/v1/common.proto
+++ b/api/proto/banyandb/common/v1/common.proto
@@ -93,8 +93,8 @@ message ResourceOpts {
   IntervalRule ttl = 3;
   // stages defines the ordered lifecycle stages. Data progresses through these stages sequentially.
   repeated LifecycleStage stages = 4;
-  // default_node_selector is the default node selector for queries if node_selector is not specified
-  string default_node_selector = 5;
+  // default_stage is the name of the default stage
+  string default_stage = 5;
 }
 
 // Group is an internal object for Group management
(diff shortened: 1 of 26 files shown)


hanahmily:
e2e failed due to java client's break

hanahmily:
@wu-sheng @wankai123, Would you please take a look at the tweaked API? If you want those, I will build a custom OAP to temporarily make the e2e pass.

api/proto/banyandb/common/v1/common.proto
```diff
@@ -93,8 +93,8 @@ message ResourceOpts {
   IntervalRule ttl = 3;
   // stages defines the ordered lifecycle stages. Data progresses through these stages sequentially.
   repeated LifecycleStage stages = 4;
-  // default_node_selector is the default node selector for queries if node_selector is not specified
-  string default_node_selector = 5;
+  // default_stage is the name of the default stage
+  string default_stage = 5;
```
wankai123, line 97: if the default query stages are `hot` and `warm`, how to config this?

api/proto/banyandb/measure/v1/query.proto
```diff
@@ -111,6 +111,6 @@ message QueryRequest {
   model.v1.QueryOrder order_by = 12;
   // trace is used to enable trace for the query
   bool trace = 13;
-  // node_selector is used to specify the target node for the query
-  string node_selector = 14;
+  // stage is used to specify the stage of the data points in the lifecycle
+  string stage = 14;
```
wankai123, line 115: can we query stages list?

api/proto/banyandb/common/v1/common.proto
hanahmily: I will convert it to a list.

api/proto/banyandb/measure/v1/query.proto
hanahmily: You can use get group schema to fetch all stages

wu-sheng (approved):

PR #650, opened 2025-04-18
Change `node_selector` to `stages` for groups in UI.

Fine0830:
- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Closes apache/skywalking#13176

Video

https://github.com/user-attachments/assets/a09e7e93-a2a7-42f5-9978-5f60c73f60ee

Signed-off-by: Qiuxia Fan<qiuxiafan@apache.org>

### DIFF
--- a/ui/src/components/GroupTree/data.js
+++ b/ui/src/components/GroupTree/data.js
@@ -24,7 +24,7 @@ export const StageFields = [
   { label: 'TTL number', key: 'ttlNum' },
   { label: 'Segment interval unit', key: 'segmentIntervalUnit' },
   { label: 'Segment interval number', key: 'segmentIntervalNum' },
-  { label: 'Stages', key: 'stages' },
+  { label: 'Node Selector', key: 'nodeSelector' },
   { label: 'Close', key: 'close' },
 ];
 export const StageConfig = {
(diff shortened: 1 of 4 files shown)


ui/src/components/GroupTree/data.js
```diff
@@ -24,7 +24,7 @@ export const StageFields = [
   { label: 'TTL number', key: 'ttlNum' },
   { label: 'Segment interval unit', key: 'segmentIntervalUnit' },
   { label: 'Segment interval number', key: 'segmentIntervalNum' },
-  { label: 'Node selector', key: 'nodeSelector' },
+  { label: 'Stages', key: 'stages' },
```
hanahmily, line 27: It is not "nodeSelector"; it should be "defaultNodeSelector." This term exists in the "ResourceOpt" of "Group," but does not exist in "Stage."

hanahmily (changes requested):

ui/src/components/GroupTree/data.js
Fine0830: hmm....Currently, there doesn't seem to be a `defaultNodeSelector` in UI. Will add directly the `stages ` here.

hanahmily (approved):

PR #654, opened 2025-04-21
feat(metadata):expose auto-compaction

raiki02:
Expose "AutoCompactionMode" and "AutoCompactionRetention" of embedded etcd and add tests.


wu-sheng:
@hanahmily I think we should have default values for this? We don't need historical data actually.

wu-sheng:
I think this is related to https://github.com/apache/skywalking/issues/13182

hanahmily:
@raiki02 Please expose "--quota-backend-bytes" as well.

hanahmily:
> @hanahmily I think we should have default values for this? We don't need historical data actually.

It defaults to "periodic" and "1h"

hanahmily:
@raiki02 and you need to introduce https://etcd.io/docs/v3.5/op-guide/maintenance/#defragmentation. The etcd client may support defragmentation. You can use https://github.com/apache/skywalking-banyandb/blob/main/pkg/timestamp/scheduler.go#L48 to run the defragmentation process daily. Additionally, the cron expression for this schedule should be set as a flag in the etcd server as well.

raiki02:
Is it ok for '--quota-backend-bytes' to be set 2GiB as a default value ?

hanahmily:
> Is it ok for '--quota-backend-bytes' to be set 2GiB as a default value ?

The default value is good for me.

wu-sheng:
CI fails, please fix.

wu-sheng:
As the configuration updated, I think we should update the docs here. @hanahmily Is this doc? docs/operation/configuration.md

hanahmily (changes requested):

raiki02:
It seems that the ci failed due to the output returned less than the template expected. Is it because the configuration of retry policy did not prepared long enough to collect the data and I need to edit the 'test/e2e-v2/cases/cluster/e2e.yaml' file ?

wu-sheng:
I rerun, it seems to pass.

hanahmily:
@raiki02 Your changes look good to me.  Let's add defragmentation now to free up disk space.

banyand/metadata/embeddedserver/server.go
```diff
@@ -81,12 +108,21 @@ func (s *server) PreRun(ctx context.Context) error {
...
 func (s *server) GracefulStop() {
 	s.Service.GracefulStop()
+	if s.ecli != nil {
+		_ = s.ecli.Close()
+	}
+	if s.scheduler != nil {
+		if !s.scheduler.Closed() {
+			s.scheduler.Close()
+		}
+	}
```
hanahmily, line 125:
```suggestion
       if s.scheduler != nil {
		s.scheduler.Close()
	}
	if s.ecli != nil {
		_ = s.ecli.Close()
	}
	s.Service.GracefulStop()	
```

banyand/metadata/embeddedserver/server.go
```diff
@@ -103,3 +139,38 @@ func NewService(_ context.Context) (metadata.Service, error) {
 	}
 	return s, nil
 }
+
+func (s *server) RegisterDefrag(ctx context.Context) {
```
hanahmily, line 143:
```suggestion
func (s *server) registerDefrag() {
```

banyand/metadata/embeddedserver/server.go
```diff
@@ -103,3 +139,38 @@ func NewService(_ context.Context) (metadata.Service, error) {
...
+	var (
+		err        error
+		etcdLogger = logger.GetLogger().Named("etcd-server")
+		defrag     = func(_ time.Time, _ *logger.Logger) bool {
+			for _, endpoint := range s.listenClientURL {
+				_, err = s.ecli.Defragment(ctx, endpoint)
```
hanahmily, line 149:
```suggestion
                                ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	                        defer cancel()
				_, err = s.ecli.Defragment(ctx, endpoint)
```

banyand/metadata/embeddedserver/server.go
```diff
@@ -103,3 +139,38 @@ func NewService(_ context.Context) (metadata.Service, error) {
...
+				if err != nil {
+					etcdLogger.Error().Err(err).Msg("failed to execute defragmentation")
+					return false
+				}
+			}
+			return true
+		}
+	)
+
+	if s.ecli == nil {
```
hanahmily, line 159:
```suggestion
```

If `s.ecli` is nil, it indicates that there is a bug present. You should notify it immediately.

banyand/metadata/embeddedserver/server.go
```diff
@@ -103,3 +139,38 @@ func NewService(_ context.Context) (metadata.Service, error) {
...
+		s.ecli, err = clientv3.New(clientv3.Config{
+			Endpoints: s.listenClientURL,
+		})
+		if err != nil {
+			etcdLogger.Error().Err(err).Msg("failed to create client")
+			return
+		}
+	}
+	if s.scheduler == nil {
```
hanahmily, line 168:
```suggestion
```

docs/operation/configuration.md
```diff
@@ -111,6 +111,10 @@ The following flags are used to configure the stream storage engine:
 The following flags are used to configure the embedded etcd storage engine which is only used when running as a standalone server:
 
 - `--metadata-root-path string`: The root path of metadata (default: "/tmp").
+- `--etcd-auto-compaction-mode string`: Auto compaction mode (default: "periodic").
```
hanahmily, line 114: Update comments

hanahmily (changes requested):

banyand/metadata/embeddedserver/server.go
```diff
@@ -103,3 +139,37 @@ func NewService(_ context.Context) (metadata.Service, error) {
 	}
 	return s, nil
 }
+
+func (s *server) RegisterDefrag() {
+	var (
+		err        error
+		etcdLogger = logger.GetLogger().Named("etcd-server")
+		defrag     = func(_ time.Time, _ *logger.Logger) bool {
```
hanahmily, line 147: Please test the defragmentation (defrag) function:

1. Extract the existing logic into a new, standalone function.
2. Create a new defrag function that takes two parameters: "listenURL" and "ecli."
3. Ensure that the function returns an error.

Next, start the embedded server to verify that the defrag function does not return any errors.

banyand/metadata/embeddedserver/server.go
```diff
@@ -103,3 +139,37 @@ func NewService(_ context.Context) (metadata.Service, error) {
 	}
 	return s, nil
 }
+
+func (s *server) RegisterDefrag() {
```
hanahmily, line 143:
```suggestion
func (s *server) registerDefrag() {
```

Please don't expose this function.

banyand/metadata/embeddedserver/server.go
```diff
@@ -82,10 +109,19 @@ func (s *server) PreRun(ctx context.Context) error {
 
 func (s *server) Serve() run.StopNotify {
 	_ = s.Service.Serve()
+	s.RegisterDefrag()
 	return s.metaServer.StoppingNotify()
 }
 
 func (s *server) GracefulStop() {
+	if s.scheduler != nil {
+		if !s.scheduler.Closed() {
```
hanahmily, line 118:
```suggestion
```

You don't have to check if it has been closed.

hanahmily (changes requested):

hanahmily (approved):
LGTM

PR #663, opened 2025-05-01
bug/fix Panic when creating _top_n_result #13183

OmCheeLin:
### Fix Panic when creating _top_n_result

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13183.

Continuously retry to avoids panic until metadata is repaired.
1. After approximately 14 retries, the TopN creation retry interval backs off to 60 seconds, ensuring minimal resource contention
2. If during the topn write phase, it is determined that the _top_n_result measure exists, canceling the retry may increase code complexity.

fix some problems:
1. docs/installation/cluster.md : the `storage` command may be deprecated
2. bydbctl/internal/cmd/use.go: the `newUserCmd` may be `newUseCmd`

### DIFF
--- a/banyand/measure/metadata.go
+++ b/banyand/measure/metadata.go
@@ -323,7 +323,7 @@ func (sr *schemaRepo) createTopNResultMeasure(ctx context.Context, measureSchema
 	}
 
 	backoffStrategy := backoff.NewExponentialBackOff()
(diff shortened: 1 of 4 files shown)


wu-sheng:
Please resolve conflicts.

PR #666, opened 2025-05-10
Add documentation for skipping index

ButterBright:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#11475.


docs/concept/data-model.md
```diff
@@ -264,7 +264,7 @@ tags:
 type: TYPE_INVERTED
 ```
 
-IndexRule supports selecting two distinct kinds of index structures. The `INVERTED` index is the primary option when users set up an index rule. It's suitable for most tag indexing due to a better memory usage ratio and query performance.
+IndexRule supports two distinct kinds of index structures. The `INVERTED` index is suitable for measure tag indexing due to better query performance. The `SKIPPING` index is optimized for the majority of stream tags, which prioritizes efficient space utilization.
```
hanahmily:
```suggestion
IndexRule supports several distinct kinds of index structures. The `INVERTED` index is suitable for measure tag indexing due to better query performance. The `SKIPPING` index is optimized for the majority of stream tags, which prioritizes efficient space utilization. The `TREE` index is designed for storing hierarchical data.
```

docs/concept/tsdb.md
```diff
@@ -68,10 +68,10 @@ The diagram below shows the detailed fields within each block. The block is the
 
 In measure's timestamp file, there are version fields to record the version of the data. The version field is used to deduplicate. It determine the latest data when the data's timestamp are identical. Only the latest data will be returned to the user.
 
-![measure-block](https://skywalking.apache.org/doc-graph/banyandb/v0.7.0/measure-block.png)
+![measure-block](https://skywalking.apache.org/doc-graph/banyandb/v0.9.0/measure-block.png)
 
 Unlike the measure, there are element ids in the stream's timestamp file. The element id is used to identify the data of the same series. The data with the same timestamp but different element id will both be stored in the TSDB.
```
hanahmily:
```suggestion
Unlike the measure, there are element ids in the stream's timestamp file. The element id is used to identify the data of the same series. The data with the same timestamp but different element id will both be stored in the TSDB. This introduces a series of new files, named "*.tff", which contain bloom filters for each tag, enabling efficient skipping of irrelevant data. Additionally, min/max fields are added to the "*.tfm" file to further aid in skipping blocks.```

hanahmily (approved):

PR #674, opened 2025-05-26
fix potential oom

OmCheeLin:
### Fix potential oom

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13199.

add function AvailableBytes()

### DIFF
--- a/banyand/internal/test/mock_memory_protector.go
+++ b/banyand/internal/test/mock_memory_protector.go
@@ -1,0 +1,82 @@
+// Licensed to Apache Software Foundation (ASF) under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Apache Software Foundation (ASF) licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+// Package test provides mock implementations for testing.
+package test
+
+import (
+	"context"
+
+	"github.com/apache/skywalking-banyandb/pkg/run"
+)
+
+// MockMemoryProtector is a mock implementation of a memory protector.
+type MockMemoryProtector struct {
+	acquireErr          error
+	limit               uint64
+	ExpectQuotaExceeded bool
+}
+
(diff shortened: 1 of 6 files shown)


hanahmily:
@OmCheeLin, please run "make check" locally before pushing.

banyand/protector/protector.go
```diff
@@ -37,6 +37,21 @@ import (
 
 var scope = observability.RootScope.SubScope("memory_protector")
 
+// MemoryProtector is an interface for monitoring and limiting memory usage to prevent OOM.
+type MemoryProtector interface {
```
hanahmily, line 41:
```suggestion
// Memory is an interface for monitoring and limiting memory usage to prevent OOM.
type Memory interface {
```

hanahmily: and convert the struct's name from `Memory` to `memory`

banyand/protector/protector.go
```diff
@@ -37,6 +37,21 @@ import (
...
+	AvailableBytes() int64
+	GetLimit() uint64
+	AcquireResource(ctx context.Context, size uint64) error
+	Name() string
+	FlagSet() *run.FlagSet
+	Validate() error
+	PreRun(ctx context.Context) error
+	GracefulStop()
+	Serve() run.StopNotify
```
hanahmily, line 50:
```suggestion
	run.PreRunner
	run.Config
	run.Service
```

banyand/protector/protector.go
```diff
@@ -103,6 +118,18 @@ func (m *Memory) GetLimit() uint64 {
 	return m.limit
 }
 
+// AvailableBytes returns the available memory (limit - usage).
+func (m *Memory) AvailableBytes() int64 {
+	if m.limit == 0 {
```
hanahmily, line 123: `limit` should be atomic type.

banyand/stream/query_by_idx.go
hanahmily, line 1: `blockScanner.scan` should be changed as well

docs/concept/clustering.md
```diff
@@ -111,7 +111,7 @@ A measure named `service_cpm` belonging to `measure-minute` with an entity of `s
 
 Similarly, a stream named `system_log` belonging to `stream-log` with an entity combination of `service_id` "frontend" and `instance_id` "10.0.0.1" will be written to a specific shard based on the hashed value of the sharding key `system_log:frontend|10.0.0.1`.
 
-> Note: If there are ":" or "|" in the entity, they will be prefixed with a backslash "\".
+> Note: If there are ":" or "|" in the entity, they will be prefixed with a backslash "\\".
```
hanahmily, line 114: Why do you change it?

hanahmily (changes requested):

banyand/protector/protector.go
hanahmily: Yes, the role of the interface is to conceal the structure. If you introduce the interface, the exported structure no longer makes sense.

PR #681, opened 2025-06-12
Implement versioning properties and replace physical deletion with the tombstone mechanism for the property database

mrproliu:
1. Using the https://github.com/apache/skywalking-banyandb/blob/main/api/proto/banyandb/common/v1/common.proto#L47 as the versioning item for property items. 
2. Add `deleted` field in the `index.Document` to mark the document is deleted or not(not physical).
3. Update the property query protocol between liaison and data nodes for the tombstone mechanism query.
4.  Disable all physical property deletion mechanism(mark as TODO), will fix in future PR.
5. Notify all replicas when apply the property. 
6. Added an IT for checking the whole process of property operation. 

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13279.


banyand/liaison/grpc/property.go
```diff
@@ -150,38 +150,37 @@ func (ps *propertyServer) Apply(ctx context.Context, req *propertyv1.ApplyReques
...
-				var ids [][]byte
-				for _, p := range qResp.Properties {
-					ids = append(ids, getPropertyID(p))
-				}
-				if err = ps.remove(ids); err != nil {
-					err = multierr.Append(err, errors.New("fail to remove old properties"))
-				}
-			}
-		}()
+		// TODO(mrproliu) find older property by different version, should be remove it
```
hanahmily: You can do it in apache/skywalking#13281

test/cases/property/property.go
hanahmily, line 1: The test cases should be written in `bydbctl/internal/property_test.go`.

hanahmily (changes requested):

wu-sheng:
Need to fix CI.

bydbctl/internal/cmd/property_test.go
```diff
@@ -126,6 +126,20 @@ tags:
 		Expect(out).To(ContainSubstring("tagsNum: 2"))
 	})
 
+	It("update property", func() {
```
hanahmily, line 96: More tests should be introduced:

### 1. Query: Select the latest version.
You should set up a cluster with different versions:

1. Create two test directories to save the same key with different versions. You can use standalone to do it.
2. Start up two data nodes(setup.DataNode) on them 
3. Start up a liaison node(setup.LiaisonNode) to connect the two data nodes.
4. Query the property through the liaison node to get the latest version.

### 2. Delete:

1. As the Query's first step. But you should delete one of the versions.
2. If the latest version is deleted, the query should return an empty result.
3. Else, the query should return a property.

hanahmily (changes requested):

pkg/test/setup/setup.go
```diff
@@ -236,48 +233,58 @@ func startDataNode(etcdEndpoint string, flags ...string) (string, string, func()
 		return helpers.ListKeys(etcdEndpoint, fmt.Sprintf("/%s/nodes/%s:%d", metadata.DefaultNamespace, nodeHost, ports[0]))
 	}, testflags.EventuallyTimeout).Should(gomega.HaveLen(1))
 
-	return addr, path, closeFn, deferFn
+	return addr, dataDir, closeFn
```
hanahmily:
```suggestion
	return addr, closeFn
```

pkg/test/setup/setup.go
```diff
@@ -204,10 +204,7 @@ func CMD(flags ...string) func() {
 	}
 }
 
-func startDataNode(etcdEndpoint string, flags ...string) (string, string, func(), func()) {
-	path, deferFn, err := test.NewSpace()
-	gomega.Expect(err).NotTo(gomega.HaveOccurred())
-
+func startDataNode(etcdEndpoint, dataDir string, flags ...string) (string, string, func()) {
```
hanahmily:
```suggestion
func startDataNode(etcdEndpoint, dataDir string, flags ...string) (string, func()) {
```

pkg/test/setup/setup.go
```diff
@@ -236,48 +233,58 @@ func startDataNode(etcdEndpoint string, flags ...string) (string, string, func()
...
+	addr, dir, closeFn := startDataNode(etcdEndpoint, path, flags...)
 	return addr, dir, func() {
 		closeFn()
 		deferFn()
 	}
 }
 
 // LiaisonNode runs a liaison node.
-func LiaisonNode(etcdEndpoint string, flags ...string) (string, func()) {
+func LiaisonNode(etcdEndpoint string, flags ...string) (grpcAddr, httpAddr string, closeFn func()) {
```
hanahmily: Could you add a "LiaisonNodeWithHTTP" to expose the httpAddr instead of changing this function's signature?

hanahmily (changes requested):

hanahmily (approved):

PR #682, opened 2025-06-12
Implement dictionary encoding

ButterBright:
Benchmark results:
BenchmarkEncodeDictionary/size=1000_cardinality=100-8         	   30502	     38081 ns/op	         3.992 compress_ratio	   25208 B/op	      19 allocs/op
BenchmarkEncodeDictionary/size=10000_cardinality=100-8        	    3502	    340304 ns/op	         4.456 compress_ratio	  436341 B/op	      31 allocs/op
BenchmarkEncodeDictionary/size=100000_cardinality=100-8       	     356	   3482206 ns/op	         4.508 compress_ratio	 4640720 B/op	      49 allocs/op

BenchmarkDecodeDictionary/size=1000_cardinality=100-8         	   38172	     30559 ns/op	   42675 B/op	      33 allocs/op
BenchmarkDecodeDictionary/size=10000_cardinality=100-8        	    4405	    270175 ns/op	  521599 B/op	      49 allocs/op
BenchmarkDecodeDictionary/size=100000_cardinality=100-8       	     452	   2648951 ns/op	 6098789 B/op	      82 allocs/op

- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13272.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,13 +14,17 @@ Release Notes.
 - Replica: Support configurable replica count on Group.
 - Replica: Move the TopN pre-calculation flow from the Data Node to the Liaison Node.
 - Add a wait and retry to write handlers to avoid the local metadata cache being loaded.
+- Implement primary block cache for measure.
 - Implement versioning properties and replace physical deletion with the tombstone mechanism for the property database.
+- Add Load Balancer Feature to Liaison. 
+- Implement fadvise for large files to prevent page cache pollution.
 
 ### Bug Fixes
 
 - Fix the deadlock issue when loading a closed segment.
 - Fix the issue that the etcd watcher gets the historical node registration events.
 - Fix the crash when collecting the metrics from a closed segment.
+- Fix topN parsing panic when the criteria is set.
 
 ## 0.8.0
 
--- a/api/proto/banyandb/property/v1/rpc.proto
+++ b/api/proto/banyandb/property/v1/rpc.proto
@@ -122,7 +122,8 @@ message InternalDeleteRequest {
 message InternalQueryResponse {
   repeated bytes sources = 1;
   common.v1.Trace trace = 2;
-  // deletes indicates the property is deleted or not
+  // deletes indicates the property is deleted timestamps,
   // it's mapping to the sources in the same order
-  repeated bool deletes = 3;
+  // if the value is 0, it means the property is not deleted
+  repeated int64 deletes = 3;
 }
--- a/banyand/backup/lifecycle/service.go
+++ b/banyand/backup/lifecycle/service.go
@@ -170,6 +170,8 @@ func (l *lifecycleService) action() error {
 		return err
 	}
 
+	l.l.Info().Msgf("starting migration for %d groups: %v", len(groups), getGroupNames(groups))
+
 	if len(groups) == 0 {
 		l.l.Info().Msg("no groups to process, all groups already completed")
 		progress.Remove(l.progressFilePath, l.l)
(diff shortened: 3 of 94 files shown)


pkg/encoding/dictionary.go
```diff
@@ -0,0 +1,250 @@
...
+package encoding
+
+import (
+	"bytes"
+	"fmt"
+	"math/bits"
+)
+
+// Dictionary is used for dictionary encoding.
+type Dictionary struct {
```
hanahmily, line 29: You need to provide a method that indicates whether the block’s values are suitable for the dictionary encoding. If the values differ too significantly, we should skip this encoding.

pkg/encoding/dictionary.go
```diff
@@ -0,0 +1,250 @@
...
+			continue
+		}
+		dst = append(dst, data[:sLen])
+		data = data[sLen:]
+	}
+
+	return dst, tail, nil
+}
+
+func encodeRLE(dst []uint32, src []uint32) []uint32 {
```
hanahmily, line 132: The values suitable for a dictionary are brief. I don't think you'll achieve a better compression ratio. It was shown in your test case.

hanahmily (changes requested):

pkg/encoding/dictionary.go
```diff
@@ -0,0 +1,250 @@
...
+
+// Reset resets the dictionary.
+func (d *Dictionary) Reset() {
+	clear(d.m)
+	d.values = d.values[:0]
+	d.indices = d.indices[:0]
+}
+
+// Add adds a value to the dictionary.
+func (d *Dictionary) Add(value []byte) {
```
hanahmily:
```suggestion
func (d *Dictionary) Add(value []byte) bool {
```

If the size of unique values exceeds a threshold(128 or 256?), returning false indicates that the block cannot use dict encoding.

pkg/encoding/dictionary.go
```diff
@@ -0,0 +1,250 @@
...
+	index := uint32(len(d.values) - 1)
+	d.m[string(value)] = index
+	d.indices = append(d.indices, index)
+}
+
+// Encode encodes the dictionary.
+func (d *Dictionary) Encode(dst []byte) []byte {
+	dst = VarUint64ToBytes(dst, uint64(len(d.values)))
+	dst = EncodeBytesBlock(dst, d.values)
+	re := make([]uint32, 0)
```
hanahmily: Reuse this slice

pkg/encoding/dictionary.go
```diff
@@ -0,0 +1,250 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"math/bits"
+)
+
+// Dictionary is used for dictionary encoding.
+type Dictionary struct {
+	m       map[string]uint32
```
hanahmily:
```suggestion
```
You don't need a map. Instead, adding can loop values to find the index. The volume is reasonable for looping.

pkg/encoding/dictionary.go
```diff
@@ -0,0 +1,250 @@
...
+	if err != nil {
+		return err
+	}
+	d.values = values
+	d.m = make(map[string]uint32, len(d.values))
+	for i, value := range d.values {
+		d.m[string(value)] = uint32(i)
+	}
+
+	bd := make([]uint32, 0)
```
hanahmily: Reuse this slice. It can be field shared between Encode and Decode.

pkg/encoding/dictionary.go
```diff
@@ -0,0 +1,250 @@
...
+	dst = EncodeBytesBlock(dst, d.values)
+	re := make([]uint32, 0)
+	re = encodeRLE(re, d.indices)
+	be := encodeBitPacking(re)
+	dst = append(dst, be...)
+	return dst
+}
+
+// Decode decodes the dictionary.
+func (d *Dictionary) Decode(src []byte) error {
```
hanahmily: Would you write a benchmark for Encode and Decode?

hanahmily (changes requested):

PR #684, opened 2025-06-17
Physics delete property when merging segment

mrproliu:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13280.

### DIFF
--- a/banyand/property/service.go
+++ b/banyand/property/service.go
@@ -72,7 +72,7 @@ func (s *service) FlagSet() *run.FlagSet {
 	flagS.DurationVar(&s.flushTimeout, "property-flush-timeout", defaultFlushTimeout, "the memory data timeout of measure")
 	flagS.IntVar(&s.maxDiskUsagePercent, "property-max-disk-usage-percent", 95, "the maximum disk usage percentage allowed")
 	flagS.IntVar(&s.maxFileSnapshotNum, "property-max-file-snapshot-num", 2, "the maximum number of file snapshots allowed")
-	flagS.DurationVar(&s.expireTimeout, "property-expire-delete-timeout", time.Hour*24*7, "the duration of the expired data needs to be delete")
+	flagS.DurationVar(&s.expireTimeout, "property-expire-delete-timeout", time.Hour*24*7, "the duration of the expired data needs to be deleted")
 	return flagS
 }
 


banyand/property/shard.go
```diff
@@ -80,11 +80,12 @@ func (db *database) newShard(ctx context.Context, id common.ShardID, _ int64) (*
...
-		Logger:               si.l,
-		Metrics:              inverted.NewMetrics(db.omr.With(propertyScope.ConstLabels(meter.LabelPairs{"shard": sName}))),
-		BatchWaitSec:         0,
-		PrepareMergeCallback: si.prepareForMerge,
+		Path:                        location,
+		Logger:                      si.l,
+		Metrics:                     inverted.NewMetrics(db.omr.With(propertyScope.ConstLabels(meter.LabelPairs{"shard": sName}))),
+		BatchWaitSec:                0,
+		PrepareMergeCallback:        si.prepareForMerge,
+		MinSegmentsForInMemoryMerge: 1,
```
hanahmily: The merging does not start up when there is only one segment. 

Our design goal is to leverage the normal merging process when there are multiple segments (more than one). If only one segment exists, I am fine with leaving the deleted files within it.

mrproliu: Yeah, it's just works for make the testing easy. If all the segments should be deleted in the single merging phase, we facing the same problem.

mrproliu: I will deleted it after the testing finished.

wu-sheng: We should design carefully for the test of this feature. As discussed, the deletion may occur a few days after the logic deletion is flagged due to new merging. 

How long do we set for this physical deletion now?

hanahmily: After 7 days, I would like to propose.

wu-sheng: 7 days and only when new data is merging into?

hanahmily: > 7 days and only when new data is merging into?

Yep.

wu-sheng: Could we document this somewhere, as it is important for property type data?

mrproliu: This field have been deleted, only for testing.

wu-sheng: I mean for how long time, this physical deletion could happen. This is a config, right? Otherwise, we could face data conflict quickly.

banyand/property/shard.go
```diff
@@ -225,3 +228,33 @@ func (s *shard) search(ctx context.Context, indexQuery index.Query, limit int,
...
+			hasDeleted := false
+			_ = seg.VisitStoredFields(docID, func(field string, value []byte) bool {
+				if field == deleteField {
+					hasDeleted = convert.BytesToBool(value)
+					return false
+				}
+				return true
+			})
+
+			if !hasDeleted {
```
hanahmily: You should set a flag to control whether the data is removed after a fixed duration, with a default duration of 7 days.

hanahmily (changes requested):

banyand/property/service.go
```diff
@@ -71,6 +72,7 @@ func (s *service) FlagSet() *run.FlagSet {
 	flagS.DurationVar(&s.flushTimeout, "property-flush-timeout", defaultFlushTimeout, "the memory data timeout of measure")
 	flagS.IntVar(&s.maxDiskUsagePercent, "property-max-disk-usage-percent", 95, "the maximum disk usage percentage allowed")
 	flagS.IntVar(&s.maxFileSnapshotNum, "property-max-file-snapshot-num", 2, "the maximum number of file snapshots allowed")
+	flagS.DurationVar(&s.expireTimeout, "property-expire-delete-timeout", time.Hour*24*7, "the duration of the expired data need to be deleted")
```
wu-sheng:
```suggestion
	flagS.DurationVar(&s.expireTimeout, "property-expire-delete-timeout", time.Hour*24*7, "the duration of the expired data needs to be deleted")
```

wu-sheng (approved):

banyand/property/shard.go
```diff
@@ -225,3 +232,45 @@ func (s *shard) search(ctx context.Context, indexQuery index.Query, limit int,
 	}
 	return data, nil
 }
+
+func (s *shard) prepareForMerge(src []*roaring.Bitmap, segments []segment.Segment, _ uint64) (dest []*roaring.Bitmap, err error) {
```
hanahmily, line 243: If there is no error returned, why was an error return value added?

banyand/property/shard.go
```diff
@@ -225,3 +232,45 @@ func (s *shard) search(ctx context.Context, indexQuery index.Query, limit int,
...
+	if len(segments) == 0 || len(src) == 0 || len(segments) != len(src) {
+		return src, nil
+	}
+	for segID, seg := range segments {
+		var docID uint64
+		for ; docID < seg.Count(); docID++ {
+			hasDeleted := false
+			var sourceData []byte
+			_ = seg.VisitStoredFields(docID, func(field string, value []byte) bool {
```
hanahmily: I want to recommend that you handle the error.

banyand/property/shard.go
```diff
@@ -225,3 +232,45 @@ func (s *shard) search(ctx context.Context, indexQuery index.Query, limit int,
...
+				if field == deleteField {
+					hasDeleted = convert.BytesToBool(value)
```
hanahmily: Please include the deletion time for this field. By doing so, there’s no need to parse the source field.

banyand/property/shard.go
```diff
@@ -225,3 +232,45 @@ func (s *shard) search(ctx context.Context, indexQuery index.Query, limit int,
...
+				continue
+			}
+
+			// parsing the property data to check the data is expired or not
+			var p propertyv1.Property
+			if err := protojson.Unmarshal(sourceData, &p); err != nil {
+				s.l.Warn().Msgf("unmarshal property failure when merging segments %d: %v", segID, err)
+				continue
+			}
+			if time.Now().Unix()-p.Metadata.ModRevision < s.expireToDeleteSec {
```
hanahmily: The `p.Metadata.ModRevision` represents the time in nanoseconds

banyand/property/shard_test.go
```diff
@@ -0,0 +1,113 @@
...
+		properties = append(properties, property)
+	}
+
+	// apply the property
+	for _, p := range properties {
+		if err = newShard.update(GetPropertyID(p), p); err != nil {
+			t.Fatal(err)
+		}
+	}
+
```
hanahmily, line 97: Add a step to verify the update by querying all properties. The returned value should match the inserted volume.

banyand/property/shard_test.go
```diff
@@ -0,0 +1,113 @@
...
+	commonv1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/common/v1"
+	modelv1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/model/v1"
+	propertyv1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/property/v1"
+	"github.com/apache/skywalking-banyandb/banyand/observability"
+	"github.com/apache/skywalking-banyandb/pkg/fs"
+	"github.com/apache/skywalking-banyandb/pkg/test"
+)
+
+func TestMergeDeleted(t *testing.T) {
```
hanahmily, line 34: Could you please add a new case to verify the properties that have been deleted but are still unexpired?

hanahmily (changes requested):

banyand/property/service.go
```diff
@@ -71,6 +72,7 @@ func (s *service) FlagSet() *run.FlagSet {
 	flagS.DurationVar(&s.flushTimeout, "property-flush-timeout", defaultFlushTimeout, "the memory data timeout of measure")
 	flagS.IntVar(&s.maxDiskUsagePercent, "property-max-disk-usage-percent", 95, "the maximum disk usage percentage allowed")
 	flagS.IntVar(&s.maxFileSnapshotNum, "property-max-file-snapshot-num", 2, "the maximum number of file snapshots allowed")
+	flagS.DurationVar(&s.expireTimeout, "property-expire-delete-timeout", time.Hour*24*7, "the duration of the expired data needs to be delete")
```
hanahmily:
```suggestion
	flagS.DurationVar(&s.expireTimeout, "property-expire-delete-timeout", time.Hour*24*7, "the duration of the expired data needs to be deleted")
```

hanahmily (changes requested):

hanahmily (approved):

PR #686, opened 2025-06-20
Support delete properties when query

mrproliu:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13281.


hanahmily (changes requested):
You should insert the latest value to the node containing expired data. That is how the fixing work.

mrproliu:
> You should insert the latest value to the node containing expired data. That is how the fixing work.

You mean test? This feature is only works on the query phase, why I need to insert?

wu-sheng:
> > You should insert the latest value to the node containing expired data. That is how the fixing work.
> 
> You mean test? This feature is only works on the query phase, why I need to insert?

I think he talked about how to prepare the data for verifying the fixing.

banyand/liaison/grpc/property.go
```diff
@@ -357,34 +357,76 @@ func (ps *propertyServer) Query(ctx context.Context, req *propertyv1.QueryReques
...
+				shouldDeleteOlderProperties = append(shouldDeleteOlderProperties, propertypkg.GetPropertyID(cur.Property))
+			case cur.Metadata.ModRevision == propertyMetadata.Metadata.ModRevision: // same revision
+				cur.AddExistNode(n)
+			default: // cur.Metadata.ModRevision > propertyMetadata.Metadata.ModRevision
+				shouldDeleteOlderProperties = append(shouldDeleteOlderProperties, propertypkg.GetPropertyID(propertyMetadata.Property))
 			}
 		}
 	}
+	if len(shouldDeleteOlderProperties) > 0 {
+		if err := ps.remove(shouldDeleteOlderProperties); err != nil {
```
hanahmily: To enhance performance, implement a repair operation in the data nodes. This operation receives the latest data, deletes all expired data, and then applies the latest one.

banyand/liaison/grpc/property.go
```diff
@@ -357,34 +357,76 @@ func (ps *propertyServer) Query(ctx context.Context, req *propertyv1.QueryReques
...
+		if !ok {
+			return nil, errors.New("failed to get group copies")
+		}
+		if copies != uint32(len(p.existNodes)) {
+			id, err := partition.ShardID(convert.StringToBytes(entity), group.ResourceOpts.ShardNum)
+			if err != nil {
+				return nil, err
+			}
+			nodes := make([]string, 0, copies)
+			for i := range copies {
```
hanahmily: You should repair the node with expired data, not all nodes. Your current solution will introduce more traffic.

hanahmily (changes requested):

banyand/liaison/grpc/property.go
```diff
@@ -405,11 +411,73 @@ func (ps *propertyServer) Query(ctx context.Context, req *propertyv1.QueryReques
...
+		return nil
+	}
+	id, err := partition.ShardID(convert.StringToBytes(entity), group.ResourceOpts.ShardNum)
+	if err != nil {
+		return err
+	}
+	// building the repair data
+	repairReq := &propertyv1.InternalRepairRequest{ShardId: uint64(id)}
+	if p.deletedTime > 0 { // building the delete request
+		repairReq.Operation = &propertyv1.InternalRepairRequest_Delete{
```
hanahmily: If the deletion's version is less than the latest version on the data node, the deletion should be dropped. I recommend using a unique repair request that contains a "deleted" field.

banyand/liaison/grpc/property.go
```diff
@@ -405,11 +411,73 @@ func (ps *propertyServer) Query(ctx context.Context, req *propertyv1.QueryReques
...
+	}
+	if copies == uint32(len(p.existNodes)) {
+		return nil
+	}
+	id, err := partition.ShardID(convert.StringToBytes(entity), group.ResourceOpts.ShardNum)
+	if err != nil {
+		return err
+	}
+	// building the repair data
+	repairReq := &propertyv1.InternalRepairRequest{ShardId: uint64(id)}
```
hanahmily: repairReq would be initialized when there are nodes that need to be repaired.

banyand/liaison/grpc/property.go
```diff
@@ -504,3 +575,23 @@ type propertyWithMetadata struct {
...
+
+func newPropertyWithCounts(p *propertyWithMetadata, existNode string) *propertyWithCount {
+	res := &propertyWithCount{
+		propertyWithMetadata: p,
+	}
+	res.AddExistNode(existNode)
+	return res
+}
+
+func (p *propertyWithCount) AddExistNode(node string) {
```
hanahmily:
```suggestion
func (p *propertyWithCount) addExistNode(node string) {
```

banyand/liaison/grpc/property.go
```diff
@@ -405,11 +411,73 @@ func (ps *propertyServer) Query(ctx context.Context, req *propertyv1.QueryReques
...
+		return errors.Errorf("group %s not found", p.Metadata.Group)
+	}
+	copies, ok := ps.groupRepo.copies(p.Metadata.GetGroup())
+	if !ok {
+		return errors.New("failed to get group copies")
+	}
+	if copies == uint32(len(p.existNodes)) {
+		return nil
+	}
+	id, err := partition.ShardID(convert.StringToBytes(entity), group.ResourceOpts.ShardNum)
```
hanahmily:
```suggestion
	shardID, err := partition.ShardID(convert.StringToBytes(entity), group.ResourceOpts.ShardNum)
```

banyand/property/db.go
```diff
@@ -219,6 +221,77 @@ func (db *database) collect() {
 	}
 }
 
+func (db *database) repairFromApplyProperty(ctx context.Context, id uint64, property *propertyv1.InternalApplyProperty) error {
```
hanahmily:
```suggestion
func (db *database) repairFromApplyProperty(ctx context.Context, shardID uint64, property *propertyv1.InternalApplyProperty) error {
```

banyand/property/db.go
```diff
@@ -219,6 +221,77 @@ func (db *database) collect() {
...
+	if err != nil {
+		return fmt.Errorf("query older properties failed: %w", err)
+	}
+	docIDList := make([][]byte, 0, len(olderProperties))
+	for _, p := range olderProperties {
+		if p.deleteTime > 0 {
+			// If the property is already deleted, ignore it.
+			continue
+		}
+		docIDList = append(docIDList, p.id)
```
hanahmily: If the property's version is older than the latest version in olderProperties, the repair should be stopped.

hanahmily: Then, keep the latest property ( whether it has been deleted or not), and update all expired properties to be deleted.

hanahmily (changes requested):

banyand/property/db.go
mrproliu: You mean the liaison node may request an older version to repair?

banyand/liaison/grpc/property.go
mrproliu: Yeah. I have checked the latest version replicas count is the same as the group replicas count: https://github.com/apache/skywalking-banyandb/pull/686/files#diff-ec9aafdd27b3b36de2fae1ed134371dd730232bbd344d77197eebd1b0927221eR424
If they are not the same, it will only send the repair request to the non-sync nodes. All the nodes are using the same request.

banyand/liaison/grpc/property.go
mrproliu: The property in the liaison node will only use the latest version: https://github.com/apache/skywalking-banyandb/pull/686/files#diff-ec9aafdd27b3b36de2fae1ed134371dd730232bbd344d77197eebd1b0927221eR375
So, if the deletion's version is less than the latest version, it will still be an applied request, not a deleted request. When the applied request is sent to the data node, the data will be deleted and all older versions, and apply the new one.

hanahmily: Other users may trigger additional writing during this process, making it impossible to guarantee that the data remains consistent after the query.

It's important to remember that all data can change over time, which is why distributed operations can be challenging.

banyand/liaison/grpc/property.go
hanahmily: If all nodes are the same, no repair is required. The repairReq should not be constructed.

banyand/property/db.go
hanahmily: Absolutely. We accept concurrent operations.

banyand/liaison/grpc/property.go
mrproliu: Yes, that's true. If the latest version property is already in all nodes, the repair request will not be constructed. 
I have already checked in there: https://github.com/apache/skywalking-banyandb/pull/686/files#diff-ec9aafdd27b3b36de2fae1ed134371dd730232bbd344d77197eebd1b0927221eR424

banyand/property/shard.go
```diff
@@ -244,13 +248,102 @@ func (s *shard) search(ctx context.Context, indexQuery index.Query, limit int,
...
+		if err != nil {
+			return fmt.Errorf("build update document failed: %w", err)
+		}
+		doc.DeletedTime = deleteTime
+		return s.updateDocuments(index.Documents{*doc})
+	}
+
+	// if the lastest property in shard is bigger(or equals) than the repaired property,
+	// then the repaired process should be stopped.
+	// only deleted the older properties(exclude lastest one).
```
hanahmily:
```suggestion
```

banyand/property/shard.go
```diff
@@ -244,13 +248,102 @@ func (s *shard) search(ctx context.Context, indexQuery index.Query, limit int,
...
+		if len(olderProperties) > 1 {
+			docIDList := s.buildNotDeletedDocIDList(olderProperties[0 : len(olderProperties)-1])
+			var deletedDocuments []index.Document
+			deletedDocuments, err = s.buildDeleteFromTimeDocuments(ctx, docIDList, time.Now().UnixNano())
+			if err != nil {
+				return fmt.Errorf("build delete documents failed: %w", err)
+			}
+			documents = append(documents, deletedDocuments...)
+		}
+		return s.updateDocuments(documents)
```
hanahmily:
```suggestion
		return nil
```

In this branch, the data node has a newer version than liaison. I recommand that you stop trying to fix it in this round because you can not repair based on the expired property.

banyand/property/shard.go
```diff
@@ -233,13 +247,103 @@ func (s *shard) search(ctx context.Context, indexQuery index.Query, limit int,
...
+	if err != nil {
+		return fmt.Errorf("build delete older documents failed: %w", err)
+	}
+	// update the property to mark it as delete
+	updateDoc, err := s.buildUpdateDocument(id, property)
+	if err != nil {
+		return fmt.Errorf("build repair document failure: %w", err)
+	}
+	// set the delete time(could be zero) to the latest delete time
+	updateDoc.DeletedTime = deleteTime
```
hanahmily: Please remove the deletedTime from the Document. All fields related to the property should be defined in the property package, like "_source". I'm sorry for not bringing this issue to your attention in the previous PR.

hanahmily (changes requested):

banyand/liaison/grpc/property.go
```diff
@@ -405,11 +411,62 @@ func (ps *propertyServer) Query(ctx context.Context, req *propertyv1.QueryReques
 	return &propertyv1.QueryResponse{Properties: properties, Trace: trace}, nil
 }
 
+func (ps *propertyServer) repairPropertyIfNeed(ctx context.Context, entity string, p *propertyWithCount, groups map[string]*commonv1.Group) error {
```
hanahmily: The `repairPropertyIfNeed` function should consider limiting the number of concurrent repair tasks to avoid overwhelming the data nodes with excessive repair traffic. I suggest introducing a limit-size async-queue for repair tasks. 

When adding a new repair task to the queue, please also check for duplicates by comparing the entity and ModRevision to prevent redundant repairs for the same property version. 

This would help ensure the system remains stable under heavy load and avoids unnecessary duplicate repair operations.

hanahmily (changes requested):

banyand/liaison/grpc/property.go
```diff
@@ -405,11 +414,32 @@ func (ps *propertyServer) Query(ctx context.Context, req *propertyv1.QueryReques
...
+		return errors.New("failed to get group copies")
+	}
+	if copies == uint32(len(p.existNodes)) {
+		return nil
+	}
+	shardID, err := partition.ShardID(convert.StringToBytes(entity), group.ResourceOpts.ShardNum)
+	if err != nil {
+		return err
+	}
+	ps.repairQueue.addProperty(entity, shardID, p)
```
hanahmily: You should obtain the repair target (the node that needs to be repaired) before sending the queue. In other words, if no node needs to be repaired, nothing should be sent to the queue.

banyand/liaison/grpc/property.go
```diff
@@ -500,7 +533,152 @@ func (ps *propertyServer) remove(ids [][]byte) error {
...
+		entity:   entity,
+		property: p,
+	}
+	// adding the task to the queue, if the queue is full, then ignore it
+	select {
+	case rq.queue <- task:
+	default:
+		return false
+	}
+	return true
```
hanahmily:
```suggestion
	task := &repairTask{
		shardID:  shardID,
		entity:   entity,
		property: p,
	}
	// adding the task to the queue, if the queue is full, then ignore it
	select {
	case rq.queue <- task:
	        rq.inProcess[key] = struct{}{}
	default:
		return false
	}
	return true
```

banyand/liaison/grpc/property.go
```diff
@@ -500,7 +533,152 @@ func (ps *propertyServer) remove(ids [][]byte) error {
...
+}
+
+func (rq *repairQueue) Start(stop chan struct{}) {
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		<-stop
+		cancel()
+	}()
```
hanahmily: No need to use this technology. The pipeline.publish has an internal timeout mechanism that is not affected by the context. You can simply use the stop channel to exit the goroutine of the task execution loop.

banyand/liaison/grpc/property.go
```diff
@@ -500,7 +533,152 @@ func (ps *propertyServer) remove(ids [][]byte) error {
...
+			}
+		}
+	}()
+}
+
+func (rq *repairQueue) processTask(ctx context.Context, t *repairTask) error {
+	defer func() {
+		rq.processLocker.Lock()
+		defer rq.processLocker.Unlock()
+		key := repairInProcessKey{
```
hanahmily: Would you put the key in the repairTask?

banyand/liaison/grpc/property.go
```diff
@@ -500,7 +533,152 @@ func (ps *propertyServer) remove(ids [][]byte) error {
...
+}
+
+func (p *propertyWithCount) addExistNode(node string) {
+	if p.existNodes == nil {
+		p.existNodes = make(map[string]bool)
+	}
+	p.existNodes[node] = true
+}
+
+type repairQueue struct {
```
hanahmily, line 580: You should add some metrics of this queue to help us know the status of the repair work.

hanahmily (changes requested):

PR #688, opened 2025-06-22
Encode each column based on its data type

OmCheeLin:
### Encode each column based on its data type

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes parts of apache/skywalking#13254.


wu-sheng:
Why open a new pull request?

OmCheeLin:
All concurrency issues have been resolved. However, please note that the 
`valueType` of `columnMetadata` and `column`,
`valueType` of `tagMetadata` and `tag`,
may be different, and only if it is nil or string(null).

See `TODO` part.

OmCheeLin:
> Why open a new pull request?

The previous PR was closed, and I ran CI with the same PR in my private repo to avoid any impact

banyand/measure/column.go
```diff
@@ -18,13 +18,33 @@
...
 	"github.com/apache/skywalking-banyandb/pkg/encoding"
 	"github.com/apache/skywalking-banyandb/pkg/fs"
 	"github.com/apache/skywalking-banyandb/pkg/logger"
 	pbv1 "github.com/apache/skywalking-banyandb/pkg/pb/v1"
 )
 
+var (
+	int64SlicePool = sync.Pool{
+		New: func() interface{} {
+			s := make([]int64, 0, 1024)
```
hanahmily: You should avoid pre-allocating the size of the slice. Instead, resize the slice as needed.

banyand/measure/column.go
```diff
@@ -60,11 +80,88 @@ func (c *column) mustWriteTo(cm *columnMetadata, columnWriter *writer) {
...
-	// marshal values
-	bb.Buf = encoding.EncodeBytesBlock(bb.Buf[:0], c.values)
+	encodeDefault := func() {
+		bb.Buf = encoding.EncodeBytesBlock(bb.Buf[:0], c.values)
+	}
+
+	// select encoding based on data type
+encodeSwitch:
+	switch c.valueType {
+	case pbv1.ValueTypeInt64:
```
hanahmily, line 110: Encapsulate the encoding logic in a function. We will have several encoders later on.

banyand/measure/column.go
```diff
@@ -18,13 +18,33 @@
...
 	"github.com/apache/skywalking-banyandb/pkg/bytes"
+	"github.com/apache/skywalking-banyandb/pkg/convert"
 	"github.com/apache/skywalking-banyandb/pkg/encoding"
 	"github.com/apache/skywalking-banyandb/pkg/fs"
 	"github.com/apache/skywalking-banyandb/pkg/logger"
 	pbv1 "github.com/apache/skywalking-banyandb/pkg/pb/v1"
 )
 
+var (
+	int64SlicePool = sync.Pool{
```
hanahmily: You should utilize the pool in our library to monitor the pool leak in the test case.

banyand/measure/column.go
```diff
@@ -60,11 +80,88 @@ func (c *column) mustWriteTo(cm *columnMetadata, columnWriter *writer) {
...
+		} else {
+			intValues = intValues[:len(c.values)]
+		}
+		defer func() {
+			*intValuesPtr = intValues[:0]
+			int64SlicePool.Put(intValuesPtr)
+		}()
+
+		for i, v := range c.values {
+			if len(v) != 8 {
```
hanahmily: If the value is malformed, use logger.Panic to notify this bug.

banyand/measure/column.go
```diff
@@ -60,11 +80,88 @@ func (c *column) mustWriteTo(cm *columnMetadata, columnWriter *writer) {
...
+			*floatValuesPtr = floatValues[:0]
+			float64SlicePool.Put(floatValuesPtr)
+		}()
+
+		for i, v := range c.values {
+			floatValues[i] = convert.BytesToFloat64(v)
+		}
+		// use XOR encoding for float column
+		bb.Buf = bb.Buf[:0]
+		writer := encoding.NewWriter()
```
hanahmily: To prevent unnecessary allocation, please pool this writer.

banyand/measure/column_metadata.go
```diff
@@ -28,24 +28,63 @@ import (
...
 type columnMetadata struct {
 	name string
+	encodeBlock
 	dataBlock
 	valueType pbv1.ValueType
 }
 
+type encodeBlock struct {
+	encodeType encoding.EncodeType
+	firstValue int64
```
hanahmily: The first value should be at the column instead of the meta. It can not help us search the block more than a overhead.

hanahmily (changes requested):

hanahmily:
@OmCheeLin According to the benchmark, you should convert float64 to uint64 and then use integer encoding.

OmCheeLin:
| Pattern                           | Type           | ns/op    | Compression Ratio | Encode Time (ns/op) | Decode Time (ns/op) | Encode+Decode (ns/op) |
|-----------------------------------|----------------|----------|-------------------|----------------------|----------------------|------------------------|
| const_old_str_10000              | old_str        | 303893   | 0.0004375         | 151835               | 333274               | 485110                 |
| const_new_float64_10000          | new_float64    | 336210   | 0.0001625         | 139007               | 339532               | 478540                 |
| incrementing_old_str_10000       | old_str        | 292817   | 0.4868            | 161775               | 295961               | 457737                 |
| incrementing_new_float64_10000   | new_float64    | 353505   | 0.0001750         | 160999               | 331875               | 492874                 |
| small_fluctuations_old_str_10000 | old_str        | 1541257  | 1.449             | 1240453              | 499237               | 1739690                |
| small_fluctuations_new_float64_10000 | new_float64 | 977772   | 0.8722            | 652133               | 481447               | 1133580                |
| random_old_str_10000             | old_str        | 459497   | 2.180             | 328865               | 285355               | 614220                 |
| random_new_float64_10000         | new_float64    | 1053347  | 1.050             | 710005               | 508315               | 1218320                |

This is float64 (traditional encoding vs decimal encoding).

for this 
```go
if len(v) != 8 {
	var val int64
	for j := 0; j < len(v); j++ {
		val = (val << 8) | int64(v[j])
	}
	intValues[i] = val
} else {
	intValues[i] = convert.BytesToInt64(v)
}
```
I have tried all methods in `encoding`, but convert failed, I might not be able to resolve it. I think it’s best if you take a look after the code is merged and then address it in the next PR.

And in `float.go`, I do not use `shopspring/decimal` to ensure float64 precision, as it increases the encoding and decoding time by 10 times, which might be unacceptable.

hanahmily:
> I have tried all methods in encoding, but convert failed, I might not be able to resolve it. I think it’s best if you take a look after the code is merged and then address it in the next PR.

The encoding and decoding functions are both sourced from the `convert` package. If this is accurate, our test cases should fail on the first day of execution. Can you write a simple test case to verify your code, or identify which existing case might lead to this issue?

banyand/measure/column.go
```diff
@@ -60,16 +105,103 @@ func (c *column) mustWriteTo(cm *columnMetadata, columnWriter *writer) {
 	bb := bigValuePool.Generate()
 	defer bigValuePool.Release(bb)
 
-	// marshal values
-	bb.Buf = encoding.EncodeBytesBlock(bb.Buf[:0], c.values)
+	c.encodeColumnValues(cm, bb)
```
hanahmily: No I don't want to move "switch" to a function. What I want is 

```suggestion
	switch c.valueType {
	         case pbv1.ValueTypeInt64:
	                  c.encodeInt64()
	         case pbv1.ValueTypeFloat64:
	                 c.encodeFloat64()
	          default:
	                 bb.Buf = encoding.EncodeBytesBlock(bb.Buf[:0], c.values)
	}
```

OmCheeLin:
@hanahmily sorry, I made a mistake in `banyand/stream/part_test.go`, please run ci again

.github/workflows/dependency-review.yml
```diff
@@ -26,6 +26,7 @@ permissions:
 
 jobs:
   dependency-review:
+    if: github.repository == 'apache/skywalking-banyandb'
```
hanahmily:
```suggestion
```

pkg/encoding/xor_test.go
```diff
@@ -50,6 +52,30 @@ func TestXOR(t *testing.T) {
 	verify(d, a, uint64(100))
 }
 
+func TestXOR1(t *testing.T) {
```
hanahmily: Why did you add a copy of TestXOR？

pkg/encoding/float.go
```diff
@@ -0,0 +1,85 @@
...
+		dst = append(dst, float64(v)/scale)
+	}
+	return dst, nil
+}
+
+// countDecimalPlaces estimates the number of decimal places in a float64 up to a given maximum.
+func countDecimalPlaces(f float64, maxPlace int) int32 {
+	for i := 0; i < maxPlace; i++ {
+		f *= 10
+		if math.Abs(f-math.Round(f)) < 1e-9 {
```
hanahmily, line 84: Do you tend to get the fractional part? If so, how about using `math.Modf` to extract the fractional part? What is the reason for using `1e-9`? Could you provide some boundary examples or test cases?

banyand/measure/column.go
```diff
@@ -60,16 +105,98 @@ func (c *column) mustWriteTo(cm *columnMetadata, columnWriter *writer) {
...
+		}()
+
+		for i, v := range c.values {
+			floatValues[i] = convert.BytesToFloat64(v)
+		}
+		intValues, exp, err := encoding.Float64ListToDecimalIntList(intValues[:0], floatValues)
+		if err != nil {
+			if exp == -1 {
+				// TODO c.valueType = pbv1.ValueTypeStr
+				cm.valueType = pbv1.ValueTypeStr
```
hanahmily: You cannot change the value type, as this will cause a decode failure. How about adding a new encode type instead?

banyand/measure/column.go
```diff
@@ -60,16 +105,98 @@ func (c *column) mustWriteTo(cm *columnMetadata, columnWriter *writer) {
...
+				encodeDefault()
+				return
+			}
+			logger.Panicf("cannot convert Float64List to DecimalIntList : %v", err)
```
hanahmily: You should revert to string encoding for a Nan or Inf float instead of panicking on errors.

pkg/encoding/float.go
```diff
@@ -0,0 +1,85 @@
...
+
+import (
+	"fmt"
+	"math"
+
+	"github.com/apache/skywalking-banyandb/pkg/logger"
+)
+
+// Float64ListToDecimalIntList converts float64 values to int64s with a common decimal scale factor.
+func Float64ListToDecimalIntList(dst []int64, nums []float64) ([]int64, int32, error) {
```
hanahmily:
```suggestion
func Float64ListToDecimalIntList(dst []int64, nums []float64) ([]int64, int16, error) {
```

.github/workflows/dependency-review.yml
```diff
@@ -26,7 +26,6 @@ permissions:
 
 jobs:
   dependency-review:
-    if: github.repository == 'apache/skywalking-banyandb'
```
wu-sheng: Why do you remove this?

banyand/measure/column.go
```diff
@@ -60,16 +105,112 @@ func (c *column) mustWriteTo(cm *columnMetadata, columnWriter *writer) {
...
+		}
+		floatValues[i] = convert.BytesToFloat64(v)
+	}
+	intValues, exp, err := encoding.Float64ListToDecimalIntList(intValues[:0], floatValues)
+	if err != nil {
+		if exp == -1 {
+			doEncodeDefault()
+			return
+		}
+		logger.Panicf("cannot convert Float64List to DecimalIntList : %v", err)
```
hanahmily:
```suggestion
		logger.Errorf("cannot convert Float64List to DecimalIntList: %w", err)
                doEncodeDefault()
                return
```

banyand/stream/tag.go
```diff
@@ -85,6 +137,95 @@ func (t *tag) mustWriteTo(tm *tagMetadata, tagWriter *writer, tagFilterWriter *w
...
+		}
+		floatValues[i] = convert.BytesToFloat64(v)
+	}
+	intValues, exp, err := encoding.Float64ListToDecimalIntList(intValues[:0], floatValues)
+	if err != nil {
+		if exp == -1 {
+			doEncodeDefault()
+			return
+		}
+		logger.Panicf("cannot convert Float64List to DecimalIntList : %v", err)
```
hanahmily:
```suggestion
		logger.Errorf("cannot convert Float64List to DecimalIntList: %w", err)
                doEncodeDefault()
                return
```

OmCheeLin:
@hanahmily If modified this way, it will no longer be able to handle float overflow scenarios.

OmCheeLin:
@hanahmily The original CI tests were able to pass.

OmCheeLin:
If err is not nil and exp == -1, it indicates an overflow scenario.

hanahmily:
> If err is not nil and exp == -1, it indicates an overflow scenario.

You should disregard any errors. There is no need to panic. Before the float encoding is introduced, we remain calm. After the PR is merged, I hope this behavior does not change.

OmCheeLin:
> > If err is not nil and exp == -1, it indicates an overflow scenario.
> 
> You should disregard any errors. There is no need to panic. Before the float encoding is introduced, we remain calm. After the PR is merged, I hope this behavior does not change.

But you may should use %v instead of %w？This will cause err

github.com/apache/skywalking-banyandb/banyand/measure
# [github.com/apache/skywalking-banyandb/banyand/measure]
Error: ./column.go:190:3: github.com/apache/skywalking-banyandb/pkg/logger.Errorf does not support error-wrapping directive %w

Failed to compile stream:

# github.com/apache/skywalking-banyandb/banyand/stream
# [github.com/apache/skywalking-banyandb/banyand/stream]
Error: ./tag.go:205:3: github.com/apache/skywalking-banyandb/pkg/logger.Errorf does not support error-wrapping directive %w

composite coverage: 34.1% of statements

Ginkgo ran 52 suites in 2m20.039938161s

There were failures detected in the following suites:
  measure ./banyand/measure [Compilation failure]
   stream ./banyand/stream [Compilation failure]

Test Suite Failed
make: *** [Makefile:75: test-ci] Error 1
Error: Process completed with exit code 2.

wu-sheng:
Please use current markdown format to quote the logs.

hanahmily (approved):

PR #691, opened 2025-07-01
Introduce the `Trace` data model to store the trace/span data

hanahmily:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13328.

### DIFF
--- a/api/proto/banyandb/database/v1/schema.proto
+++ b/api/proto/banyandb/database/v1/schema.proto
@@ -163,6 +163,8 @@ message IndexRule {
   enum Type {
     TYPE_UNSPECIFIED = 0;
     TYPE_INVERTED = 1;
+    // TYPE_TREE is a tree index, which is used for storing hierarchical data.
+    TYPE_TREE = 3;
   }
   // type is the IndexType of this IndexObject.
   Type type = 3 [(validate.rules).enum.defined_only = true];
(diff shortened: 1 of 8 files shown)


api/proto/banyandb/database/v1/schema.proto
```diff
@@ -163,6 +163,8 @@ message IndexRule {
   enum Type {
     TYPE_UNSPECIFIED = 0;
     TYPE_INVERTED = 1;
+    // TYPE_TREE is a tree index, which is used for storing hierarchical data.
+    TYPE_TREE = 3;
```
hanahmily, line 168: `2` is for the skipping index.

wu-sheng: From this one? https://github.com/apache/skywalking-banyandb/pull/665

api/proto/banyandb/trace/v1/query.proto
```diff
@@ -0,0 +1,72 @@
...
+message QueryResponse {
+  // spans is a list of spans that match the query.
+  repeated Span spans = 1;
+  // trace_query_result contains the trace of the query execution if tracing is enabled.
+  common.v1.Trace trace_query_result = 2;
+}
+
+// QueryRequest is the request contract for query.
+message QueryRequest {
+  // groups indicate the physical data location.
```
wu-sheng:
```suggestion
  // groups indicates the physical data location.
```

By following your comment grammar style, you treat a field as a single rather than a plural.

api/proto/banyandb/trace/v1/query.proto
```diff
@@ -0,0 +1,72 @@
...
+  string name = 2 [(validate.rules).string.min_len = 1];
+  // time_range is a range query with begin/end time of entities in the timeunit of milliseconds.
+  // In the context of trace, it represents the range of the `startTime` for spans/segments,
+  // it is always recommended to specify time range for performance reason
+  model.v1.TimeRange time_range = 3;
+  // offset is used to support pagination, together with the following limit
+  uint32 offset = 4;
+  // limit is used to impose a boundary on the number of spans being returned
+  uint32 limit = 5;
+  // order_by is given to specify the sort for a tag. So far, only fields in the type of Integer are supported
```
wu-sheng: Only fields? Could you clarify this? I think we only have tags for trace?

api/proto/banyandb/database/v1/schema.proto
hanahmily: Exactly

wu-sheng: fyi @wankai123 Order_by latency and start time will be treated as a composite index in one index rule. Something like an index rule has an ordered tags list, e.g. a list of latency, service_name, etc.

PR #692, opened 2025-07-01
Support save repair state tree in property

mrproliu:
Before checking data consistency verification for the property, it needs to generate the repair state file in each property shard. 

There are two fields in each repair state file:
1. Latest snapshot ID: The latest snapshot ID, if there are no snapshot updates, means this shard has no updates. 
2. Trees: The Merkle tree of each property group. It can help compare data between nodes. 

Also,  add `_sha_value` field in each property for preparing to build the Merkle tree.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Part of apache/skywalking#13284.


banyand/property/repair.go
```diff
@@ -0,0 +1,295 @@
...
+		var property propertyv1.Property
+		err = protojson.Unmarshal(source, &property)
+		if err != nil {
+			continue
+		}
+		entity := GetEntity(&property)
+		if shaValue == "" {
+			shaValue, err = r.buildingShaValue(source, deleteTime)
+			if err != nil {
+				continue
```
hanahmily: Consider logging this for debugging

banyand/property/repair.go
```diff
@@ -0,0 +1,295 @@
...
+	// save the Trees to the state
+	state = &repairStatus{
+		LastSnapshotID: latestSnapshotID,
+		Trees:          tree,
+	}
+	stateVal, err := json.Marshal(state)
+	if err != nil {
+		return nil, fmt.Errorf("marshall state failure: %w", err)
+	}
+	err = os.WriteFile(r.statePath, stateVal, 0o600)
```
hanahmily: You should use `storage.FilePerm`

banyand/property/repair.go
```diff
@@ -0,0 +1,295 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("writing state file failure: %w", err)
+	}
+	return state, nil
+}
+
+func (r *repair) buildingTree(conf bluge.Config, existingSnapshots []uint64) (map[string]*repairTree, error) {
```
hanahmily:
```suggestion
func (r *repair) buildTree(conf bluge.Config) (map[string]*repairTree, error) {
```

banyand/property/repair.go
```diff
@@ -0,0 +1,295 @@
...
+		tree, err := treeBuilder.build()
+		if err != nil {
+			return nil, fmt.Errorf("building tree for group %s failure: %w", group, err)
+		}
+		result[group] = tree
+	}
+	return result, nil
+}
+
+func (r *repair) buildingShaValue(source []byte, deleteTime int64) (string, error) {
```
hanahmily:
```suggestion
func (r *repair) buildShaValue(source []byte, deleteTime int64) (string, error) {
```

banyand/property/repair.go
hanahmily: The variable `existingSnapshots` is not utilized in a practical scenario.

banyand/property/repair.go
```diff
@@ -0,0 +1,295 @@
...
+	reader, err := bluge.OpenReader(conf)
+	if err != nil {
+		return nil, fmt.Errorf("opening index reader failure: %w", err)
+	}
+	totalCount, err := reader.Count()
+	if err != nil {
+		return nil, fmt.Errorf("counting documents failure: %w", err)
+	}
+	treeBuilders := make(map[string]*repairTreeBuilder)
+	for itemID := uint64(0); itemID < totalCount; itemID++ {
```
hanahmily: To optimize memory usage, sort the results by `group`, `name`, `entity`, and `timestamp` using the search API. The last entry for the same `group`, `name`, and `entity` should have the largest `timestamp`, indicating the most recent property.

After all properties in a group have been gathered, compute the hashed value for that group. Then, reset the entire builder to prepare for handling the next group. This means you only need to allocate space for a single group at a time.

hanahmily (changes requested):

banyand/property/repair.go
```diff
@@ -0,0 +1,467 @@
...
+	return nil
+}
+
+func (r *repairTreeComposer) writeString(val string) error {
+	bytes := []byte(val)
+	if err := binary.Write(r.file, binary.LittleEndian, int32(len(bytes))); err != nil {
+		return fmt.Errorf("writing string %s to repair tree file %s failure: %w", val, r.path, err)
+	}
+	_, err := r.file.Write(bytes)
+	return err
```
hanahmily: Please use `bufio.Writer`

banyand/property/repair.go
```diff
@@ -0,0 +1,467 @@
...
+	}
+	r.file, err = os.OpenFile(r.path, os.O_CREATE|os.O_RDWR, storage.FilePerm)
+	if err != nil {
+		return nil, fmt.Errorf("opening repair tree file %s failure: %w", r.path, err)
+	}
+	return
+}
+
+func (r *repairTreeComposer) append(group, id, shaVal string) (err error) {
+	if err = r.writeString(group); err != nil {
```
hanahmily: You can use `encoding.EncodeBytes` to encode string.

banyand/property/repair.go
```diff
@@ -0,0 +1,467 @@
...
+	if latestProperty != nil {
+		if err = treeComposer.append(latestProperty.group, latestProperty.entityID, latestProperty.shaValue); err != nil {
+			return nil, fmt.Errorf("appending latest property to tree composer failure: %w", err)
+		}
+	}
+
+	result := make(map[string]*repairTree)
+	trees, err := treeComposer.compose()
+	for group, treeBuilder := range trees {
+		tree, err := treeBuilder.build()
```
hanahmily: You should build the group's tree in the pageSearch loop.

hanahmily (changes requested):

banyand/property/repair.go
mrproliu: If I use `encoding.EncodeBytes`, It needs one more `int32` space to save the length of the byte array.

hanahmily: `encoding.EncodeBytes` encodes the length in bytes automatically. You don't need to encode the length separately, which simplifies your code.

mrproliu: But I need do decode the specific length when reading?

banyand/property/repair.go
```diff
@@ -0,0 +1,654 @@
...
+	l *logger.Logger,
+	metricsFactory *observability.Factory,
+	batchSearchSize,
+	treeSlotCount int,
+	repairBuildTreeCron string,
+	repairQuickBuildTreeTime time.Duration,
+	takeSnapshot func(string) error,
+) (r *repair, err error) {
+	repairBase := path.Join(shardPath, "repair")
+	c := cron.New()
```
hanahmily: Please use the `timestamp.Scheduler`

banyand/property/repair.go
```diff
@@ -0,0 +1,654 @@
...
+	return r, nil
+}
+
+func (r *repair) documentUpdatesNotify() {
+	if !atomic.CompareAndSwapInt32(r.quickRepairNotified, 0, 1) {
+		return
+	}
+
+	// stop the normal cron job
+	r.repairTreeCron.Remove(r.repairTreeCronTask)
```
hanahmily: You should not stop the task. Instead, skip the cron task execution if the last execution time is less than half of the interval between executions. 

For example, if the first execution occurs at 10:00 and the interval is 1 hour, the next scheduled execution should be at 11:00. If a quick repair happens at 10:20, the task should execute. However, if the quick repair occurs at 10:40, the task should be skipped.

banyand/property/repair.go
```diff
@@ -0,0 +1,654 @@
...
+	"github.com/apache/skywalking-banyandb/pkg/convert"
+	"github.com/apache/skywalking-banyandb/pkg/logger"
+	"github.com/apache/skywalking-banyandb/pkg/meter"
+)
+
+const repairBatchSearchSize = 100
+
+type repair struct {
+	takeSnapshot             func(string) error
+	ctx                      context.Context
```
hanahmily: You should use `run.Closer` instead of a context.

banyand/property/repair.go
```diff
@@ -0,0 +1,654 @@
...
+			r.repairTreeCronTask, err = r.repairTreeCron.AddJob(r.repairBuildTreeCron, r)
+			if err != nil {
+				r.l.Err(fmt.Errorf("repair build tree cron task err: %w", err))
+			}
+			atomic.StoreInt32(r.quickRepairNotified, 0)
+		}
+	}()
+}
+
+func (r *repair) Run() {
```
hanahmily:
```suggestion
func (r *repair) run() {
```

hanahmily: A function should be exported if there is no reference from outside.

banyand/property/repair.go
```diff
@@ -0,0 +1,654 @@
...
+	_, err := r.buildStatus()
```
hanahmily: If repairStatus is not used, it should be removed.

banyand/property/repair.go
hanahmily: You can use the DecodeBytes from the same pkg

banyand/property/repair.go
```diff
@@ -0,0 +1,853 @@
...
+		r.metrics.totalBuildTreeDuration.Inc(time.Since(startTime).Seconds())
+	}()
+	var state *repairStatus
+	// reading the state file to check they have any updates
+	state, err = r.readState()
+	if err != nil {
+		return fmt.Errorf("reading state failure: %w", err)
+	}
+	indexConfig := index.DefaultConfig(r.shardPath)
+	items, err := indexConfig.DirectoryFunc().List(index.ItemKindSegment)
```
hanahmily:
```suggestion
	items, err := indexConfig.DirectoryFunc().List(index.ItemKindSnapshot)
```

There is only 1 *.snp file there.

banyand/property/repair.go
```diff
@@ -0,0 +1,853 @@
...
+	if state != nil && len(items) != 0 && items[len(items)-1] == state.LastSnpID {
+		return nil
+	}
+	if len(items) == 0 {
+		return nil
+	}
+
+	// otherwise, we need to building the trees
+	// take a new snapshot first
+	err = r.takeSnapshot(r.snapshotDir)
```
hanahmily:
```suggestion
	snapshotDir,err = r.takeSnapshot()
```

The snapshot directory has a dynamic path that includes a creation timestamp.

banyand/property/repair.go
```diff
@@ -0,0 +1,853 @@
...
+		fmt.Sprintf("+%s", entityID),
+		fmt.Sprintf("+%s", timestampField),
+	})
+
+	var latestProperty *searchingProperty
+	treeComposer := newRepairTreeComposer(r.composeSlotAppendFilePath, r.composeTreeFilePathFmt, r.treeSlotCount, r.l)
+	if err != nil {
+		return fmt.Errorf("creating repair tree composer failure: %w", err)
+	}
+	err = r.pageSearch(reader, topNSearch, func(source []byte, shaValue string, deleteTime int64) error {
```
hanahmily:
```suggestion
	err = r.pageSearch(reader, topNSearch, func(shaValue string, deleteTime int64) error {
```

You can get `group`, `name` and `entityID` from SortValue

banyand/property/repair.go
```diff
@@ -0,0 +1,853 @@
...
+		entity := GetEntity(&property)
+		if shaValue == "" {
+			shaValue, err = r.buildShaValue(source, deleteTime)
+			if err != nil {
+				return fmt.Errorf("building sha value failure: %w", err)
+			}
+		}
+
+		s := newSearchingProperty(&property, shaValue, entity)
+		if latestProperty != nil {
```
hanahmily, line 191: If there is only one property, you might lose it.

banyand/property/repair.go
```diff
@@ -0,0 +1,853 @@
...
+	for _, f := range r.slotFiles {
+		if f == nil {
+			continue
+		}
+		if err = treeBuilder.appendSlot(f); err != nil {
+			return fmt.Errorf("appending slot file %s to repair tree builder for group %s failure: %w", f.path, group, err)
+		}
+	}
+	// reset the slot files for the next group
+	r.slotFiles = make([]*repairSlotFile, r.slotCount)
```
hanahmily: You can set `r.slotFiles[i]=nil` to reset the slotFiles in the upper loop.

banyand/property/repair.go
```diff
@@ -0,0 +1,853 @@
...
+		leafDataBytes, entity, err = encoding.DecodeBytes(leafDataBytes)
+		if err != nil {
+			return nil, fmt.Errorf("decoding leaf node entity from file %s failure: %w", r.file.Name(), err)
+		}
+		leafDataBytes, shaVal, err = encoding.DecodeBytes(leafDataBytes)
+		if err != nil {
+			return nil, fmt.Errorf("decoding leaf node sha value from file %s failure: %w", r.file.Name(), err)
+		}
+		reduceBytesLen = int64(len(leafDataBytes))
+		nodes = append(nodes, &repairTreeNode{
```
hanahmily, line 501: You should use the streaming method to load leaves. Loading all data into memory at once will cause OOM errors. Providing a readLine method for retrieving leaves is a better solution.

banyand/property/repair.go
mrproliu: I have added a check after the page search finished: https://github.com/apache/skywalking-banyandb/pull/692/files#diff-aa9f3b360d7011ead46823bdacff82ddf3f4786ac0ac66e14cac2683246265cbR229-R237.
Also, I have a UT for this case: https://github.com/apache/skywalking-banyandb/pull/692/files#diff-f8090c2e2cf3484ed68becf46cc06103a87968a548c613d9a118bb8a5cf36630R68-R79

banyand/property/repair.go
mrproliu: I have added the sortValue in this function, but the source is still needed because the older property may not have the SHA value, so it needs to be re-generated.

banyand/property/repair.go
```diff
@@ -0,0 +1,947 @@
...
+	state, err := r.readState()
+	if err != nil {
+		return false, fmt.Errorf("reading state failure: %w", err)
+	}
+	indexConfig := index.DefaultConfig(r.shardPath)
+	items, err := indexConfig.DirectoryFunc().List(index.ItemKindSnapshot)
+	if err != nil {
+		return false, fmt.Errorf("reading item kind snapshot failure: %w", err)
+	}
+	sort.Sort(snapshotIDList(items))
```
hanahmily, line 110: There is only one file here; just select the first item.

banyand/property/repair.go
```diff
@@ -0,0 +1,947 @@
...
+			continue
+		}
+		if err = treeBuilder.appendSlot(f); err != nil {
+			return fmt.Errorf("appending slot file %s to repair tree builder for group %s failure: %w", f.path, group, err)
+		}
+	}
+	// reset the slot files for the next group
+	for i := range r.slotFiles {
+		r.slotFiles[i] = nil
+	}
```
hanahmily:
```suggestion
			return fmt.Errorf("appending slot file %s to repair tree builder for group %s failure: %w", f.path, group, err)
		}
		r.slotFiles[i] = nil
	}
```

banyand/property/repair.go
```diff
@@ -0,0 +1,947 @@
...
+		return nil, nil
+	}
+
+	// otherwise, reading the leaf nodes
+	err = r.seekPosition(parent.leafStart, io.SeekStart)
+	if err != nil {
+		return nil, fmt.Errorf("seeking to leaf node offset %d in file %s failure: %w", r.footer.leafNodeFinishedOffset, r.file.Name(), err)
+	}
+	var entity, shaVal []byte
+	for i := int64(0); i < parent.leafCount; i++ {
```
hanahmily: Instead of loading all the data at once, you should implement pagination for the leaf nodes. When comparing leaf nodes between two slots, each slot should load a specific number of lines—such as 10—for comparison. If all the lines are identical, the page should automatically scroll to the next set of 10 items.

banyand/property/repair.go
```diff
@@ -0,0 +1,947 @@
...
+	topNSearch.SortBy([]string{
+		fmt.Sprintf("+%s", groupField),
+		fmt.Sprintf("+%s", nameField),
+		fmt.Sprintf("+%s", entityID),
+		fmt.Sprintf("+%s", timestampField),
+	})
+
+	var latestProperty *searchingProperty
+	treeComposer := newRepairTreeComposer(r.composeSlotAppendFilePath, r.composeTreeFilePathFmt, r.treeSlotCount, r.l)
+	if err != nil {
```
hanahmily: Where to set the err?

banyand/property/repair.go
```diff
@@ -0,0 +1,947 @@
...
+
+type repairTreeNode struct {
+	shaValue  string
+	id        string
+	tp        repairTreeNodeType
+	leafStart int64
+	leafCount int64
+}
+type repairTreeReader struct {
+	file   *os.File
```
hanahmily, line 327: How the file get closed after reading

banyand/property/repair.go
```diff
@@ -0,0 +1,947 @@
...
+	buildTreeScheduleInterval time.Duration
+	quickBuildTreeTime        time.Duration
+	buildTreeLocker           sync.Mutex
+}
+
+func newRepairScheduler(
+	l *logger.Logger,
+	omr observability.MetricsRegistry,
+	cronExp string,
+	quickBuildTreeTime time.Duration,
```
hanahmily, line 844: It is accessed and modified without proper synchronization, potentially causing race conditions.

banyand/property/repair.go
```diff
@@ -0,0 +1,947 @@
...
+		return fmt.Errorf("writing root node to repair tree file failure: %w", err)
+	}
+
+	// write footer
+	data = make([]byte, 0)
+	data = encoding.VarInt64ToBytes(data, leafNodesFinishedOffset) // end offset of the slot nodes
+	data = encoding.VarInt64ToBytes(data, int64(len(r.Slots)))     // count of the slot nodes
+	data = encoding.VarInt64ToBytes(data, slotNodesFinishedOffset) // end offset of the root node
+	data = encoding.VarInt64ToBytes(data, int64(writedLen))        // length of the root node
+	footerLen := uint8(len(data))
```
hanahmily: No validation that the footer length fits within uint8 bounds, which could cause silent truncation.

banyand/property/repair.go
```diff
@@ -0,0 +1,947 @@
...
+		return []*repairTreeNode{
+			{
+				shaValue: string(shaValue),
+				tp:       repairTreeNodeTypeRoot,
+			},
+		}, nil
+	}
+
+	var err error
+	nodes := make([]*repairTreeNode, 0)
```
hanahmily: No capacity hints for slices, leading to multiple reallocations during growth.

banyand/property/repair.go
```diff
@@ -0,0 +1,997 @@
...
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to add repair build tree cron task: %w", err)
+	}
+	interval, nextTime, exist := c.Interval("repair")
+	if !exist {
+		return nil, fmt.Errorf("failed to get repair build tree cron task interval")
+	}
+	s.buildTreeScheduleInterval = interval
+	s.latestBuildTreeSchedule = nextTime.Add(-interval)
```
hanahmily: r.lastBuildTimeLocker should protect them.

hanahmily (approved):

PR #694, opened 2025-07-03
Push down min/max aggregation operation to data nodes

sollhui:
- [x] a part of https://github.com/apache/skywalking/issues/13291.


sollhui:
@hanahmily I would like to implement pushing down mean operator in a separate PR, so this PR implements the aggregation operators pushing down besides the mean operator.

pkg/query/logical/measure/measure_plan_aggregation.go
```diff
@@ -39,18 +39,26 @@ var (
...
+	needCompletePushDownAgg bool
 }
 
-func newUnresolvedAggregation(input logical.UnresolvedPlan, aggrField *logical.Field, aggrFunc modelv1.AggregationFunction, isGroup bool) logical.UnresolvedPlan {
+func newUnresolvedAggregation(
+	input logical.UnresolvedPlan,
+	aggrField *logical.Field,
+	aggrFunc modelv1.AggregationFunction,
+	isGroup bool,
+	needCompletePushDownAgg bool,
```
hanahmily:
```suggestion
```

To aggregate a node and access the first field, modify the schema returned by distributedPlan.Schema(). The new schema.CreateFieldRef() should produce a field that represents the aggregated result.

pkg/query/logical/measure/measure_analyzer.go
```diff
@@ -143,8 +145,13 @@ func DistributedAnalyze(criteria *measurev1.QueryRequest, ss []logical.Schema) (
 		}
 	}
 
+	needCompletePushDownAgg := criteria.GetAgg() != nil &&
```
hanahmily, line 148:
```suggestion
       // TODO: to support all aggregation functions
	needCompletePushDownAgg := criteria.GetAgg() != nil &&
```

hanahmily (changes requested):

hanahmily (approved):

PR #695, opened 2025-07-05
Support dictionary encoding for low cardinality columns

ButterBright:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13272.

### DIFF
--- a/banyand/measure/column.go
+++ b/banyand/measure/column.go
@@ -104,15 +104,16 @@ func (c *column) mustWriteTo(cm *columnMetadata, columnWriter *writer) {
 
 	bb := bigValuePool.Generate()
 	defer bigValuePool.Release(bb)
+	tmp := make([]uint32, 0)
 
 	// select encoding based on data type
 	switch c.valueType {
 	case pbv1.ValueTypeInt64:
-		c.encodeInt64Column(bb)
+		c.encodeInt64Column(bb, tmp)
 	case pbv1.ValueTypeFloat64:
-		c.encodeFloat64Column(bb)
+		c.encodeFloat64Column(bb, tmp)
 	default:
-		c.encodeDefault(bb)
+		c.encodeDefault(bb, tmp)
 	}
 	cm.size = uint64(len(bb.Buf))
 	if cm.size > maxValuesBlockSize {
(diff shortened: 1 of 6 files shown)


pkg/encoding/dictionary.go
```diff
@@ -73,24 +73,30 @@ func (d *Dictionary) Encode(dst []byte, tmp []uint32) []byte {
 }
 
 // Decode decodes the dictionary.
-func (d *Dictionary) Decode(src []byte, tmp []uint32) error {
+func (d *Dictionary) Decode(dst [][]byte, src []byte, tmp []uint32, itemsCount uint64) ([][]byte, error) {
```
hanahmily: Could you move the `tmp` to a field of the Dictionary?

banyand/measure/column.go
```diff
@@ -205,8 +206,17 @@ func (c *column) encodeFloat64Column(bb *bytes.Buffer) {
 	)
 }
 
-func (c *column) encodeDefault(bb *bytes.Buffer) {
-	bb.Buf = encoding.EncodeBytesBlock(bb.Buf[:0], c.values)
+func (c *column) encodeDefault(bb *bytes.Buffer, tmp []uint32) {
+	dict := encoding.NewDictionary()
```
hanahmily: Reuse it by the pool.

banyand/measure/column.go
```diff
@@ -104,15 +104,16 @@ func (c *column) mustWriteTo(cm *columnMetadata, columnWriter *writer) {
 
 	bb := bigValuePool.Generate()
 	defer bigValuePool.Release(bb)
+	tmp := make([]uint32, 0)
```
hanahmily: Could you move the `tmp` to a field of `Dictionary`?

hanahmily (changes requested):
Add a benchmark as https://github.com/apache/skywalking-banyandb/blob/main/banyand/measure/benchmark_encode_test.go did.

ButterBright:
> Add a benchmark as https://github.com/apache/skywalking-banyandb/blob/main/banyand/measure/benchmark_encode_test.go did.

The benchmark was added in pkg/encoding/dictionary_test.go.

pkg/encoding/dictionary_test.go
```diff
@@ -114,13 +114,12 @@ func BenchmarkDecodeDictionary(b *testing.B) {
 				dict.Add(value)
 			}
 			encoded := make([]byte, 0)
-			tmp := make([]uint32, 0)
-			encoded = dict.Encode(encoded, tmp)
+			encoded = dict.Encode(encoded)
 
 			b.ResetTimer()
 			for i := 0; i < b.N; i++ {
 				decoded := NewDictionary()
```
hanahmily: Could you move this line outside the loop? Instead, you should reset inside the loop to verify the Decoder pool scenario.

hanahmily (changes requested):
Could you please update the benchmark results, as we have made many improvements to the allocation?

hanahmily (approved):

PR #697, opened 2025-07-05
Push down aggregation for topN query

sollhui:
- [x] close https://github.com/apache/skywalking/issues/13357.

### DIFF
--- a/banyand/query/processor.go
+++ b/banyand/query/processor.go
@@ -24,9 +24,12 @@ import (
 	"runtime/debug"
 	"time"
 
+	"golang.org/x/exp/slices"
+
 	"github.com/apache/skywalking-banyandb/api/common"
 	commonv1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/common/v1"
 	measurev1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/measure/v1"
+	modelv1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/model/v1"
 	streamv1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/stream/v1"
 	"github.com/apache/skywalking-banyandb/banyand/measure"
 	"github.com/apache/skywalking-banyandb/banyand/stream"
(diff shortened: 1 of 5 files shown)


banyand/query/processor.go
```diff
@@ -160,6 +161,62 @@ func (p *measureQueryProcessor) Rev(ctx context.Context, message bus.Message) (r
...
+			}
+			if len(entity) > 0 {
+				entities = append(entities, entity)
+			}
+		}
+		rewriteQueryCriteria := &measurev1.QueryRequest{
+			Groups:          queryCriteria.Groups,
+			Name:            queryCriteria.Name,
+			TimeRange:       queryCriteria.TimeRange,
+			Criteria:        queryCriteria.Criteria,
```
hanahmily: You should update `Criteria` like 

```yaml
criteria:
  le:
    op: "LOGICAL_OP_AND"
    left:
      condition:
        name: "tag1_in_groupBy"
        op: "BINARY_OP_IN"
        value:
          strArray:
            value: ["a", "b", "c"]
     right:
      condition:
        name: "tag2_in_groupBy"
        op: "BINARY_OP_IN"
        value:
          strArray:
            value: ["1", "2", "3"]
```

If there is only a tag in groupBy fileld, the expression can be simplied:

```yaml
criteria:
  condition:
    name: "tag1_in_groupBy"
    op: "BINARY_OP_IN"
    value:
      strArray:
        value: ["a", "b", "c"]
```

banyand/query/processor.go
```diff
@@ -160,6 +161,62 @@ func (p *measureQueryProcessor) Rev(ctx context.Context, message bus.Message) (r
...
+			TagProjection:   queryCriteria.TagProjection,
+			FieldProjection: queryCriteria.FieldProjection,
+		}
+		resp = p.executeQuery(ctx, rewriteQueryCriteria)
+		result = resp.Data().(*measurev1.QueryResponse).DataPoints
+		for _, dp := range result {
+			for _, entity := range entities {
```
hanahmily: You should use `criteria` to filter the data points.

hanahmily (changes requested):

banyand/query/processor.go
```diff
@@ -160,6 +163,53 @@ func (p *measureQueryProcessor) Rev(ctx context.Context, message bus.Message) (r
 		resp = bus.NewMessage(bus.MessageID(now), common.NewError("invalid event data type"))
 		return
 	}
+	resp = p.executeQuery(ctx, queryCriteria)
```
hanahmily, line 169: We should increase the top N size by a factor of 2, as indicated in https://github.com/apache/skywalking/issues/13357#issuecomment-3036272668.

If the original topn size is 5, there should be 10.

hanahmily (changes requested):

banyand/query/processor.go
hanahmily, line 169: If the resp.Data() returns `*common.Error`, you should return the resp directly.

banyand/query/processor.go
```diff
@@ -160,6 +163,53 @@ func (p *measureQueryProcessor) Rev(ctx context.Context, message bus.Message) (r
...
+		rewriteQueryCriteria := &measurev1.QueryRequest{
+			Groups:          queryCriteria.Groups,
+			Name:            queryCriteria.Name,
+			TimeRange:       queryCriteria.TimeRange,
+			Criteria:        rewriteCriteria,
+			TagProjection:   queryCriteria.TagProjection,
+			FieldProjection: queryCriteria.FieldProjection,
+		}
+		resp = p.executeQuery(ctx, rewriteQueryCriteria)
+		result = resp.Data().(*measurev1.QueryResponse).DataPoints
```
hanahmily: If the resp.Data() returns `*common.Error`, you should return the resp directly.

PR #698, opened 2025-07-09
Introduct gossip propagation protocol

mrproliu:
Introduce the gossip propagation protocol for the data nodes to communicate without liaison. It's based on the queue package as a basic protocol. 

When using gossip propagation message, it still needs to register the protocol and topic, same as the queue. And pass the `nodes` list to let these nodes communicate self, then return the Future for the result notification. 

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Part of Apache/Skywalking #13284.

### DIFF
--- a/banyand/gossip/client.go
+++ b/banyand/gossip/client.go
@@ -1,0 +1,98 @@
+// Licensed to Apache Software Foundation (ASF) under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Apache Software Foundation (ASF) licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package gossip
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"google.golang.org/protobuf/proto"
+
+	"github.com/apache/skywalking-banyandb/api/data"
+	gossipv1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/gossip/v1"
+	"github.com/apache/skywalking-banyandb/banyand/queue"
+	"github.com/apache/skywalking-banyandb/pkg/bus"
+)
+
+func (s *service) Propagation(nodes []string, topic string, message bus.Message) (Future, error) {
+	if len(nodes) < 2 {
+		return nil, fmt.Errorf("must provide at least 2 node")
+	}
+
+	// building propagation context
+	ctx := &gossipv1.PropagationContext{
+		Nodes:           nodes,
+		OriginNode:      s.nodeID,
+		OriginMessageId: uint64(message.ID()),
+	}
+	if len(nodes) == 2 {
+		ctx.MaxPropagationCount = 1
+	} else {
+		// two rounds of all nodes except the lasted node
+		// such when there have three nodes A, B, C,
+		// the propagation will be A -> B, B -> C, C -> A, A -> B
+		ctx.MaxPropagationCount = int32(len(nodes)*2 - 2)
+	}
+
+	// building propagation message request
+	d, ok := message.Data().(proto.Message)
+	if !ok {
+		return nil, fmt.Errorf("invalid message type %T", message.Data())
+	}
+	payload, err := proto.Marshal(d)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal message %T: %w", d, err)
+	}
+	request := &gossipv1.PropagationMessageRequest{
+		Topic:     topic,
+		MessageId: uint64(message.ID()),
+		Context:   ctx,
+		Body:      payload,
+	}
+
+	totalTimeout := time.Second * 15 * time.Duration(ctx.MaxPropagationCount)
+
+	var sendMsg bus.Message
+	var sendTo queue.Client
+	if nodes[0] == s.nodeID {
+		sendMsg = bus.NewMessage(message.ID(), request)
+		sendTo = s.local
+	} else {
+		sendMsg = bus.NewMessageWithNode(message.ID(), nodes[0], request)
+		sendTo = s.remoteClient
+	}
+
+	return s.handleSending(sendTo, sendMsg, ctx.MaxPropagationCount > 1, totalTimeout)
+}
+
+func (s *service) handleSending(sendTo queue.Client, msg bus.Message, hasRemote bool, timeout time.Duration) (Future, error) {
+	ctx, cancelFunc := context.WithTimeout(context.Background(), timeout)
+	defer cancelFunc()
+	publish, err := sendTo.Publish(ctx, data.TopicGossipPropagation, msg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to publish gossip message: %w", err)
+	}
+
+	f := newFuture(s, msg, publish, hasRemote, timeout)
+	// only add to wait list if there is a remote node to propagate
+	if hasRemote {
+		s.addToWait(f)
+	}
+	return f, nil
+}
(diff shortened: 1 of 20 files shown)


pkg/cmdsetup/standalone.go
```diff
@@ -50,7 +51,7 @@ func newStandaloneCmd(runners ...run.Unit) *cobra.Command {
 	}
 	metricSvc := observability.NewMetricService(metaSvc, dataPipeline, "standalone", nil)
 	pm := protector.NewMemory(metricSvc)
-	propertySvc, err := property.NewService(metaSvc, dataPipeline, metricSvc, pm)
+	propertySvc, err := property.NewService(metaSvc, dataPipeline, metricSvc, pm, gossip.NewLocalMessenger())
```
hanahmily: The standalone doesn't need repair since it is in this mode.

banyand/gossip/client.go
```diff
@@ -0,0 +1,98 @@
...
+		return nil, fmt.Errorf("failed to marshal message %T: %w", d, err)
+	}
+	request := &gossipv1.PropagationMessageRequest{
+		Topic:     topic,
+		MessageId: uint64(message.ID()),
+		Context:   ctx,
+		Body:      payload,
+	}
+
+	totalTimeout := time.Second * 15 * time.Duration(ctx.MaxPropagationCount)
```
hanahmily: The timeout is independent of the MaxPropagationCount.

pkg/cmdsetup/data.go
```diff
@@ -49,7 +52,9 @@ func newDataCmd(runners ...run.Unit) *cobra.Command {
 	metricSvc := observability.NewMetricService(metaSvc, metricsPipeline, "data", nil)
 	pm := protector.NewMemory(metricSvc)
 	pipeline := sub.NewServer(metricSvc)
-	propertySvc, err := property.NewService(metaSvc, pipeline, metricSvc, pm)
+	client := pub.New(metaSvc, databasev1.Role_ROLE_DATA)
```
hanahmily: You should avoid using the `pub` package. Its client may reconnect to the node, which is not needed by the gossip protocol. According to the design, gossip should stop syncing when the connection is lost.

banyand/gossip/client.go
hanahmily, line 1: Move the gossip pkg into `property`

pkg/cmdsetup/data.go
```diff
@@ -49,7 +52,9 @@ func newDataCmd(runners ...run.Unit) *cobra.Command {
...
+	messenger := gossip.NewMessenger(metricSvc, client, pipeline)
```
hanahmily: The gossip server should be initialized within the `property` module, as it is not a top-level module.

banyand/gossip/client.go
```diff
@@ -0,0 +1,98 @@
...
+		OriginMessageId: uint64(message.ID()),
+	}
+	if len(nodes) == 2 {
+		ctx.MaxPropagationCount = 1
+	} else {
+		// two rounds of all nodes except the lasted node
+		// such when there have three nodes A, B, C,
+		// the propagation will be A -> B, B -> C, C -> A, A -> B
+		ctx.MaxPropagationCount = int32(len(nodes)*2 - 3)
+	}
```
hanahmily: Why use two branches?  `2 * 2 - 3 != 1` ?

banyand/gossip/client.go
```diff
@@ -0,0 +1,98 @@
...
+		MessageId: uint64(message.ID()),
+		Context:   ctx,
+		Body:      payload,
+	}
+
+	totalTimeout := time.Second * 15 * time.Duration(ctx.MaxPropagationCount)
+
+	var sendMsg bus.Message
+	var sendTo queue.Client
+	if nodes[0] == s.nodeID {
```
hanahmily: This scenario should not happen: In a node, there are no two replicas for an identical shard.

banyand/gossip/client.go
```diff
@@ -0,0 +1,98 @@
...
+		sendMsg = bus.NewMessage(message.ID(), request)
+		sendTo = s.local
+	} else {
+		sendMsg = bus.NewMessageWithNode(message.ID(), nodes[0], request)
```
hanahmily: If this function always gets the first node, how about using a node string instead of a slice?

hanahmily: It's not a message that is used in a pipeline. You don't have to use the `bus` pkg.

hanahmily (changes requested):
The gossip pg should not use the terms `bus`, `pub`, `sub`, or `queue`, as they do not reflect the functionality of gossip protocols. There are differences:

1. Streaming in a `queue` is unidirectional, whereas gossip requires a bidirectional communication mechanism.
2. In a `queue`, clients continually attempt to reconnect to nodes upon failure. However, in gossip protocols, this behavior should be avoided.

pkg/cmdsetup/data.go
```diff
@@ -49,7 +50,8 @@ func newDataCmd(runners ...run.Unit) *cobra.Command {
 	metricSvc := observability.NewMetricService(metaSvc, metricsPipeline, "data", nil)
 	pm := protector.NewMemory(metricSvc)
 	pipeline := sub.NewServer(metricSvc)
-	propertySvc, err := property.NewService(metaSvc, pipeline, metricSvc, pm)
+	messenger := gossip.NewMessenger(metricSvc, metaSvc)
```
hanahmily: Would you please move Messenger into the property package? It should be an embedded component of the property service. The gRPC port could be returned by the property service.

banyand/property/gossip/client.go
```diff
@@ -0,0 +1,143 @@
...
+	var sendTo func(context.Context, *propertyv1.PropagationRequest) (*propertyv1.PropagationResponse, error)
+
+	// send it to the current node if it is the first node
+	if nodes[0] == s.nodeID {
+		sendTo = func(ctx context.Context, req *propertyv1.PropagationRequest) (*propertyv1.PropagationResponse, error) {
+			return (&protocolHandler{s: s}).Propagation(ctx, req)
+		}
+	} else {
+		sendTo = func(ctx context.Context, request *propertyv1.PropagationRequest) (*propertyv1.PropagationResponse, error) {
+			node, exist := s.registered[nodes[0]]
```
hanahmily: Data race

banyand/property/gossip/server.go
```diff
@@ -0,0 +1,326 @@
...
+	}
+	return s.listeners[0]
+}
+
+type protocolHandler struct {
+	propertyv1.UnimplementedGossipServiceServer
+	s *service
+}
+
+func (q *protocolHandler) Propagation(ctx context.Context, request *propertyv1.PropagationRequest) (resp *propertyv1.PropagationResponse, err error) {
```
hanahmily: Could you create a channel with a single size to avoid the concurrent propagation task, as mentioned in the document?

You might want the client to handle the error using the retry strategy. But I recommend that you do it on the node side. 

By doing this, the client side doesn't need the asynchronous access pattern (further)

banyand/property/gossip/server.go
```diff
@@ -0,0 +1,370 @@
...
+			continue
+		}
+
+		executeSuccess = true
+		break
+	}
+
+	// if no node is available or sync error for gossip message propagation, then it should be notified finished
+	if !executeSuccess {
+		_ = nextNodeConn.Close()
```
hanahmily: nextNodeConn might be nil.

banyand/property/gossip/server.go
```diff
@@ -0,0 +1,370 @@
...
+		totalFinished: factory.NewCounter("total_finished", "group"),
+		totalErr:      factory.NewCounter("total_err", "group"),
+		totalLatency:  factory.NewCounter("total_latency", "group"),
+
+		totalPropagationCount: factory.NewCounter("total_propagation_count", "group", "original_node"),
+		totalPropagationPercent: factory.NewHistogram("total_propagation_percent",
+			meter.Buckets{0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1}, "group"),
+
+		totalPropagationStarted:  factory.NewCounter("total_msg_received", "group"),
+		totalPropagationFinished: factory.NewCounter("total_msg_received_err", "group"),
```
hanahmily: Some metrics have inconsistent naming

hanahmily (changes requested):
Some nits.

hanahmily (approved):

PR #700, opened 2025-07-10
Feature/azure support

Eternity1824:
### feat: initial Azure Blob Storage support for backup and restore

This PR introduces the initial support for Azure Blob Storage (via Azurite) in the BanyanDB backup/restore framework.

- [x] Implemented `azure.NewFS` compatible with existing remote FS interface.
- [x] Integrated Azure logic into backup + timedir structure.
- [x] Added restore support for Azure.
- [x] Introduced Azurite mock container for integration testing.

---

### Notes:

CI currently fails due to a `panic` caused by `fs == nil`, likely related to Azurite not being fully ready during test initialization.  
We plan to fix this in the next update and provide a more complete Azurite-based verification to ensure Azure SDK behavior is correct under test conditions.


azurite_data/AzuriteConfig
hanahmily, line 1: `azurite_data` is a temporary directory? If so, please don't submit it. If it gets created at runtime, adding the folder name to the .gitignore is a good idea.

.licenserc.yaml
```diff
@@ -85,6 +85,7 @@ header: # `header` section is configurations for source codes license header.
     - 'ui'
     - '.github/PULL_REQUEST_TEMPLATE'
     - '**/*.prof'
+    - 'azurite_data/**'
```
hanahmily:
```suggestion
```
Do not include this rule since it is marked in .gitignore

pkg/fs/remote/config/credentials.json
hanahmily, line 1: Move this file to the test pkg if it's for testing purposes.

pkg/fs/remote/azure/azblob.go
```diff
@@ -0,0 +1,281 @@
...
+		}
+	}
+	return fmt.Errorf("failed to ensure container %s exists: %w", containerName, err)
+}
+
+func (b *blobFS) getFullPath(p string) string {
+	if b.basePath == "" {
+		return p
+	}
+	return path.Join(b.basePath, p)
```
hanahmily, line 170: The function uses `path.Join()` which is platform-specific. For cloud storage, it should handle path separators explicitly for cross-platform compatibility.

pkg/fs/remote/azure/azblob.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+func (b *blobFS) Upload(ctx context.Context, p string, data io.Reader) error {
```
hanahmily, line 173: It doesn't validate that data is not nil.

hanahmily (changes requested):

test/stress/istio/testdata/metrics/data.csv
hanahmily, line 1: These csv files should not be updated

hanahmily (approved):

PR #701, opened 2025-07-10
Fix panic in Bloom filter decoding

ButterBright:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13364.


banyand/stream/block.go
```diff
@@ -112,7 +113,7 @@ func (b *block) processTags(tf tagValues, tagFamilyIdx, i int, elementsLen int)
 			tags[j].filter = filter
 		}
 		tags[j].filter.SetN(elementsLen)
```
ButterBright, line 115: By the way, perhaps we should use the number of unique elements to initialize the Bloom filter?

hanahmily: It makes sense, especially for low-cardinality tags.

PR #711, opened 2025-07-28
Add chunked data sync to improve memory efficiency and performance during data transfer operations

hanahmily:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13380.

### DIFF
--- a/banyand/measure/flusher.go
+++ b/banyand/measure/flusher.go
@@ -211,7 +211,10 @@ func (tst *tsTable) flusherLoopNoMerger(flushCh chan *flusherIntroduction, intro
 				}()
 				curSnapshot := tst.currentSnapshot()
 				if curSnapshot != nil {
-					defer curSnapshot.decRef()
+					defer func() {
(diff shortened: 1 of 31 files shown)


wu-sheng (approved):

banyand/measure/flusher.go
```diff
@@ -211,7 +211,10 @@ func (tst *tsTable) flusherLoopNoMerger(flushCh chan *flusherIntroduction, intro
 				}()
 				curSnapshot := tst.currentSnapshot()
 				if curSnapshot != nil {
-					defer curSnapshot.decRef()
+					defer func() {
+						curSnapshot.decRef()
+						curSnapshot = nil
+					}()
```
ButterBright, line 217: I think stream need the change as well?

hanahmily: No, it's for debugging purposes. I will align it with the stream later.

PR #713, opened 2025-08-03
New storage engine for trace

ButterBright:
- [ ] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13329.


ButterBright:
I will continue to add unit tests in this PR and implement the traceID-based query in the next one.

wu-sheng:
@ButterBright Please fix CI.

banyand/trace/block.go
```diff
@@ -0,0 +1,659 @@
...
+	"github.com/apache/skywalking-banyandb/pkg/logger"
+	pbv1 "github.com/apache/skywalking-banyandb/pkg/pb/v1"
+	"github.com/apache/skywalking-banyandb/pkg/pool"
+	"github.com/apache/skywalking-banyandb/pkg/query/model"
+)
+
+type block struct {
+	spans      [][]byte
+	tags       []tag
+	timestamps []int64
```
hanahmily:
```suggestion
    minTS int64
    maxTS int64
```

banyand/trace/tag.go
```diff
@@ -0,0 +1,392 @@
...
+}
+
+var (
+	int64SlicePool   = pool.Register[*[]int64]("trace-int64Slice")
+	float64SlicePool = pool.Register[*[]float64]("trace-float64Slice")
+	dictionaryPool   = pool.Register[*encoding.Dictionary]("trace-dictionary")
+)
+
+type tag struct {
+	tagFilter
```
hanahmily:
```suggestion
```
The traceID storage doesn't support the skipping index

banyand/trace/block_metadata.go
```diff
@@ -0,0 +1,290 @@
...
+	return src
+}
+
+type blockMetadata struct {
+	tags                      map[string]*dataBlock
+	tagType                   map[string]pbv1.ValueType
+	spans                     *dataBlock
+	tagProjection             *model.TagProjection
+	traceID                   string
+	timestamps                timestampsMetadata
```
hanahmily, line 69:
```suggestion
```
traceID storage doesn't need timestamps

banyand/trace/block_metadata.go
```diff
@@ -0,0 +1,290 @@
...
+	d.size = n
+	return src
+}
+
+type blockMetadata struct {
+	tags                      map[string]*dataBlock
+	tagType                   map[string]pbv1.ValueType
+	spans                     *dataBlock
+	tagProjection             *model.TagProjection
+	traceID                   string
```
hanahmily, line 68: The spans in a block could come from different traces. It can not contain a single traceID here.

hanahmily: it could be the maxTracID and minTraceID

hanahmily (changes requested):
There are design issues with the block.

banyand/trace/block_metadata.go
ButterBright: We need it only for writing part metadata.

banyand/trace/part.go
```diff
@@ -0,0 +1,692 @@
...
+	var tidPrev string
+	uncompressedSpansSizeBytes := uint64(0)
+	var indexPrev int
+	for i := range ts.spans {
+		tid := ts.traceIDs[i]
+		if tidPrev == "" {
+			tidPrev = tid
+		}
+
+		if uncompressedSpansSizeBytes >= maxUncompressedSpanSize || tid != tidPrev {
```
ButterBright, line 458: Is it necessary to keep the first condition in case there are anomalous traces, considering that each block is expected to be relatively small?

hanahmily: In some cases, the span volume may be significantly larger. For example, a loop with a high limit could generate over a hundred spans.

wu-sheng: We saw 10k level spans in zipkin ecosystem.

ButterBright: Okay, I will keep it then.

PR #714, opened 2025-08-04
Introduce Trace System in background repair property feature

mrproliu:
The background repair property feature is complex, so it needs to introduce a trace system to track each step. It will be stored in the internal stream(`_property_gossip` group, `_property_gossip_trace_stream` name). 
It can be activated `property-repair-gossip-trace-log` flag(default is active). 


pkg/cmdsetup/data.go
```diff
@@ -49,7 +51,8 @@ func newDataCmd(runners ...run.Unit) *cobra.Command {
 	metricSvc := observability.NewMetricService(metaSvc, metricsPipeline, "data", nil)
 	pm := protector.NewMemory(metricSvc)
 	pipeline := sub.NewServer(metricSvc)
-	propertySvc, err := property.NewService(metaSvc, pipeline, metricSvc, pm)
+	pipelineClient := pub.New(metaSvc, databasev1.Role_ROLE_DATA)
+	propertySvc, err := property.NewService(metaSvc, pipeline, pipelineClient, metricSvc, pm)
```
hanahmily: To create a queue.Local(), then pass it to streamSvc. The pattern should like the metricsPipeline.

banyand/property/gossip/trace.go
```diff
@@ -0,0 +1,447 @@
...
+				return fmt.Errorf("failed to publish trace span: %w", err)
+			}
+		}
+	}
+	return nil
+}
+
+func (s *service) readAllReadySendTraceSpan() []*recordTraceSpan {
+	s.traceSpanLock.RLock()
+	defer s.traceSpanLock.RUnlock()
```
hanahmily:
```suggestion
	s.traceSpanLock.Lock()
	defer s.traceSpanLock.Unlock()
```

banyand/property/gossip/trace.go
```diff
@@ -0,0 +1,447 @@
...
+	spans := make([]*recordTraceSpan, 0, len(s.traceSpanSending))
+	spans = append(spans, s.traceSpanSending...)
+	s.traceSpanSending = make([]*recordTraceSpan, 0)
```
hanahmily:
```suggestion
	s.traceSpanSending = s.traceSpanSending[:0]
```

banyand/property/gossip/trace.go
```diff
@@ -0,0 +1,447 @@
...
+				Unit: commonv1.IntervalRule_UNIT_DAY,
+				Num:  1,
+			},
+			Ttl: &commonv1.IntervalRule{
+				Unit: commonv1.IntervalRule_UNIT_DAY,
+				Num:  3,
+			},
+		},
+	}
+	return s.metadata.GroupRegistry().CreateGroup(ctx, g)
```
hanahmily, line 170: You will face an error when there is a same name group.

PR #723, opened 2025-08-14
Update storage cache and protector implementations

hanahmily:


ButterBright (dismissed):

banyand/measure/part_iter.go
```diff
@@ -207,7 +207,9 @@ func (pi *partIter) readPrimaryBlock(bms []blockMetadata, mr *primaryBlockMetada
 	}
 	bmPtrs := make([]*blockMetadata, 0, len(bms))
 	for _, bm := range bms {
-		bmPtrs = append(bmPtrs, &bm)
+		bmCopy := &blockMetadata{}
```
hanahmily, line 210: @ButterBright, copying it tends to fix an issue, that is, `bm` is already cached by another pool.

wu-sheng (approved):

ButterBright (approved):

PR #724, opened 2025-08-15
Trace sidx design

hanahmily:
related to #713 

It's a detailed design of the secondary index.

### DIFF
--- a/banyand/internal/sidx/DESIGN.md
+++ b/banyand/internal/sidx/DESIGN.md
@@ -154,7 +154,6 @@ type partMetadata struct {
 type element struct {
     seriesID  common.SeriesID
     userKey   int64      // The ordering key from user (replaces timestamp)
-    elementID uint64     // Internal element identifier  
     data      []byte     // User payload data (pooled slice)
     tags      []tag      // Individual tags (pooled slice)
     
@@ -166,7 +165,6 @@ type element struct {
 func (e *element) reset() {
     e.seriesID = 0
     e.userKey = 0
-    e.elementID = 0
     if cap(e.data) <= maxPooledSliceSize {
         e.data = e.data[:0]  // Reuse slice if not too large
     } else {
@@ -188,7 +186,6 @@ func (e *element) reset() {
 type elements struct {
     seriesIDs []common.SeriesID  // Pooled slice
     userKeys  []int64           // Pooled slice
-    elementIDs []uint64         // Pooled slice
     data      [][]byte          // Pooled slice of slices
     tags      [][]tag           // Pooled slice of tag slices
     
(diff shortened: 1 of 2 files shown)


hanahmily:
> Please update the doc menu file.

This is not necessary for the end user; it is intended for contributors to align their objectives, given the many detailed codes and contexts involved. 

This documentation may be removed once the module is fully implemented.

wu-sheng (approved):

banyand/internal/sidx/DESIGN.md
```diff
@@ -0,0 +1,1095 @@
...
+func (pw *partWrapper) isRemovable() bool {
+    return atomic.LoadInt32(&pw.state) == int32(partStateRemoving)
+}
+
+// Enhanced part structure with resource management
+type part struct {
+    // File readers with standardized naming
+    primary     fs.Reader    // primary.bin - Block metadata
+    data        fs.Reader    // data.bin - User data payloads  
+    userKeys    fs.Reader    // keys.bin - User-provided int64 keys
```
ButterBright, line 115: We might need a map to support multiple secondary indices.

banyand/internal/sidx/DESIGN.md
```diff
@@ -0,0 +1,1095 @@
...
+## Use Case Examples
+
+### 1. Timestamp-based Secondary Index
+```go
+// User converts timestamp to int64
+func writeEvent(event Event) {
+    req := WriteRequest{
+        SeriesID: event.ServiceID,
+        Key:      event.Timestamp.UnixNano(),  // User generates
+        Data:     event.Payload,
```
ButterBright, line 906: Should the key be obtained from the schema rather than being passed in the request? And what is the data?

banyand/internal/sidx/DESIGN.md
```diff
@@ -0,0 +1,1095 @@
...
+            Min: int64(minMs * 1000),  // User converts to microseconds
+            Max: int64(maxMs * 1000),
+        },
+        Order: DESC,  // Highest latency first
+    }
+    return sidx.Query(req)
+}
+```
+
+### 3. Sequence Number Index
```
ButterBright, line 952: When do we use the sequence number index and the score-based index?

banyand/internal/sidx/DESIGN.md
hanahmily: This structure is designed for a single secondary index. In our case, we should create several secondary indexes in the initialization process. And put that map into a group-like wall structure.

banyand/internal/sidx/DESIGN.md
hanahmily: The key is to store the timestamps or the latency. Both of them are from a request. The data field is used to store the row ID, which, as we mentioned before, is a combination of two offsets.

banyand/internal/sidx/DESIGN.md
hanahmily: It is a case to explain that the secondary index use cases are more than the time series.

PR #725, opened 2025-08-15
Add benchmark and fix some issue of background property repair

mrproliu:
Add 1 UT and three E2E benchmark tests of the background property repair feature.

UT: Add Property Cluster Resilience test.
Benchmark: Support different data sizes of the background property repair test benchmark. 


wu-sheng:
About benchmarks, could you post the results here and eventually in commit logs?
We should document them if anything changed.

api/proto/banyandb/property/v1/repair.proto
```diff
@@ -71,6 +71,19 @@ message PropertySync {
 
 message NoMorePropertySync {}
```
hanahmily:
```suggestion
```

hanahmily (approved):

PR #729, opened 2025-08-19
Secondary index for trace: interface

hanahmily:
related to https://github.com/apache/skywalking/issues/13330


wu-sheng:
Please fix CI.

wu-sheng (approved):

PR #737, opened 2025-08-27
Add BDSL documentation for BanyanDB Query Language

hanahmily:
This change introduces a comprehensive document outlining the BanyanDB Data Query Language (BDSL), detailing its purpose, core concepts, and syntax for querying various data models, including Streams, Measures, Properties, and Top-N queries. The document includes grammar specifications, mapping to request messages, and practical examples for each query type, enhancing user understanding and interaction with BanyanDB's data models.

### DIFF
--- a/docs/concept/dbsl.md
+++ b/docs/concept/dbsl.md
@@ -23,20 +23,18 @@ It also provides a specialized syntax for optimized **Top-N** queries against me
 
 BDSL queries will be processed by a classic three-stage compiler front-end architecture, which ensures modularity and maintainability.
 
-```mermaid
-graph TD
-    A[BDSL Query String] --> B{Lexer};
-    B --> C{Parser};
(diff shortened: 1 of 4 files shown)


docs/concept/dbsl.md
```diff
@@ -0,0 +1,354 @@
+# BanyanDB Query Language (BDSL)
```
wu-sheng: Please link this in the menu, as this is an important no matter in CLI, UI, and even in the future in API.

wu-sheng: The contents are good for me. Just one to confirm, as you are considering the API endpoint as the default context(default group, metrics, etc.).
Is the first implementation still CLI only?

wu-sheng:
`BDSL` should be renamed to `BDQL`

wu-sheng (approved):

wankai123 (approved):

PR #755, opened 2025-09-05
Enhance query access logging across gRPC services

hanahmily:
- Updated Query methods in measure, property, stream, and trace services to log query execution time and errors.
- Introduced a new WriteQuery method in the access log interface to capture detailed timing information for queries.

### DIFF
--- a/banyand/trace/query.go
+++ b/banyand/trace/query.go
@@ -52,6 +52,9 @@ func (t *trace) Query(ctx context.Context, tqo model.TraceQueryOptions) (model.T
 	if tqo.TimeRange == nil {
 		return nil, errors.New("invalid query options: timeRange are required")
 	}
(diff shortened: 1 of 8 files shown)


banyand/trace/query.go
```diff
@@ -52,6 +52,9 @@ func (t *trace) Query(ctx context.Context, tqo model.TraceQueryOptions) (model.T
 	if tqo.TimeRange == nil {
 		return nil, errors.New("invalid query options: timeRange are required")
 	}
+	if len(tqo.TraceIDs) == 0 && tqo.Order == nil {
```
hanahmily, line 55: @wu-sheng @wankai123 the validation of trace ids and order is introduced

wu-sheng (approved):

PR #757, opened 2025-09-06
Fix bugs time query range will not automatically update and add some new time query ranges

peachisai:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#<issue number>.

https://github.com/apache/skywalking/issues/13476
image


Fine0830 (commented):
Please address feedbacks from Copilot and execute `npm run format` to fix CI

PR #761, opened 2025-09-09
[RFC][Draft] eBPF Sidecar Design Document

Eternity1824:
## Summary

This PR introduces a **design document** for the eBPF Sidecar Agent of BanyanDB.

The doc outlines the high-level scope, architecture, and modularization plan for kernel-level observability, focusing on **page cache miss monitoring** and related metrics.

## Motivation

- Provide a dedicated place for discussion and review of the design.
- Align with the maintainer on **scope** and **planned modularization** before implementation.

## Key Topics

- High-level architecture of the eBPF sidecar (collector, exporter, loader, server).
- Planned modularization (I/O monitor, cache monitor, export, integration).

## Notes

- The design doc is **draft / RFC**, not final.
- Attachment points (tracepoints/kprobes) listed are **illustrative**, subject to change during implementation.
- No implementation code is included in this PR.

## Next Steps

- Discuss with @hanahmily in the upcoming meeting.
- Based on feedback, refine scope and split follow-up design docs by module.


docs/ebpf-sidecar-design.md
```diff
@@ -0,0 +1,605 @@
+# eBPF Sidecar Design Document
+
+## 1. Executive Summary
+
+The eBPF Sidecar Agent is a kernel-level observability component for Apache SkyWalking BanyanDB that primarily monitors **page cache miss rates** - a critical metric for database performance. High cache miss rates indicate that BanyanDB is frequently reading from disk instead of memory, directly impacting query latency and throughput. By tracking these metrics at the kernel level, we can identify performance bottlenecks and optimize storage patterns.
```
wu-sheng: A sidecar deployment is a relatively heavy resource cost. Please be clear, is `page cache miss rate` the only purpose of this deployment? This is more like a on-demand tooling rather than a sidecar.
@hanahmily

docs/ebpf-sidecar-design.md
```diff
@@ -0,0 +1,605 @@
...
+
+```
+┌─────────────────────────────────────────────────────────┐
+│                     User Applications                   │
+│  ┌──────────────┐   ┌──────────────┐   ┌──────────────┐ │
+│  │   BanyanDB   │   │  Prometheus  │   │   Grafana    │ │
+│  └──────┬───────┘   └──────┬───────┘   └──────┬───────┘ │
+└─────────┼──────────────────┼──────────────────┼─────────┘
+          │ Native Export    │ Scrape           │ Query
+          ▼                  ▼                  ▼
```
wu-sheng: Why does Grafana even get involved? This seems unreasonable. Query from Grafana should go to Prometheus.
And if Prometheus needs to scrape metrics from BanyanDB, please do via BanyanDB Prometheus endpoint directly. 
There is no point to propagate and store metrics in two places.

docs/ebpf-sidecar-design.md
hanahmily: There are several concerns to address:

1. We want to avoid including the eBPF runtime in the images of the data nodes. 
2. The page cache monitoring agent is a process that periodically analyzes relevant metrics and exports the results through Prometheus endpoints. It is not an on-demand tool.
3. Regarding the integration of this agent with the banyand process: it may be possible, but I do not currently intend to implement it, as I do not want this code to crash the data nodes or render them unbootable.

docs/ebpf-sidecar-design.md
hanahmily: Exactly. Grafana should query metrics from prom .

docs/ebpf-sidecar-design.md
wu-sheng: It is acceptable to have ebpf only for sidecar, but please make sure this is an optional in deployment. In product cases, we should only have this through operator on-demand, because even with cache missing, in most cases, there is nothing we cam do unless optimize codes and redeploy.

hanahmily: Yes, its deployment strategy will resemble a backup or lifecycle approach. We will use it only if we want to explore the role of the page cache on query performance.

docs/ebpf-sidecar-design.md
Eternity1824: Got it. Metrics will be scraped by Prometheus from BanyanDB, and Grafana will only be used for visualization by querying Prometheus.

PR #771, opened 2025-09-16
Introduce the disk monitor to force cleanup disk

hanahmily:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13325, apache/skywalking#13324.

### DIFF
--- a/docs/operation/disk-management.md
+++ b/docs/operation/disk-management.md
@@ -4,7 +4,7 @@ BanyanDB provides comprehensive disk management capabilities to prevent disk spa
 
 ## Overview
 
-BanyanDB includes a disk monitor that automatically triggers forced retention cleanup when disk usage exceeds configured thresholds. This feature helps prevent disk space exhaustion by automatically removing old data segments and snapshots.
+BanyanDB includes a disk monitor that can automatically trigger forced retention cleanup when disk usage exceeds configured thresholds. This feature helps prevent disk space exhaustion by automatically removing old data segments and snapshots.
 
 ## Configuration
 
@@ -18,6 +18,7 @@ The following flags are used to configure forced retention cleanup for data and 
 - `--measure-retention-low-watermark float`: Disk usage percentage where forced retention cleanup stops (0-100, default: 85.0).
 - `--measure-retention-check-interval duration`: Interval for checking disk usage (default: 5m).
 - `--measure-retention-cooldown duration`: Cooldown period between forced segment deletions (default: 30s).
+- `--measure-retention-force-cleanup-enabled bool`: Enable forced retention cleanup when disk usage exceeds high watermark (default: false).
 
 #### Stream Service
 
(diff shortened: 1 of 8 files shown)


docs/operation/disk-management.md
```diff
@@ -0,0 +1,185 @@
...
+
+- `--measure-max-disk-usage-percent int`: The maximum disk usage percentage allowed for measure data (0-100, default: 95).
+- `--stream-max-disk-usage-percent int`: The maximum disk usage percentage allowed for stream data (0-100, default: 95).
+- `--trace-max-disk-usage-percent int`: The maximum disk usage percentage allowed for trace data (0-100, default: 95).
+
+**Write Queue Mechanism**: Liaison servers maintain a write queue that buffers incoming data before syncing it to data servers. When the queue fills up and disk usage exceeds the configured threshold, the liaison server throttles incoming writes to allow the sync process to catch up and free up disk space.
+
+### Property Service
+
+- `--property-max-disk-usage-percent int`: The maximum disk usage percentage allowed (0-100, default: 95).
```
wu-sheng, line 51: What is the policy of this? Reject write and the client received an error? Please be clear.

docs/forced-ttl-cleanup-design.md
```diff
@@ -0,0 +1,152 @@
+# Design: Forced TTL Cleanup on High Disk Usage
+
+**Author**: Gemini
+
+**Date**: 2025-09-12
+
+**Status**: Proposed
```
wu-sheng: What is this?

hanahmily: A brainstorm result from the Gemini. I will make sure to remove it later.

docs/operation/disk-management.md
hanahmily: Write throttle like other models. I will clarify this.

docs/forced-ttl-cleanup-design.md
wu-sheng: I am not sure about `**Author**: Gemini` part is acceptable. Owner means copyright owner, it doesn't fit ASF policy. You should be the owner of it, or keep it locallly or your fork.

docs/operation/disk-management.md
```diff
@@ -0,0 +1,185 @@
+# Disk Management
+
+BanyanDB provides comprehensive disk management capabilities to prevent disk space exhaustion and ensure optimal storage utilization. This includes both automatic forced retention cleanup and disk usage monitoring.
+
+## Overview
+
+BanyanDB includes a disk monitor that automatically triggers forced retention cleanup when disk usage exceeds configured thresholds. This feature helps prevent disk space exhaustion by automatically removing old data segments and snapshots.
+
+**Note**: The disk monitor with forced retention cleanup is only available for **standalone** and **data** servers. **Liaison** servers use the `*-max-disk-usage-percent` flags to throttle writes when the write queue fills up, allowing the sync process to catch up with incoming data.
```
wu-sheng: You only mentioned that the liasion will throttle writes, but how about data node?

I assume all writes could be throttled until more available volume space is available no matter of clean up and manual operations.

wu-sheng (approved):

PR #777, opened 2025-09-21
Fix memory leaks and OOM issues in streaming processing

hanahmily:
Fix the issue by implementing deduplication logic in priority queues
 and improving sliding window memory management.

related to apache/skywalking#13501


hanahmily:
@mrproliu, could you redeploy the BanyanDB with this change to verify if the OOM issue persists?

PR #778, opened 2025-09-21
feat: add QL protocol

kezhenxu94:


wu-sheng:
Do we have specific reasons to add type specific query API?

kezhenxu94:
> Do we have specific reasons to add type specific query API?

that is the only way to support optional `FROM` clause:

https://github.com/apache/skywalking-banyandb/blob/04e17d0bb19fbcaa0e1eb8443d2029a222635dbb/docs/interacting/bydbql.md?plain=1#L54

wu-sheng:
API doc seems not updated.

api/proto/banyandb/bydbql/v1/query.proto
```diff
@@ -0,0 +1,177 @@
...
+option java_package = "org.apache.skywalking.banyandb.bydbql.v1";
+
+// QueryExecutionContext provides resource context when FROM clause is omitted
+message QueryExecutionContext {
+  // catalog indicates the type of resource (STREAM, MEASURE, PROPERTY, TRACE)
+  common.v1.Catalog catalog = 1 [(validate.rules).enum.defined_only = true];
+  // name is the identity of the resource
+  string name = 2 [(validate.rules).string.min_len = 1];
+  // group indicates where the data point is stored
+  string group = 3;
```
hanahmily: `group` is mandatory.

api/proto/banyandb/bydbql/v1/query.proto
```diff
@@ -0,0 +1,177 @@
...
+    measure.v1.QueryResponse measure_result = 2;
+    // property_result is returned for property queries
+    property.v1.QueryResponse property_result = 3;
+    // trace_result is returned for trace queries
+    trace.v1.QueryResponse trace_result = 4;
+    // topn_result is returned for TopN queries
+    measure.v1.TopNResponse topn_result = 5;
+  }
+  // trace contains the trace information of the query when trace is enabled
+  common.v1.Trace trace = 6;
```
hanahmily: Each response contains a `trace` field. What's the purpose of this `trace`?

api/proto/banyandb/bydbql/v1/query.proto
```diff
@@ -0,0 +1,177 @@
...
+  // error contains error information if the query failed
+  string error = 7;
```
hanahmily: If the query failed, a gRPC error will be sent to the client. I don't see the use case of this field

api/proto/banyandb/bydbql/v1/rpc.proto
```diff
@@ -0,0 +1,112 @@
...
+  // The stream name is provided in the URL path, eliminating the need
+  // for a FROM clause in the query string
+  rpc QueryStream(QueryStreamRequest) returns (QueryStreamResponse) {
+    option (google.api.http) = {
+      post: "/v1/bydbql/stream/{group}/{name}/query"
+      body: "*"
+      additional_bindings: {
+        post: "/v1/bydbql/stream/{name}/query"
+        body: "*"
+      }
```
hanahmily:
```suggestion
```

As I mentioned above, the `group` is not optional.

PR #784, opened 2025-09-23
Add the span_id field in Span and WriteRequest

ButterBright:


wu-sheng:
According to latest discussion, as we already have span ID tag in implementation, we should use `span_id_tag_name`.

PR #788, opened 2025-09-24
Fix syncSnapshot for trace

ButterBright:


hanahmily:
@ButterBright  Please change all the debug log levels to "debug." Then, enable them by setting the "--logging-modules" and "--logging-levels" options to activate logging in the liaison and data nodes. Once these changes are made, we can merge this PR to verify the issue under different scenarios.

banyand/trace/syncer.go
```diff
@@ -189,24 +189,16 @@ func (tst *tsTable) syncSnapshot(curSnapshot *snapshot, syncCh chan *syncIntrodu
...
-		if len(sidxParts) == 0 {
-			continue
-		}
-		hasSidxParts = true
-		break
-	}
-	if len(partsToSync) == 0 && !hasSidxParts {
-		return nil
+	for name, sidxParts := range sidxPartsToSync {
+		tst.l.Debug().
```
hanahmily: Wrap with "if l:=tst.l.Debug();l.Enabled(){}"

hanahmily (changes requested):
Wrap all debug logs within "if l := tst.l.Debug(); l.Enabled() {}" to decrease the performance impact of debugging logs.

PR #794, opened 2025-09-29
 Deduplicate spans across replicas using span_id

ButterBright:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13516.


api/proto/banyandb/database/v1/schema.proto
```diff
@@ -243,4 +243,6 @@ message Trace {
   string timestamp_tag_name = 4 [(validate.rules).string.min_len = 1];
   // updated_at indicates when the trace resource is updated.
   google.protobuf.Timestamp updated_at = 5;
+  // span_id_tag_name is the name of the tag that stores the span ID.
+  string span_id_tag_name = 6 [(validate.rules).string.min_len = 1];
```
wu-sheng, line 247: I think this one is only required in sync and client?

wu-sheng: fyi @wankai123

ButterBright: Yes, only the client requires it for data writes.

wu-sheng:
You have to skip e2e as the new span id tag is mandatory in schema, and we don't have that in OAP.
Chicken and egg story.

ButterBright:
Okay.

wu-sheng:
The latest oap image includes this setup.
Feel free to try it.

wu-sheng:
I think lifecycle is not working anymore because of trace module.

PR #796, opened 2025-09-30
Update api to support deduplication of spans across replicas

ButterBright:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13516.


hanahmily (approved):

wu-sheng:
Due to protobuf compability, we have used new client with current DB version, just ignoring the new field.
So, you could focus on the implementation, and use upcoming OAP to test.

PR #797, opened 2025-09-30
Add trace query command in bydbctl

mrproliu:

### DIFF
--- a/banyand/internal/sidx/block_reader_test.go
+++ b/banyand/internal/sidx/block_reader_test.go
@@ -206,7 +206,7 @@ func Test_blockReader_nextBlock(t *testing.T) {
 					mpp = append(mpp, mp)
 					mp.mustInitFromElements(elems)
 					mp.mustFlush(fileSystem, partPath(tmpPath, uint64(i)))
-					filePW := newPartWrapper(nil, mustOpenPart(partPath(tmpPath, uint64(i)), fileSystem))
+					filePW := newPartWrapper(nil, mustOpenPart(uint64(i), partPath(tmpPath, uint64(i)), fileSystem))
 					filePW.p.partMetadata.ID = uint64(i)
 					fpp = append(fpp, filePW)
 					pp = append(pp, filePW.p)
--- a/banyand/internal/sidx/interfaces.go
+++ b/banyand/internal/sidx/interfaces.go
@@ -52,9 +52,9 @@ type SIDX interface {
 	// Flush flushes the SIDX instance to disk.
 	Flush() error
 	// Merge merges the specified parts into a new part.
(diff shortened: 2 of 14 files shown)


bydbctl/internal/cmd/trace_test.go
```diff
@@ -18,20 +18,33 @@
...
 	"github.com/apache/skywalking-banyandb/pkg/test/setup"
+	"github.com/apache/skywalking-banyandb/pkg/timestamp"
+	cases_trace_data "github.com/apache/skywalking-banyandb/test/cases/trace/data"
 )
 
+func TestStable(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Stable Suite", Label("integration", "slow"))
+}
+
```
hanahmily:
```suggestion
```

Each pkg only needs a Test function.

bydbctl/internal/cmd/trace.go
```diff
@@ -137,8 +137,22 @@ func newTraceCmd() *cobra.Command {
...
-	traceCmd.AddCommand(getCmd, createCmd, deleteCmd, updateCmd, listCmd)
+	queryCmd := &cobra.Command{
+		Use:     "query [-s start_time] [-e end_time] -f [file|dir|-]",
+		Version: version.Build(),
+		Short:   "Query data in a trace",
+		Long:    timeRangeUsage,
+		RunE: func(cmd *cobra.Command, _ []string) (err error) {
+			return rest(func() ([]reqBody, error) { return parseTimeRangeFromFlagAndYAML(cmd.InOrStdin()) },
+				func(request request) (*resty.Response, error) {
+					return request.req.SetBody(request.data).Post(getPath("/api/v1/trace/data"))
```
hanahmily: Convert the string to a global constant.

banyand/liaison/grpc/discovery.go
```diff
@@ -234,7 +234,9 @@ func (e *entityRepo) OnAddOrUpdate(schemaMetadata schema.Metadata) {
 	case schema.KindTrace:
 		trace := schemaMetadata.Spec.(*databasev1.Trace)
 		id = getID(trace.GetMetadata())
+		e.RWMutex.Lock()
```
hanahmily: It is good but does not cover all scenarios: e.traceIDIndexMap should be protected by a mutex.

hanahmily (changes requested):

bydbctl/internal/cmd/trace.go
hanahmily: https://github.com/apache/skywalking-banyandb/blob/main/bydbctl/internal/cmd/trace.go#L31

put "/api/v1/trace/data" here

PR #1001, opened 2026-03-10
Support float64 in the streaming process

### LINKED ISSUE #10277: [BanyanDB] Support float64 in the streaming process
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

This is a follow-up task after https://github.com/apache/skywalking-banyandb/pull/240

The server side should handle 64-bit float data more than 64-bit integers. 

### Use case

_No response_

### Related issues

_No response_

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

eye-gu:
### <Feature description>

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#10277.

1. Support float64 type field for topn
2. Fix some errors in topn when distributed scenarios


banyand/measure/topn.go
hanahmily, line 1: 1. Utilize Go's generic types to implement topNStreamingProcessor, TopNValue, and related components. It is preferable to minimize the "switch" logic in this context.
2. Convert []float to []int using encoding.Float64ListToDecimalIntList. Subsequently, marshal the int64 list following the existing logic.

hanahmily (changes requested):

banyand/measure/topn.go
eye-gu: Thank you for the valuable feedback. I have refactored the code to use Go's generic types for TopNValue and related components. And for float64 encoding, used encoding.Float64ListToDecimalIntList to convert float64 values to int64 before marshaling.
Please let me know if you have any further suggestions or concerns.

pkg/flow/streaming/topn.go
hanahmily, line 1: Introduce a generic type:

```
// TopSortKey is the constraint for TopN sort keys (int64 or float64).
type TopSortKey interface {
	int64 | float64
}
```

topNAggregatorGroup, TopNOption and etc should this type.

pkg/query/logical/measure/topn_plan_localscan.go
```diff
@@ -207,49 +207,78 @@ func (ei *topNMIterator) Next() bool {
...
 			}
-			dp.TagFamilies = append(dp.TagFamilies, tagFamily)
-			for k, entityName := range entityNames {
-				tagFamily.Tags = append(tagFamily.Tags, &modelv1.Tag{
-					Key:   entityName,
-					Value: entities[j][k],
-				})
+		} else {
+			topNValue := measure.GenerateTopNValueInt()
+			defer measure.ReleaseTopNValueInt(topNValue)
```
hanahmily, line 232: Move them outside the loop. Reuse the topNValue in the loop. defer's scope is the function not the for loop body.

banyand/dquery/topn.go
```diff
@@ -207,6 +215,53 @@ func (s *sortedTopNList) Val() *comparableTopNItem {
 	return &comparableTopNItem{s.Items[s.index-1]}
 }
 
+func processTopNResponse[N aggregation.Number](
```
hanahmily, line 218: In the old code, partial results from successful nodes were still used even when some nodes errored. Now, if 1 out of 10 nodes fails, the entire query returns an error. This is a regression for distributed scenarios.

banyand/measure/topn_post_processor.go
```diff
@@ -207,24 +253,34 @@ func (taggr *topNPostProcessor) Put(entityValues pbv1.EntityValues, val int64, t
...
 
-	newItem := &topNAggregatorItem{
+	newItem := &topNAggregatorItem[N]{
 		val:     val,
 		key:     key,
 		values:  entityValues,
 		version: version,
 	}
 
+	if taggr.topN <= 0 {
```
hanahmily, line 263: This will cause OOM

banyand/dquery/topn.go
```diff
@@ -103,6 +131,10 @@ func (t *topNQueryProcessor) Rev(ctx context.Context, message bus.Message) (resp
 			span.Stop()
 		}()
 	}
+	originalTopN := request.GetTopN()
+	// Set topN to 0 to disable truncation on data nodes for all aggregation functions.
+	// This ensures coordinator-side aggregation can see all relevant candidates.
+	request.TopN = 0
```
hanahmily, line 137: For high-cardinality entities, this could OOM data nodes during distributed TopN queries.

banyand/dquery/topn.go
```diff
@@ -46,6 +49,31 @@ type topNQueryProcessor struct {
 	*bus.UnImplementedHealthyListener
 }
 
+func (t *topNQueryProcessor) getTopNFieldType(ctx context.Context, group, topNName string) databasev1.FieldType {
```
hanahmily, line 52: It falls back silently to FIELD_TYPE_UNSPECIFIED → treated as int64. It's not correct.

banyand/measure/topn.go
```diff
@@ -881,8 +1021,157 @@ func (t *TopNValue) marshal(dst []byte) ([]byte, error) {
...
+	return databasev1.FieldType_FIELD_TYPE_INT, nil
+}
+
+// MergeTopNBinaryValues merges two TopN binary values and returns the merged binary value.
+func MergeTopNBinaryValues(
+	left, right []byte, topN int32, sort modelv1.Sort, decoder *encoding.BytesBlockDecoder,
+	timestamp uint64, leftVersion, rightVersion int64,
+) ([]byte, error) {
+	fieldType, detectErr := DetectFieldTypeFromBinary(left)
+	if detectErr != nil {
```
hanahmily, line 1097: If left fails detection but right succeeds and returns FLOAT, then left (which might be int64 data from an older format) will be deserialized as float64 — causing data corruption.

hanahmily (changes requested):

banyand/dquery/topn.go
```diff
@@ -111,53 +143,26 @@ func (t *topNQueryProcessor) Rev(ctx context.Context, message bus.Message) (resp
...
-					}
-					entityValues := make(pbv1.EntityValues, 0, len(tn.Entity))
-					for _, e := range tn.Entity {
-						entityValues = append(entityValues, e.Value)
-					}
-					aggregator.Put(entityValues, tn.Value.GetInt().GetValue(), uint64(tn.Timestamp.AsTime().UnixMilli()), tn.Version)
-				}
-			}
-		}
+	fieldType := t.getTopNFieldType(ctx, request.Groups[0], request.GetName())
```
paultanay, line 146: I think there is still a robustness regression in `banyand/dquery/topn.go`: the new field-type lookup indexes `request.Groups[0]`, but `validateRequest()` still doesn’t check that `Groups` is non-empty.

`TopNRequest` does declare `groups` with `min_items = 1`, so normal gRPC requests should usually be rejected earlier, but this path can still be reached outside the interceptor layer. Since the old distributed TopN path didn’t index `Groups[0]`, it looks safer to guard this explicitly here as well.

hanahmily: It makes sense

banyand/dquery/topn.go
eye-gu: > For high-cardinality entities, this could OOM data nodes during distributed TopN queries.

min/max can be truncated via TopN at the datanode, but sum/count/mean must read the full dataset to calculate accurate results. To avoid OOM, one approach is to follow a MapReduce pattern, but even then, sum/count/mean still need to collect all the pre-calculated data. If you agree, I can open a separate issue to fix this bug first, and then come back to this PR.

hanahmily:
@eye-gu I will close this. Feel free to reopen it if you would like to continue.

eye-gu:
> @eye-gu I will close this. Feel free to reopen it if you would like to continue.

Okay.

The core issue right now is that setting TopN to 0 to fetch all data can lead to OOM. As I mentioned earlier, min/max can be safely truncated, but truncating count/sum/mean will produce incorrect results. 

My thought is to implement TopN pushdown to the DataNode to reduce the data volume returned. I agree that this optimization shouldn't be included in this PR. 

If you agree, I will open a separate issue to address this OOM problem first, and then come back to this PR later.

hanahmily:
> > @eye-gu I will close this. Feel free to reopen it if you would like to continue.
> 
> Okay.
> 
> The core issue right now is that setting TopN to 0 to fetch all data can lead to OOM. As I mentioned earlier, min/max can be safely truncated, but truncating count/sum/mean will produce incorrect results.
> 
> My thought is to implement TopN pushdown to the DataNode to reduce the data volume returned. I agree that this optimization shouldn't be included in this PR.
> 
> If you agree, I will open a separate issue to address this OOM problem first, and then come back to this PR later.


The issue you mentioned is unrelated to this PR. Why do you switch to another issue?

I have left a comment on the new issue. What you're concerned about does not seem to be a problem.

eye-gu:
@hanahmily I have pushed new commits. Could you please reopen this PR and take another look? Thank you.

PR #1063, opened 2026-04-11
feat(topn): PushDown TopN to all data nodes and reduce at liaison

### LINKED ISSUE #13811: [Bug] Fix distributed TopNAggregation accuracy for SUM, COUNT, and MEAN aggregations
### Search before asking

## Problem

In distributed TopN queries, each data node sets `agg` to `AGGREGATION_FUNCTION_UNSPECIFIED` and applies the TopN limit locally before sending results to the liaison node. This produces incorrect results for SUM, COUNT, and MEAN aggregations in two ways:

1. **Cross-node truncation**: An entity that ranks low on one data node might rank high globally after merging partial results across all nodes.

2. **Intra-node truncation**: A single data node may have multiple data points for the same entity (e.g., across different shards), but only the largest one survives the local TopN truncation. For example, with COUNT aggregation and TopN=3, if a node has 2 data points for entity5, only the larger one is sent to liaison, causing the count to be 1 instead of 2.

MIN and MAX are not affected since they are monotonic — the global MIN/MAX always appears in at least one node's local TopN, and truncation within a node doesn't discard the extremum.

## Proposed Solution

- For MIN/MAX: keep applying TopN limit at data nodes (safe to truncate)
- For SUM/COUNT/MEAN: push the TopN query to all data nodes **without truncation** (TopN=0) and with the actual `agg` function and `emit_partial=true`, so data nodes return partial aggregated values (e.g., `[sum, count]` for MEAN) along with `shard_id` metadata
- At the liaison node, merge partial results by entity using a min-heap aggregator, then compute the final TopN
- Include `shard_id` in the group-by key for non-MIN/MAX to prevent incorrect cross-shard merging

Distributed TopN queries should return accurate results for all aggregation types (MIN, MAX, SUM, COUNT, MEAN) regardless of the number of data nodes.

For SUM/COUNT/MEAN, the final TopN should reflect the true aggregated value computed from all data points across all data nodes, not just the top-N per-node subset.

### How to reproduce

1. Deploy BanyanDB in distributed mode with at least 2 data nodes.
2. Ingest data points where the same entity has data on multiple data nodes or across multiple shards.
3. Issue a TopN query with COUNT, SUM, or MEAN aggregation.
4. Expected: entity with true aggregated value rank N should appear in TopN.
   Actual: entity is missing because its per-node/sub-shard partial value was below the local TopN threshold and was truncated.

### Anything else

_No response_

### Code of Conduct

eye-gu:

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13811

## Changes
﻿
### Proto
- **query.proto**: Add `shard_id` to `DataPoint` for carrying shard metadata in distributed query responses
- **topn.proto**: Add `values` (repeated field for mean results) and `shard_id` to `TopNList.Item`; add `emit_partial` to `TopNRequest`
﻿
### Data node
- **query/processor_topn.go**: When `emit_partial=true`, return all field values and `shard_id` instead of a single aggregated value
- **measure/topn.go**: Include `shard_id` in `TopNDataPoint`
﻿
### Liaison node
- **dquery/topn.go**: Set `TopN=0` for non-MIN/MAX aggregations, set `emit_partial=true`, then merge responses via `reduceTopNAggregator`
- **dquery/reduce_topn_aggregator.go** (new): A min-heap based aggregator that merges partial results from all data nodes by entity
﻿
### Analyzer
- **topn_analyzer.go**: Pass `groupByShard` flag for non-MIN/MAX aggregations; pass `emit_partial` to aggregation plan
- **measure_analyzer.go**: Update `newUnresolvedGroupBy` call with new `groupByShard` parameter
- **measure_plan_groupby.go**: Include `shard_id` in group key when `groupByShard=true`
﻿
### TopQueue
- **measure_top.go**: Support `n<=0` to collect all elements without truncation
﻿
### Tests
- Add TopN distributed test cases covering COUNT, SUM, MEAN, MIN, MAX aggregations


OmCheeLin:
The trade-off between TopN accuracy and the risk of OOM is indeed a concern. Could we introduce a Streaming Reduce mechanism to avoid loading all results into memory? By leveraging iterators or stream-based processing, we can aggregate data incrementally as it is received.

eye-gu:
> The trade-off between TopN accuracy and the risk of OOM is indeed a concern. Could we introduce a Streaming Reduce mechanism to avoid loading all results into memory? By leveraging iterators or stream-based processing, we can aggregate data incrementally as it is received.

During the discussion of the issue, it was confirmed that in terms of design, data for the same entity would not span across nodes (i.e., shard tags cannot be set arbitrarily and must be prefixed by entity tags). Therefore, there is no need to set topn=0 to obtain all results, and thus there is no risk of OOM.

PR #1066, opened 2026-04-13
Support multiple types for the same tag in a single trace or sidx part

ButterBright:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Related to apache/skywalking#13662.


banyand/internal/sidx/tag.go
```diff
@@ -28,6 +29,64 @@ import (
...
+
+func encodeTypedTag(name string, vt pbv1.ValueType) string {
+	suffix, ok := valueTypeToSuffix[vt]
+	if !ok || suffix == "" {
+		return name
+	}
+	return name + "." + suffix
+}
+
+func decodeTypedTag(key string) string {
```
hanahmily, line 64: This is unsafe for tag names that naturally end with a type suffix:                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                                                                  
  Tag name: "my.str"       → hasTypeSuffix returns TRUE (false positive)                                                                                                                                                                                                                                                                                                                                          
  Tag name: "arr.int"      → hasTypeSuffix returns TRUE (false positive)                                                                                                                                                                                                                                                                                                                                          
  Tag name: "data.bin"     → hasTypeSuffix returns TRUE (false positive)                                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                                                                  
  decodeTypedTag("my.str") would return "my" instead of "my.str". If a real tag is named my.str, it would be misinterpreted as a type-encoded version of tag my. This can cause silent data corruption during both merge and query.

banyand/internal/sidx/query_result.go
```diff
@@ -114,17 +114,23 @@ func (qr *queryResult) loadBlockData(tmpBlock *block, p *part, bm *blockMetadata
...
-		}
-
-		if !qr.loadTagData(tmpBlock, p, tagName, &tagBlockInfo, int(bm.count), decoder) {
-			// Continue loading other tags even if one fails
+		if exists {
+			qr.loadTagData(tmpBlock, p, tagName, &tagBlockInfo, int(bm.count), decoder)
 			continue
 		}
+		for typedTag, typedBlock := range bm.tagsBlocks {
+			if decodeTypedTag(typedTag) == tagName {
```
hanahmily: Map iteration order in Go is random. If both status.int and status.str exist, whichever is found first gets returned. Unlike the trace package (which checks schemaTagTypes), the sidx package has no schema-aware filtering — it returns whichever type variant the map iteration hits first. This could return wrong-typed data.

banyand/internal/sidx/query_result.go
```diff
@@ -114,17 +114,23 @@ func (qr *queryResult) loadBlockData(tmpBlock *block, p *part, bm *blockMetadata
...
+					tmpBlock.tags[tagName] = td
+					delete(tmpBlock.tags, typedTag)
+				}
+				break
+			}
+		}
 	}
-
-	return len(tmpBlock.userKeys) > 0
+	return true
```
hanahmily:
```suggestion
	return len(tmpBlock.userKeys) > 0
```

hanahmily (changes requested):

hanahmily (approved):

PR #1069, opened 2026-04-14
Validate ShardingKey against Entity tags

### LINKED ISSUE #13814: feat(banyandb/measure): add validation to enforce ShardingKey compatibility with Entity
## Background

`Measure` defines two fields that affect data routing:

- **`entity`** (`Entity.tag_names`) — determines the series and shard for a data point
- **`sharding_key`** (`ShardingKey.tag_names`) — intended to enhance TopN streaming performance by overriding shard routing

```protobuf
message Measure {
  Entity entity = 4;
  ShardingKey sharding_key = 8;
}
```

## Problem

Although `ShardingKey` was designed to augment TopN performance while preserving entity locality, there is no enforcement of this contract at the schema level. A user could configure:

```
entity.tag_names       = ["service_id"]
sharding_key.tag_names = ["instance_id"]
```

With this configuration, data points sharing the same `service_id` (same entity) but different `instance_id` values would be routed to different shards/nodes. Each node would then hold only a **partial view** of that entity, producing incorrect TopN aggregation results and breaking query correctness.

The rule that *the same entity must always map to the same node* is currently implicit and relies entirely on the caller (e.g., OAP server) following the convention correctly.

## Proposed Fix

Add a server-side validation step when a `Measure` schema is created or updated.

**Rule:** `ShardingKey.tag_names` must be a **superset** of `Entity.tag_names` — it may add extra tags for finer-grained routing, but it must include all entity tags to preserve locality.

**Validation pseudocode:**

```
if sharding_key is set:
    for each tag in entity.tag_names:
        assert tag ∈ sharding_key.tag_names,
            "ShardingKey must contain all Entity tags to guarantee entity locality"
```

## Acceptance Criteria

- [ ] Schema creation/update for `Measure` returns a validation error when `sharding_key.tag_names` does not contain all tags in `entity.tag_names`
- [ ] The error message clearly explains the rule and provides guidance
- [ ] Existing schemas that already satisfy the rule (e.g., OAP-generated ones) are unaffected
- [ ] Unit tests cover both valid and invalid configurations

## References

- Proto definition: `api/proto/banyandb/database/v1/schema.proto`

aliyasirnac:
### Fix missing validation between Measure.ShardingKey and Measure.Entity

`ShardingKey` must be a superset of `Entity` to preserve entity locality;
otherwise TopN results become incorrect. This PR enforces the rule at
schema validation time.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13814.
- [x] Update the `CHANGES` log.


ButterBright:
LGTM

ButterBright:
@aliyasirnac Please fix the ci failures caused by the new restriction.

aliyasirnac:
Hello, following your feedback, I reviewed and tested my changes which only add a schema-level validation ensuring `ShardingKey` contains all `Entity` tags. The `TestCollectWithPartialClosedSegments` test in the storage package fails both on my branch and on main, so the issue is pre-existing. Setting `SegmentIdleTimeout` to 300ms resolves it, pointing to a timing sensitivity in the segment lifecycle. 

Any suggestions on the right approach here? @ButterBright

hanahmily:
> Hello, following your feedback, I reviewed and tested my changes which only add a schema-level validation ensuring `ShardingKey` contains all `Entity` tags. The `TestCollectWithPartialClosedSegments` test in the storage package fails both on my branch and on main, so the issue is pre-existing. Setting `SegmentIdleTimeout` to 300ms resolves it, pointing to a timing sensitivity in the segment lifecycle.
> 
> Any suggestions on the right approach here? @ButterBright

It looks good to me

api/validate/validate.go
```diff
@@ -164,6 +164,18 @@ func Measure(measure *databasev1.Measure) error {
 	if measure.IndexMode && len(measure.Fields) > 0 {
 		return errors.New("index mode is enabled, but fields are not empty")
 	}
+
+	if measure.ShardingKey != nil && len(measure.ShardingKey.TagNames) > 0 {
+		shardingKeyTags := make(map[string]struct{}, len(measure.ShardingKey.TagNames))
+		for _, tag := range measure.ShardingKey.TagNames {
```
hanahmily: The sequence of tags is crucial here. For instance, if the entity is "A,B,C," the sharding key must be "A, B" or "A." Patterns like "C," "B, A," or others are incorrect.

hanahmily (changes requested):

ButterBright:
> Hello, following your feedback, I reviewed and tested my changes which only add a schema-level validation ensuring `ShardingKey` contains all `Entity` tags. The `TestCollectWithPartialClosedSegments` test in the storage package fails both on my branch and on main, so the issue is pre-existing. Setting `SegmentIdleTimeout` to 300ms resolves it, pointing to a timing sensitivity in the segment lifecycle.
> 
> Any suggestions on the right approach here? @ButterBright

Test data under /pkg/test/measure (e.g., service_instance_cpm_minute.json) should be updated to pass the validation checks.

api/validate/validate.go
```diff
@@ -164,6 +164,17 @@ func Measure(measure *databasev1.Measure) error {
...
 		return errors.New("index mode is enabled, but fields are not empty")
 	}
+
+	if measure.ShardingKey != nil && len(measure.ShardingKey.TagNames) > 0 {
+		if len(measure.ShardingKey.TagNames) > len(measure.Entity.TagNames) {
+			return errors.New("ShardingKey must be a prefix of Entity tags to guarantee entity locality")
+		}
+		for i, tag := range measure.ShardingKey.TagNames {
+			if measure.Entity.TagNames[i] != tag {
+				return errors.New("ShardingKey must be a prefix of Entity tags to guarantee entity locality")
```
hanahmily, line 174: The e2e failed. Could you print the` measure.ShardingKey.TagNames` and `measure.Entity.TagNames`. Then I can see the detail of the error.

aliyasirnac:
I investigated the CI failures. These are not caused by the validation changes in this PR, the same tests also fail on the main branch.

  The failures are infrastructure/timing issues in the distributed integration
  tests:
  - stat /tmp/.../discovery.yaml: no such file or directory
  - rpc error: code = Unavailable desc = connection refused
  - fail to locate test-trace-group/sw(1,0): no nodes available

  All measure schemas pass the new ShardingKey validation without issues.

api/validate/validate.go
aliyasirnac: I added debug logging as suggested. Here are the results the validation passes correctly, no errors are returned:                                                                                                                
{"level":"info","time":"2026-04-16T18:53:23+03:00","message":"Full ShardingKey.TagNames: [service_id entity_id]"}
{"level":"info","time":"2026-04-16T18:53:23+03:00","message":"Full Entity.TagNames: [service_id entity_id]"}

hanahmily: Based on that, the validation should pass when ShardingKey is identical to the Entity, not just a subset.

hanahmily:
> I investigated the CI failures. These are not caused by the validation changes in this PR, the same tests also fail on the main branch.
> 
> The failures are infrastructure/timing issues in the distributed integration tests:
> 
> * stat /tmp/.../discovery.yaml: no such file or directory
> * rpc error: code = Unavailable desc = connection refused
> * fail to locate test-trace-group/sw(1,0): no nodes available
> 
> All measure schemas pass the new ShardingKey validation without issues.

The e2e failed due to the validation. I have provided a suggestion. Please follow it to improve the validation logic.

hanahmily:
@aliyasirnac The OAP get the error:

```
2026-04-22 01:11:10,484 org.apache.skywalking.oap.server.starter.OAPServerBootstrap 64 [main] ERROR [] - fail to create schema endpoint_cpm
org.apache.skywalking.oap.server.library.module.ModuleStartException: fail to create schema endpoint_cpm
	at org.apache.skywalking.oap.server.storage.plugin.banyandb.BanyanDBStorageProvider.start(BanyanDBStorageProvider.java:243)
	at org.apache.skywalking.oap.server.library.module.BootstrapFlow.start(BootstrapFlow.java:46)
	at org.apache.skywalking.oap.server.library.module.ModuleManager.init(ModuleManager.java:75)
	at org.apache.skywalking.oap.server.starter.OAPServerBootstrap.start(OAPServerBootstrap.java:52)
	at org.apache.skywalking.oap.server.starter.OAPServerStartUp.main(OAPServerStartUp.java:23)
Caused by: org.apache.skywalking.oap.server.core.storage.StorageException: fail to create schema endpoint_cpm
	at org.apache.skywalking.oap.server.storage.plugin.banyandb.BanyanDBIndexInstaller.createTable(BanyanDBIndexInstaller.java:262)
	at org.apache.skywalking.oap.server.core.storage.model.ModelInstaller.whenCreating(ModelInstaller.java:66)
	at org.apache.skywalking.oap.server.core.storage.model.StorageModels.addModelListener(StorageModels.java:197)
	at org.apache.skywalking.oap.server.storage.plugin.banyandb.BanyanDBStorageProvider.start(BanyanDBStorageProvider.java:241)
	... 4 more
Caused by: org.apache.skywalking.library.banyandb.v1.client.grpc.exception.UnknownException: io.grpc.StatusRuntimeException: UNKNOWN: ShardingKey must be a prefix of Entity tags to guarantee entity locality
	at org.apache.skywalking.library.banyandb.v1.client.grpc.exception.BanyanDBApiExceptionFactory.createException(BanyanDBApiExceptionFactory.java:61)
	at org.apache.skywalking.library.banyandb.v1.client.grpc.exception.BanyanDBGrpcApiExceptionFactory.create(BanyanDBGrpcApiExceptionFactory.java:52)
	at org.apache.skywalking.library.banyandb.v1.client.grpc.exception.BanyanDBGrpcApiExceptionFactory.createException(BanyanDBGrpcApiExceptionFactory.java:40)
	at org.apache.skywalking.library.banyandb.v1.client.grpc.HandleExceptionsWith.callAndTranslateApiException(HandleExceptionsWith.java:47)
	at org.apache.skywalking.library.banyandb.v1.client.grpc.MetadataClient.execute(MetadataClient.java:97)
	at org.apache.skywalking.library.banyandb.v1.client.metadata.MeasureMetadataRegistry.create(MeasureMetadataRegistry.java:39)
	at org.apache.skywalking.library.banyandb.v1.client.BanyanDBClient.define(BanyanDBClient.java:395)
	at org.apache.skywalking.oap.server.storage.plugin.banyandb.BanyanDBIndexInstaller.createTable(BanyanDBIndexInstaller.java:224)
	... 7 more
Caused by: io.grpc.StatusRuntimeException: UNKNOWN: ShardingKey must be a prefix of Entity tags to guarantee entity locality
	at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:351)
	at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:332)
	at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:174)
	at org.apache.skywalking.banyandb.database.v1.MeasureRegistryServiceGrpc$MeasureRegistryServiceBlockingStub.create(MeasureRegistryServiceGrpc.java:428)
	at org.apache.skywalking.library.banyandb.v1.client.metadata.MeasureMetadataRegistry.lambda$create$0(MeasureMetadataRegistry.java:40)
	at org.apache.skywalking.library.banyandb.v1.client.grpc.HandleExceptionsWith.callAndTranslateApiException(HandleExceptionsWith.java:45)
	... 11 more
2026-04-22T01:11:10.487063473Z pool-1-thread-1 INFO Stopping configuration XmlConfiguration[location=/skywalking/config/log4j2.xml, lastModified=2026-04-22T01:08:59.742Z]...
2026-04-22T01:11:10.487307938Z pool-1-thread-1 INFO Configuration XmlConfiguration[location=/skywalking/config/log4j2.xml, lastModified=2026-04-22T01:08:59.742Z] stopped
```

In order to analyze the failed reason, you should append  the `measure.ShardingKey.TagNames` and `measure.Entity.TagNames` to the error message. You can run e2e test locally to debug the root cause.

aliyasirnac:
I ran into a Docker-related issue while trying to run the E2E tests locally. I've been quite busy with work over the past few days, but I plan to dedicate some time this weekend to resolve it.

ButterBright:
[Here](https://skywalking.apache.org/docs/skywalking-banyandb/next/contribute/debugging/) is a guide about debugging e2e tests locally that might help. Feel free to share the errors if you encounter any issues.

aliyasirnac:
> [Here](https://skywalking.apache.org/docs/skywalking-banyandb/next/contribute/debugging/) is a guide about debugging e2e tests locally that might help. Feel free to share the errors if you encounter any issues.

Thanks..

aliyasirnac:
Got the e2e working locally and tracked down the root cause. Legacy schemas like `endpoint_cpm` have a ShardingKey that isn't a prefix of their Entity tags, so BanyanDB rejects them on startup and OAP can't boot.

Instead of hard-failing, I changed the validation to log a warning instead of returning an error. Schemas register fine, OAP starts up, and operators still see the bad locality in the logs. Let me know what you think.

hanahmily:
> Got the e2e working locally and tracked down the root cause. Legacy schemas like `endpoint_cpm` have a ShardingKey that isn't a prefix of their Entity tags, so BanyanDB rejects them on startup and OAP can't boot.
> 
> Instead of hard-failing, I changed the validation to log a warning instead of returning an error. Schemas register fine, OAP starts up, and operators still see the bad locality in the logs. Let me know what you think.

Could you show me the ShardingKey and Entity for `endpoint_cpm`? We might enhance the validation logic.

hanahmily:
Tested this PR against the standard OAP compose (`test/e2e-v2/cases/storage/banyandb/`, OAP @ `b4a8811d`). The advisory check fires **174 times across 87 unique measures** at OAP startup — every one of them with the same mismatch:

```
ShardingKey [service_id] must be a prefix of Entity tags [entity_id]
```

The affected measures are the entire endpoint-scoped, browser-page, service-relation, instance-relation, and kubernetes-endpoint metric families. The schema OAP installs:

```yaml
entity:      { tagNames: [entity_id] }
shardingKey: { tagNames: [service_id] }
```

This is OAP's intentional composite-id pattern: `entity_id` is a base64 that already contains `service_id`, and `service_id` is exposed separately so service-level TopN aggregation is correct. The current "prefix of Entity tags" rule can't express this — it compares tag names literally.

Two other issues with the rule as written:

1. The PR description says "ShardingKey must be a *superset* of Entity," but the code checks for a *prefix*. The real invariant for entity locality is `ShardingKey ⊆ Entity` (set-wise).
2. The test case `invalid sharding key (subset with different tag)` (Entity=`[s,i]`, ShardingKey=`[i]`) is rejected, but locality is fine here.

### Proposal

```go
// (a) Skip when len(Entity.TagNames) == 1 — composite-id pattern (OAP's entity_id).
// (b) Otherwise require ShardingKey ⊆ Entity AND shared tags in the same relative order.
```

(a) eliminates the 87 false positives; (b) catches reorder bugs *and* the superset case (ShardingKey tag absent from Entity), both of which the current rule either misses or over-rejects.

aliyasirnac:
Thanks for the feedback. I'll rewrite the code.

27 Nis 2026 Pzt 16:10 tarihinde Gao Hongtao ***@***.***> şunu
yazdı:

hanahmily:
@aliyasirnac any updates?

aliyasirnac:
I’ll be making a few updates today; I’ll commit the changes once I’ve
confirmed that the end-to-end test is working.

3 May 2026 Paz 07:42 tarihinde Gao Hongtao ***@***.***> şunu
yazdı:

api/validate/validate.go
```diff
@@ -175,9 +175,41 @@ func Measure(measure *databasev1.Measure) error {
...
+		}
+		if pos <= prevPos {
+			return fmt.Errorf("ShardingKey %v is not in the same relative order as Entity tags %v",
+				measure.ShardingKey.TagNames, measure.Entity.TagNames)
+		}
+		prevPos = pos
+	}
+	return nil
+}
+
```
aliyasirnac: Have you had a chance to look at this code?

hanahmily (approved):

PR #1084, opened 2026-04-22
feat: implement panic diagnostics and FODC crash reporting pipeline

Fine0830:
## Summary

This PR adds an end-to-end panic diagnostics pipeline for BanyanDB and FODC, covering runtime crash output, recovered panic artifacts, breadcrumb/state capture, agent-side crash collection, and proxy-side diagnostics aggregation.

The goal is to make production failures easier to inspect while keeping the runtime path safe and low-overhead during normal operation.

### Panic diagnostics foundation
- add `pkg/panicdiag` as the shared panic diagnostics package
- add recovered panic artifact writing (`panic.json`, `crash.txt`)
- add process-wide panic metrics and reporter hooks
- add runtime crash output bootstrap via `runtime/debug.SetCrashOutput`
- add bounded state dumping and breadcrumb support
- add gRPC breadcrumb interceptors

### BanyanDB integration
- wire panic diagnostics into BanyanDB startup
- expose panic artifact directory configuration through observability/service flags
- extend recovery coverage to more production goroutines and run-group helpers

### FODC agent
- add in-process panic store for recovered panics
- add filesystem watcher and crash analyzer for persisted crash artifacts
- add combined collection provider to merge in-process and filesystem-backed diagnostics
- add diagnostics endpoint exposure from the agent HTTP server
- add cleanup for empty `runtime-crash-*.txt` files on normal exit
- preserve non-empty runtime crash output files for postmortem analysis

### FODC proxy
- add diagnostics aggregation across agents
- add proxy-side diagnostics HTTP/API endpoint
- add metrics/registration flow updates needed for diagnostics collection

### Test coverage and hardening
- add unit and integration coverage for:
  - panic artifact persistence
  - breadcrumb/state-dump capture
  - crash collection and deduplication
  - agent/proxy diagnostics flow
  - Prometheus export and FODC integration stability
- fix related lint issues and flaky test behavior uncovered while adding the diagnostics pipeline


pkg/panicdiag/artifact.go
```diff
@@ -0,0 +1,144 @@
...
+	artifactDir := filepath.Join(aw.rootDir, aw.dirName(record.Component, recordedAt))
+	if err := os.MkdirAll(artifactDir, 0o755); err != nil {
+		return "", fmt.Errorf("create artifact dir: %w", err)
+	}
+
+	summaryPath := filepath.Join(artifactDir, crashTextFileName)
+	summary := buildCrashSummary(record)
+	if writeErr := os.WriteFile(summaryPath, []byte(summary), 0o600); writeErr != nil {
+		return "", fmt.Errorf("write crash summary: %w", writeErr)
+	}
```
hanahmily: **C1: `panic.json` is never written; breadcrumbs / metadata / state-dump status are lost on disk**

The constant `panicRecordFileName = "panic.json"` is declared at line 29 but `Write` only persists `crash.txt` here. `buildCrashSummary` (lines 142-156) emits only `OccurredAt / Component / Recovered / Panic / Stack` — **no `Breadcrumbs`, no `ProcessMetadata`, no `StateDump` status**.

The reader in `collections.go` (`readCollection`) is already coded to prefer `panic.json` and silently fall back to text-parsing `crash.txt` when it's absent. Today that fallback is always taken, so every postmortem read of an artifact — the actual production scenario, where the process that panicked may already be gone — produces a `PanicRecord` with `Breadcrumbs == nil`. The headline benefit of the design (breadcrumbs survive into the artifact) does not hold.

**Suggested fix** — persist the structured record first (source of truth), then the human-readable summary as best-effort:

```go
// 1) Structured record — the source of truth for postmortem readers.
recordPath := filepath.Join(artifactDir, panicRecordFileName)
recordBytes, marshalErr := json.MarshalIndent(record, "", "  ")
if marshalErr != nil {
    return "", fmt.Errorf("marshal panic record: %w", marshalErr)
}
if writeErr := os.WriteFile(recordPath, recordBytes, 0o600); writeErr != nil {
    return "", fmt.Errorf("write panic record: %w", writeErr)
}

// 2) Human-readable summary. Best-effort — don't fail the artifact if this errors.
summaryPath := filepath.Join(artifactDir, crashTextFileName)
if writeErr := os.WriteFile(summaryPath, []byte(buildCrashSummary(record)), 0o600); writeErr != nil {
    return artifactDir, fmt.Errorf("write crash summary: %w", writeErr)
}
```

Two follow-ups in adjacent files:
- `collections.go` `readCollection`: prefer `panic.json` when present and treat unmarshal failure as a hard error rather than silently falling back to the text parser.
- `collections.go` `PruneArtifacts`: recognize a directory by `panic.json` **or** `crash.txt`, so a partial write doesn't make a directory invisible to LRU eviction.

A test that proves the fix and would have caught the regression:

```go
func TestArtifactWriter_PersistsBreadcrumbs(t *testing.T) {
    dir := t.TempDir()
    aw := NewArtifactWriter(dir)
    rec := &PanicRecord{
        Component:       "test",
        PanicValue:      "boom",
        Breadcrumbs:     []Breadcrumb{{Stage: "handle", Component: "grpc"}},
        ProcessMetadata: map[string]string{"pod": "p1"},
    }
    out, err := aw.Write(rec)
    require.NoError(t, err)

    cols, err := ListCollections(dir)
    require.NoError(t, err)
    require.Len(t, cols, 1)
    require.Equal(t, filepath.Base(out), cols[0].ArtifactDir)
    require.Equal(t, "handle", cols[0].Record.Breadcrumbs[0].Stage)
    require.Equal(t, "p1", cols[0].Record.ProcessMetadata["pod"])
}
```

hanahmily: **C1: `panic.json` is never written; breadcrumbs / metadata / state-dump status are lost on disk**

The constant `panicRecordFileName = "panic.json"` is declared at line 29 but `Write` only persists `crash.txt` here. `buildCrashSummary` (lines 142-156) emits only `OccurredAt / Component / Recovered / Panic / Stack` — **no `Breadcrumbs`, no `ProcessMetadata`, no `StateDump` status**.

The reader in `collections.go` (`readCollection`) is already coded to prefer `panic.json` and silently fall back to text-parsing `crash.txt` when it's absent. Today that fallback is always taken, so every postmortem read of an artifact — the actual production scenario, where the process that panicked may already be gone — produces a `PanicRecord` with `Breadcrumbs == nil`. The headline benefit of the design (breadcrumbs survive into the artifact) does not hold.

**Suggested fix** — persist the structured record first (source of truth), then the human-readable summary as best-effort:

```go
// 1) Structured record — the source of truth for postmortem readers.
recordPath := filepath.Join(artifactDir, panicRecordFileName)
recordBytes, marshalErr := json.MarshalIndent(record, "", "  ")
if marshalErr != nil {
    return "", fmt.Errorf("marshal panic record: %w", marshalErr)
}
if writeErr := os.WriteFile(recordPath, recordBytes, 0o600); writeErr != nil {
    return "", fmt.Errorf("write panic record: %w", writeErr)
}

// 2) Human-readable summary. Best-effort — don't fail the artifact if this errors.
summaryPath := filepath.Join(artifactDir, crashTextFileName)
if writeErr := os.WriteFile(summaryPath, []byte(buildCrashSummary(record)), 0o600); writeErr != nil {
    return artifactDir, fmt.Errorf("write crash summary: %w", writeErr)
}
```

Two follow-ups in adjacent files:
- `collections.go` `readCollection`: prefer `panic.json` when present and treat unmarshal failure as a hard error rather than silently falling back to the text parser.
- `collections.go` `PruneArtifacts`: recognize a directory by `panic.json` **or** `crash.txt`, so a partial write doesn't make a directory invisible to LRU eviction.

A test that proves the fix and would have caught the regression:

```go
func TestArtifactWriter_PersistsBreadcrumbs(t *testing.T) {
    dir := t.TempDir()
    aw := NewArtifactWriter(dir)
    rec := &PanicRecord{
        Component:       "test",
        PanicValue:      "boom",
        Breadcrumbs:     []Breadcrumb{{Stage: "handle", Component: "grpc"}},
        ProcessMetadata: map[string]string{"pod": "p1"},
    }
    out, err := aw.Write(rec)
    require.NoError(t, err)

    cols, err := ListCollections(dir)
    require.NoError(t, err)
    require.Len(t, cols, 1)
    require.Equal(t, filepath.Base(out), cols[0].ArtifactDir)
    require.Equal(t, "handle", cols[0].Record.Breadcrumbs[0].Stage)
    require.Equal(t, "p1", cols[0].Record.ProcessMetadata["pod"])
}
```

pkg/panicdiag/breadcrumb.go
```diff
@@ -0,0 +1,145 @@
...
+// gRPC recovery interceptors — which capture the context at interceptor entry,
+// before the handler runs — can still read breadcrumbs appended inside the handler.
+type mutableBreadcrumbContextKey struct{}
+
+// mutableBreadcrumbStore holds breadcrumbs that can be appended without creating
+// a new context value. A pointer to this struct is stored in the context so that
+// mutation is visible through the original context reference.
+type mutableBreadcrumbStore struct {
+	breadcrumbs []Breadcrumb
+}
```
hanahmily, line 35: **C2: `mutableBreadcrumbStore` is unsynchronized — concurrent `WithBreadcrumb` calls race**

The slice is mutated by `WithBreadcrumb` (append at line 81) and read by `BreadcrumbsFromContext` (`copy` at line 116) without any lock. Once a gRPC interceptor (`grpcinterceptors.go:24-33`) installs the store, the same `*mutableBreadcrumbStore` is reachable from every goroutine that derives a context from the request — including the many `run.Go` workers spawned in handlers throughout this PR. Concurrent `append` tears the slice header; `go test -race` will fire as soon as any test exercises more than one goroutine on a shared context.

The reader runs from inside the panic-recovery defer (`recovery.go:50`). A torn slice there can re-panic and bypass the artifact write entirely — neither breadcrumbs nor the recover-and-log path land.

**Suggested fix** — add a `sync.RWMutex` to the store and use it on both paths. Hot only on panic; on the happy path the cost is one uncontended `Lock/Unlock` per breadcrumb (negligible vs. the existing per-call `cloneStringMap`).

```go
type mutableBreadcrumbStore struct {
    mu          sync.RWMutex
    breadcrumbs []Breadcrumb
}
```

In `WithBreadcrumb`:

```go
if store, ok := ctx.Value(mutableBreadcrumbContextKey{}).(*mutableBreadcrumbStore); ok {
    store.mu.Lock()
    if len(store.breadcrumbs) < maxBreadcrumbDepth {
        store.breadcrumbs = append(store.breadcrumbs, Breadcrumb{
            Time:      nowBreadcrumbTime(),
            Stage:     stage,
            Component: component,
            Fields:    cloneStringMap(fields),
        })
    }
    store.mu.Unlock()
    return ctx
}
```

In `BreadcrumbsFromContext`:

```go
if store, ok := ctx.Value(mutableBreadcrumbContextKey{}).(*mutableBreadcrumbStore); ok {
    store.mu.RLock()
    defer store.mu.RUnlock()
    if len(store.breadcrumbs) == 0 { return nil }
    result := make([]Breadcrumb, len(store.breadcrumbs))
    copy(result, store.breadcrumbs)
    for i := range result {
        result[i].Fields = cloneStringMap(result[i].Fields)
    }
    return result
}
```

A focused race test (must fail on the current code under `-race`):

```go
func TestMutableBreadcrumbs_ConcurrentAppendSafe(t *testing.T) {
    ctx := WithMutableBreadcrumbs(context.Background())
    var wg sync.WaitGroup
    for i := 0; i < 16; i++ {
        wg.Add(1)
        go func(i int) {
            defer wg.Done()
            for j := 0; j < 100; j++ {
                WithBreadcrumb(ctx, "stage", "component", map[string]string{"i": strconv.Itoa(i)})
                _ = BreadcrumbsFromContext(ctx)
            }
        }(i)
    }
    wg.Wait()
    if got := BreadcrumbsFromContext(ctx); len(got) > maxBreadcrumbDepth {
        t.Fatalf("breadcrumb depth exceeded cap: %d", len(got))
    }
}
```

pkg/run/goroutine.go
```diff
@@ -0,0 +1,39 @@
...
+// When a panic counter is configured via RecoveryOptions.Counter or
+// panicdiag.SetDefaultPanicCounter, banyandb_panic_total is incremented with
+// component as the label.
+func Go(ctx context.Context, component string, log *logger.Logger, fn func(context.Context)) {
+	panicdiag.GoWithRecovery(ctx, panicdiag.RecoveryOptions{
+		Component: component,
+		Logger:    log,
+	}, nil, func(ctxPtr *context.Context) {
+		fn(*ctxPtr)
+	})
```
hanahmily: **C3: `run.Go` discards the ctx pointer — breadcrumbs added inside `fn` never reach the recovery defer**

`WithRecovery` reads breadcrumbs from the outer `ctx` variable in its defer (`pkg/panicdiag/recovery.go:50`). The whole point of `fn func(*context.Context)` is so the body can do `*ctxPtr = panicdiag.WithBreadcrumb(*ctxPtr, ...)` and have the defer see the new chain.

This wrapper dereferences once and passes a value-copied ctx to `fn`:

```go
}, nil, func(ctxPtr *context.Context) {
    fn(*ctxPtr)
})
```

Any `WithBreadcrumb(innerCtx, ...)` call inside `fn` builds a chain that lives only on `fn`'s stack frame; the outer `ctx` never updates; the defer reads zero breadcrumbs. The only existing escape hatch is the `mutableBreadcrumbStore` installed by gRPC interceptors — but plain `run.Go` callers (which are most of the new wrapping in this PR: `banyand/queue/pub/*`, `banyand/measure/topn.go`, `banyand/internal/storage/*`, `banyand/trace/streaming_pipeline.go`, etc.) get neither.

Combined with C1 (`panic.json` never written), this means breadcrumbs are dropped on both the in-process and on-disk paths for almost every panic site this PR adds. The headline feature is null on the way out for these call sites.

**Suggested fix (Option A) — install a mutable store inside `run.Go` automatically.** One-line change, no caller-API churn:

```go
func Go(ctx context.Context, component string, log *logger.Logger, fn func(context.Context)) {
    panicdiag.GoWithRecovery(ctx, panicdiag.RecoveryOptions{
        Component: component,
        Logger:    log,
    }, nil, func(ctxPtr *context.Context) {
        // Install a mutable breadcrumb store so breadcrumbs added inside
        // fn are visible to the recovery defer (which reads *ctxPtr after
        // fn returns or panics).
        *ctxPtr = panicdiag.WithMutableBreadcrumbs(*ctxPtr)
        fn(*ctxPtr)
    })
}
```

This relies on C2 being fixed as well — without the mutex, `run.Go` workers that share a parent's mutable store will race.

A test that proves the fix and would have caught the regression:

```go
func TestGo_RecoveryCapturesBreadcrumbsAddedInFn(t *testing.T) {
    var capturedRecord *panicdiag.PanicRecord
    panicdiag.SetDefaultReporter(func(_ context.Context, r panicdiag.RecoveryResult) {
        capturedRecord = r.Record
    })
    t.Cleanup(func() { panicdiag.SetDefaultReporter(nil) })

    done := make(chan struct{})
    run.Go(context.Background(), "test", logger.GetLogger("test"),
        func(ctx context.Context) {
            defer close(done)
            panicdiag.WithBreadcrumb(ctx, "stage-A", "component", nil)
            panicdiag.WithBreadcrumb(ctx, "stage-B", "component", nil)
            panic("boom")
        })
    <-done
    require.NotNil(t, capturedRecord)
    require.Len(t, capturedRecord.Breadcrumbs, 2)
    require.Equal(t, "stage-A", capturedRecord.Breadcrumbs[0].Stage)
    require.Equal(t, "stage-B", capturedRecord.Breadcrumbs[1].Stage)
}
```

hanahmily (changes requested):

hanahmily (commented):
## Round 2 review — design issue in `run.Go` / `panicdiag.WithRecovery`

The three new criticals from this round (NEW-C1 in `pkg/timestamp/scheduler.go`, NEW-C2 in `pkg/schema/cache.go`, NEW-C3 in `pkg/run/run.go`) and several round-1 highs (H2, H4, H6) all share a root cause: **the `run.Go` / `WithRecovery` surface is one-size-fits-all and one-way.** Once a panic enters the recovery defer, there is no way to signal the parent, fail the lifecycle, or re-raise.

### How `run.Go` handles re-raising a panic — the short answer: it doesn't

```go
// pkg/run/goroutine.go
func Go(ctx, component, log, fn) {
    panicdiag.GoWithRecovery(ctx, RecoveryOptions{...}, nil /* reporter */, func(ctxPtr) {
        *ctxPtr = WithMutableBreadcrumbs(*ctxPtr)
        fn(*ctxPtr)
    })
}

// pkg/panicdiag/recovery.go
defer func() {
    panicValue := recover()
    if panicValue == nil { return }
    // write artifact, log, callReporter
}()  // <-- defer ends here. Always swallows. No re-panic.
```

Every panic that enters this defer is swallowed. The recovered goroutine returns normally; the parent has no idea anything went wrong unless it reads a metric counter or wires a custom `Reporter` — and `run.Go` hard-codes `reporter = nil`, so even that escape hatch is unreachable through the public API.

### Design problems this creates

| # | Issue | Manifestations |
|---|---|---|
| 1 | **No re-panic variant.** Round 1 explicitly asked for `run.GoOrDie` that re-panics after writing the artifact. Not added. | H2 (corrupt-state continuation), NEW-C3 |
| 2 | **No synchronous failure channel to the parent.** The parent goroutine cannot observe whether the child panicked. To do so it must invent its own out-of-band signal (channel, metric, atomic). | NEW-C1: scheduler can't tell its action panicked, so it stalls 5 minutes waiting on `resultCh` |
| 3 | **`run.Go` hides the `Reporter` parameter.** The underlying `GoWithRecovery` accepts a `Reporter`, but `run.Go` always passes `nil`. Callers who need notification must abandon `run.Go` and call `GoWithRecovery` directly — duplicating the `WithMutableBreadcrumbs` boilerplate. | All three NEW-Cx |
| 4 | **Recovery is uniformly applied, with no per-call-site policy.** Background helpers and write paths recover identically, even though "log + continue" is wrong for the latter. | H2 from round 1, still unaddressed |
| 5 | **Asymmetric lifecycle protection in `Group.Run`.** `Serve` is wrapped; `GracefulStop` is not. Recovering only one half of the lifecycle gives a false sense of safety. | NEW-C3 |
| 6 | **The `Reporter` runs inside the recovery defer, with no second-tier safety net.** A panic inside the Reporter (or inside the artifact write — H4) propagates as a fresh panic with no further recovery. | Latent; trips when a Reporter touches optional state |
| 7 | **`WithMutableBreadcrumbs` is idempotent.** Two `run.Go` calls on the same parent ctx share a mutable store. Breadcrumbs from goroutine A appear in goroutine B's panic record. The C2 mutex is now load-bearing for correctness, not just for race-freedom. | Cross-goroutine breadcrumb leakage in cluster code that forks workers off a shared parent ctx |
| 8 | **No compile-time discipline against raw `go func()`.** The PR converted dozens of sites but missed `pkg/schema/cache.go:268`. | NEW-C2 |
| 9 | **Artifact write is synchronous inside the recover path.** A stuck disk hangs the recovering goroutine while it still holds whatever locks `fn` had at panic time. | H4 from round 1, unaddressed |

### What a sound design would look like

Three call shapes for three policies, so the caller has to pick:

```go
// (a) Background helper — recover, log, continue. Today's run.Go.
run.Go(ctx, "cache-cleaner", log, fn)

// (b) Critical path — recover for diagnostics, then re-panic so the runtime
//     handles process exit / lifecycle abort.
run.GoOrDie(ctx, "block-writer", log, fn)

// (c) Caller-managed — return panic value so the caller decides.
//     Used by scheduler, Group.Run interrupt, anywhere that needs a synchronous signal.
panicCh := run.GoWithSignal(ctx, "scheduler-action", log, fn)
select {
case r := <-panicCh:
    if r.Panicked { /* increment counter, advance task */ }
case <-timeout:
    /* … */
}
```

Plus:

- `Group.Run` should wrap **both** `Serve` and `GracefulStop` (and `PreRun`) in recovery.
- `WithMutableBreadcrumbs` called from `run.Go` should always allocate a fresh store, so per-goroutine breadcrumb sets don't bleed across goroutines that share a parent ctx.
- The artifact write should be deferred to a bounded async sink (H4 from round 1).

### Bottom line

The current design treats panic recovery as a leaf-level concern (one wrapper, one policy, one return path). In reality it's a **lifecycle concern**: callers need to choose between "recover and continue", "recover and abort", and "recover and report". `run.Go` only offers the first, so the other two get re-implemented incorrectly at every call site that needs them. The three new criticals are not coincidences — they are the same missing API showing up in three different shapes.

Without an API change, the next regression that looks like NEW-C1 is just a matter of time.

hanahmily (approved):

PR #1085, opened 2026-04-22
fix(topn): deduplicate entities in TopN aggregation query

### LINKED ISSUE #13837: [Bug] BanyanDB TopN Aggregation query returns incorrect results due to missing entity deduplication
### Search before asking

### What happened

BanyanDB TopN Aggregation queries have two issues in distributed and standalone modes:

1. Distributed mode: DataNode returns results without entity deduplication. The DataNode sorts all
time-bucket data points by value and returns the top N entries directly. If one entity's value dominates
across multiple time buckets, only that entity's data points will fill all Top N slots, preventing other
entities from appearing in the result at all.
2. Standalone mode query semantics differ from distributed mode. Standalone mode performs GROUP BY +
aggregation (e.g., MAX/SUM) before selecting Top N, while distributed mode sorts by raw value without entity
 deduplication. This causes inconsistent query results between the two deployment modes.

Both modes should unify to the following query semantics:

SELECT entity, MAX(value) as val FROM data GROUP BY entity ORDER BY val DESC LIMIT N

That is:
- Deduplicate by entity first, keeping the best value for each entity (MAX for DESC, MIN for ASC)
- Sort by value and return the top N distinct entities
- Then apply the agg aggregation function (sum/mean/count, etc.)

### How to reproduce

1. Deploy a BanyanDB distributed cluster (Liaison + DataNode)
2. Write measure data for a TopN Aggregation where one entity consistently has a much larger value than
others across multiple time buckets
3. Query TopN and observe that only the dominant entity occupies all Top N positions, with no other entities
 returned
4. Compare with Standalone mode query results — the two modes return different results

### Anything else

_No response_

### Code of Conduct

eye-gu:

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13837.

### DIFF
--- a/banyand/query/processor_topn.go
+++ b/banyand/query/processor_topn.go
@@ -219,7 +219,7 @@ func (t *topNQueryProcessor) Rev(ctx context.Context, message bus.Message) (resp
 		}
 		resp = bus.NewMessage(bus.MessageID(now), &measurev1.TopNResponse{Lists: lists})
 	} else {
-		// Distribute DataNode mode: return top N distinct entities directly
+		// Standalone mode without aggregation or distributed DataNode mode: return top N distinct entities directly
 		resp = bus.NewMessage(bus.MessageID(now), toTopNResponse(result))
 	}
 	if !request.Trace && t.slowQuery > 0 {
--- a/banyand/query/processor_topn.go
+++ b/banyand/query/processor_topn.go
@@ -32,6 +32,7 @@ import (
 	"github.com/apache/skywalking-banyandb/banyand/measure"
 	"github.com/apache/skywalking-banyandb/pkg/bus"
 	"github.com/apache/skywalking-banyandb/pkg/logger"
+	pbv1 "github.com/apache/skywalking-banyandb/pkg/pb/v1"
 	"github.com/apache/skywalking-banyandb/pkg/query"
 	"github.com/apache/skywalking-banyandb/pkg/query/executor"
 	logical_measure "github.com/apache/skywalking-banyandb/pkg/query/logical/measure"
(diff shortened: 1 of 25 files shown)


pkg/query/logical/measure/topn_plan_distinct.go
```diff
@@ -0,0 +1,271 @@
...
+
+func (h *entityDedupTopN) tryAddToHeap(key string, val int64, idp *measurev1.InternalDataPoint) {
+	if h.topN <= 0 {
+		return
+	}
+	if len(h.items) < h.topN {
+		heap.Push(h, &dedupHeapItem{key: key, val: val, idp: idp})
+		return
+	}
+	if len(h.items) == 0 {
```
hanahmily: we already know topN > 0 (first check passed) and len(h.items) >= topN (second check didn't return). Therefore len(h.items) > 0 is guaranteed. This branch is dead code.

banyand/query/processor_topn.go
```diff
@@ -181,7 +182,46 @@ func (t *topNQueryProcessor) Rev(ctx context.Context, message bus.Message) (resp
...
+		}
+		lists, err := aggregator.Val(tags)
+		if err != nil {
+			ml.Error().Err(err).RawJSON("req", logger.Proto(request)).Msg("fail to aggregate topN results")
+			resp = bus.NewMessage(bus.MessageID(now), common.NewError("fail to aggregate topN results for measure %s: %v", request.Name, err))
+			return
+		}
+		resp = bus.NewMessage(bus.MessageID(now), &measurev1.TopNResponse{Lists: lists})
+	} else {
+		// Distribute DataNode mode: return top N distinct entities directly
```
hanahmily: This else branch is also taken in standalone mode without aggregation, not only in distributed data node mode. The comment should reflect both cases.

pkg/query/logical/measure/topn_analyzer.go
```diff
@@ -109,23 +109,7 @@ func TopNAnalyze(criteria *measurev1.TopNRequest, sourceMeasureSchemaList []*dat
...
-			false,
-			false)
-	}
-
-	plan = top(plan, &measurev1.QueryRequest_Top{
-		Number:         criteria.GetTopN(),
-		FieldName:      topNAggSchema.FieldName,
-		FieldValueSort: criteria.GetFieldValueSort(),
-	})
+	plan = topNDistinct(plan, criteria.GetTopN(), criteria.GetFieldValueSort(), topNAggSchema.FieldName)
```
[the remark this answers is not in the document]
eye-gu: The design is such that it ensures the query semantics remain the same

hanahmily (changes requested):

banyand/query/processor_topn.go
eye-gu: Done.

pkg/query/logical/measure/topn_plan_distinct.go
eye-gu: Done.

hanahmily (approved):

PR #1093, opened 2026-04-26
fix: Use `topic` instead of `session_id` as the Prometheus label

OmCheeLin:
### Use `topic` instead of `session_id` as the Prometheus label on liaison `queue_sub` chunk-ordering counters to avoid unbounded metric cardinality.
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Part of apache/skywalking#13775.


banyand/queue/sub/chunked_sync.go
```diff
@@ -266,7 +268,9 @@ func (s *server) processChunkWithReordering(stream clusterv1.ChunkedSyncService_
 
 	if req.ChunkIndex > buffer.expectedIndex {
 		gap := req.ChunkIndex - buffer.expectedIndex
-		s.updateChunkOrderMetrics("out_of_order_received", req.SessionId)
+		if session.metadata != nil {
+			s.updateChunkOrderMetrics("out_of_order_received", session.metadata.Topic)
+		}
```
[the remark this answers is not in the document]
OmCheeLin: I think this might not need to be changed, since the client is usually internal gRPC within the cluster and is considered trusted.

banyand/queue/sub/server.go
```diff
@@ -435,32 +435,32 @@ func newMetrics(factory observability.Factory) *metrics {
...
 	case "out_of_order_received":
-		s.metrics.outOfOrderChunksReceived.Inc(1, sessionID)
+		s.metrics.outOfOrderChunksReceived.Inc(1, topic)
 	case "chunk_buffered":
-		s.metrics.chunksBuffered.Inc(1, sessionID)
+		s.metrics.chunksBuffered.Inc(1, topic)
 	case "buffer_timeout":
-		s.metrics.bufferTimeouts.Inc(1, sessionID)
+		s.metrics.bufferTimeouts.Inc(1, topic)
 	case "gap_too_large":
```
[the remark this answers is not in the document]
OmCheeLin: This issue already existed before and was not introduced by this PR. I will address and improve it in a follow-up PR.

docs/operation/grafana-cluster.json
```diff
@@ -2524,6 +2524,185 @@
...
+          "legendFormat": "buffer_timeout {{topic}}",
+          "range": true,
+          "refId": "C"
+        },
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${DS_PROMETHEUS}"
+          },
+          "editorMode": "code",
```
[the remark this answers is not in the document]
OmCheeLin: I’ve reverted the doc-related changes. I’ll handle the documentation in the final PR.

banyand/queue/sub/chunked_sync.go
```diff
@@ -159,7 +159,9 @@ func (s *server) SyncPart(stream clusterv1.ChunkedSyncService_SyncPartServer) er
 				if currentSession.partCtx != nil {
 					if currentSession.partCtx.Handler != nil {
 						if finishErr := currentSession.partCtx.Handler.FinishSync(); finishErr != nil {
-							s.updateChunkOrderMetrics("finish_sync_err", currentSession.sessionID)
+							if currentSession.metadata != nil {
```
hanahmily: **Issue 1 — defensive `metadata != nil` guard is dead code that silently drops metrics if the invariant ever breaks.**

This guard (and the four equivalents added at lines 269, 281, 292, 304 in `processChunkWithReordering`) is unreachable today: a `syncSession` is only constructed at line 173 inside the `if req.GetMetadata() != nil` branch (line 157), so `currentSession.metadata` is never nil after creation.

The risk is in the *future*: if a refactor ever lets `metadata` become nil, this guard silently swallows exactly the chunk-ordering metrics we use to detect bad sync state — no log line, no series, just a missing signal for `finish_sync_err` / `gap_too_large` / `buffer_full`.

Suggest one of:
- Drop the guard and rely on the documented invariant.
- Or, fall back to a sentinel topic (`"unknown"`) and emit `s.log.Warn(...)` so any breach surfaces.

banyand/queue/sub/server.go
```diff
@@ -435,32 +435,32 @@ func newMetrics(factory observability.Factory) *metrics {
...
 		return // Skip metrics if not initialized (e.g., during tests)
 	}
 	switch event {
 	case "out_of_order_received":
-		s.metrics.outOfOrderChunksReceived.Inc(1, sessionID)
+		s.metrics.outOfOrderChunksReceived.Inc(1, topic)
 	case "chunk_buffered":
-		s.metrics.chunksBuffered.Inc(1, sessionID)
+		s.metrics.chunksBuffered.Inc(1, topic)
 	case "buffer_timeout":
```
hanahmily, line 457: **Issue 2 — `"buffer_timeout"` event is wired here but never fired anywhere.**

```
$ grep -rn '"buffer_timeout"' --include='*.go' .
banyand/queue/sub/server.go:457   ← only this case statement
```

`checkBufferTimeout` (chunked_sync.go:434) is the natural producer but returns the error without recording the metric. Worse, `checkBufferTimeout` itself only has a *test* caller (`chunked_sync_test.go:280`); no production code path appears to invoke it.

Since this PR is exactly about cleaning up `queue_sub` chunk-ordering observability, please pick one:
- **Wire it up:** add `s.updateChunkOrderMetrics("buffer_timeout", session.metadata.Topic)` inside `checkBufferTimeout`, *and* invoke `checkBufferTimeout` from a real code path (e.g. `processChunkWithReordering` or a periodic janitor).
- **Delete the dead code:** remove the `bufferTimeouts` field, the `factory.NewCounter("buffer_timeouts", …)` registration, and this `case` branch.

Leaving it half-wired ships a `buffer_timeouts` metric to operators that is silently always zero.

CHANGES.md
```diff
@@ -34,6 +34,7 @@ Release Notes.
 
 ### Bug Fixes
 
+- Use `topic` instead of `session_id` as the Prometheus label on liaison `queue_sub` chunk-ordering counters to avoid unbounded metric cardinality.
```
hanahmily, line 37: **Issue 3 — docs and Grafana panel were ripped out by the `revert doc` commit (1a7d987), but the PR description still claims they ship.**

After upgrade, operators with dashboards or alerts grouped by `session_id` will silently see empty results, and the orphaned high-cardinality `*_session_id` series will linger in Prometheus storage until retention expires.

Either restore the `observability.md` and `grafana-cluster.json` updates, or extend this changelog entry with an explicit upgrade note (e.g. *"operators must update dashboards keyed on `session_id`; existing per-session series will age out with retention"*).

hanahmily, line 37: **Issue 4 — no regression test added.**

The PR template's `Add a unit test to verify that the fix works` checkbox is still unchecked. The existing fixtures in `chunked_sync_test.go` (e.g. `test-session-1` at line 307, `test-session-2` at line 364) already construct sessions with `metadata.Topic` set, so a small test that wires a mock counter into `s.metrics`, drives an out-of-order chunk, and asserts that the emitted label equals the topic (and that no per-session series is created) would be straightforward.

Without it, a future revert to `req.SessionId` would silently restore the cardinality blow-up this PR is meant to prevent.

hanahmily (changes requested):
Pending review — 4 critical issues left as inline comments. Please address before merge.

CHANGES.md
OmCheeLin: I checked `docs/operation/grafana-cluster.json` and searched for `session_id` and those metric names (like `out_of_order_chunks_received`, `chunks_buffered`, `buffer_timeouts`, etc.), but couldn’t find anything.

In `docs/operation/observability.md`, there’s only one mention of `banyandb_queue_sub_total_msg_sent_err` (line 69). There’s no `session_id`, nor any of the chunk-ordering metrics introduced in this change.

That’s why I reverted the docs for now and plan to include the original metrics as well as the additional metrics introduced later in the final PR.

banyand/queue/sub/chunked_sync.go
OmCheeLin: done

banyand/queue/sub/server.go
OmCheeLin: done

CHANGES.md
OmCheeLin: done

hanahmily (approved):

PR #1103, opened 2026-04-29
Fix FODC lifecycle endpoint wrong encoding and increase the gRPC timeout between agent and proxy

mrproliu:
1. Increase the gRPC timeout between the FODC agent and proxy for fetching data. 
2. Do not cache the data if the data fetching fails. 
3. Fix the wrong encoding of the FODC lifecycle endpoint. 

### DIFF
--- a/fodc/agent/internal/lifecycle/collector.go
+++ b/fodc/agent/internal/lifecycle/collector.go
@@ -35,6 +35,7 @@ import (
 	"google.golang.org/grpc/status"
 
 	fodcv1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/fodc/v1"
+	"github.com/apache/skywalking-banyandb/fodc/internal/timeouts"
 	"github.com/apache/skywalking-banyandb/pkg/logger"
 )
 
@@ -43,10 +44,6 @@ const (
 	DefaultReportDir  = "/tmp/lifecycle-reports"
 	maxReportFiles    = 5
 	maxReportFileSize = 5 * 1024 * 1024 // 5MB
-	// DefaultGRPCTimeout bounds how long a single InspectAll call may run. It must be
-	// less than or equal to the proxy-side per-agent collection deadline; otherwise the
-	// proxy gives up before the agent has a chance to deliver a successful response.
-	DefaultGRPCTimeout = 40 * time.Second
 )
 
 // Collector collects lifecycle data from local files.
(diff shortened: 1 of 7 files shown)


fodc/proxy/internal/lifecycle/manager.go
```diff
@@ -28,7 +28,11 @@ import (
 	"github.com/apache/skywalking-banyandb/pkg/logger"
 )
 
-const defaultCollectionTimeout = 10 * time.Second
+// defaultCollectionTimeout bounds how long the proxy waits for each agent to push back its
+// lifecycle data. It must be greater than or equal to the agent-side InspectAll timeout
+// (see lifecycle.DefaultGRPCTimeout in fodc/agent/internal/lifecycle), otherwise the proxy
+// gives up before the agent has a chance to deliver a successful response.
+const defaultCollectionTimeout = 40 * time.Second
```
hanahmily: make the proxy strictly larger, e.g. proxy = agent + 10s slack (and update the comments to say "strictly greater than"). Or pass the deadline downstream rather than reapplying it locally.

fodc/agent/internal/lifecycle/collector.go
```diff
@@ -84,9 +89,18 @@ func (c *Collector) Collect(ctx context.Context) (*fodcv1.LifecycleData, error)
 	}
 	c.mu.RUnlock()
 
+	reports := c.readReportFiles()
+	groups, err := c.collectGroups(ctx)
+	if err != nil {
+		if c.log != nil {
+			c.log.Warn().Err(err).Msg("InspectAll failed; cache untouched, downstream will see empty groups")
+		}
+		return &fodcv1.LifecycleData{Reports: reports}, nil
```
hanahmily: the only consumer of Collect cannot distinguish a transient liaison failure from "liaison is healthy and reports zero groups."

propagate the error (or set a sentinel field/flag on LifecycleData) so client.sendLifecycleData can report failure rather than send a misleading empty success upstream.

hanahmily (changes requested):

fodc/proxy/internal/lifecycle/manager.go
mrproliu: Add the `ProxySlack` const about this.

fodc/agent/internal/lifecycle/collector.go
mrproliu: Done.

hanahmily (approved):

PR #1107, opened 2026-04-30
sub pr2: queue_sub chunked sync observability

OmCheeLin:
Improves **liaison internal `queue_sub` chunked sync** observability and ties gauge lifecycle to session boundaries.

- **Saturation**: `chunked_sync_active_sessions` (by `topic`) increments on session start/switch and decrements via `releaseMetrics` on completion/switch/stream teardown to prevent leaks.  
- **Reorder backlog**: `chunk_reorder_buffered_chunks` (by `topic`) increments on buffer insert and decrements on draining buffered chunks; `releaseMetrics` reconciles any remaining buffered chunks.  
- **Completion outcomes**: On `handleCompletion`, records `chunked_sync_total_bytes_received` and `chunked_sync_duration_seconds`; failed parts use **`chunked_sync_failed_parts_total{topic}`** only (no `reason` label—avoids misleading breakdown without real per-part semantics).  
- **Chunk-ordering**: `buffer_timeout` is emitted from `checkBufferTimeout`, invoked **before** refreshing `lastActivity` in `processChunkWithReordering` so timeouts can actually fire.  
- **Refactor**: `startOrSwitchSession` / `cleanupPreviousSession` centralize session switching/cleanup; `SyncPart` defer consistently releases metric-held state.  
- **Tests**: Adds/adjusts unit tests to cover metrics wiring, buffer timeout behavior, and `reorderBuffered` interactions.

### <Feature description>

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. part of apache/skywalking#13775.

### DIFF
--- a/banyand/queue/sub/chunked_sync.go
+++ b/banyand/queue/sub/chunked_sync.go
@@ -121,17 +121,33 @@ type partProgress struct {
 	completed     bool
 }
 
+// releaseMetrics releases the gauges held by this session.
+func (s *syncSession) releaseMetrics(m *metrics) {
+	if m == nil || s.completed {
+		return
+	}
+	m.activeSyncSessions.Add(-1, s.metadata.Topic)
+	if s.chunkBuffer != nil {
+		if n := len(s.chunkBuffer.chunks); n > 0 {
+			m.reorderBuffered.Add(-float64(n), s.metadata.Topic)
+		}
+	}
+	s.completed = true
+}
+
 // SyncPart implements clusterv1.ChunkedSyncServiceServer.
 func (s *server) SyncPart(stream clusterv1.ChunkedSyncService_SyncPartServer) error {
 	ctx := stream.Context()
 	var currentSession *syncSession
 	var sessionID string
 	defer func() {
-		if currentSession != nil {
-			if currentSession.partCtx != nil {
-				if closeErr := currentSession.partCtx.Close(); closeErr != nil {
-					s.log.Error().Err(closeErr).Str("session_id", currentSession.sessionID).Msg("failed to close session partCtx")
-				}
+		if currentSession == nil {
+			return
+		}
+		currentSession.releaseMetrics(s.metrics)
+		if currentSession.partCtx != nil {
+			if closeErr := currentSession.partCtx.Close(); closeErr != nil {
+				s.log.Error().Err(closeErr).Str("session_id", currentSession.sessionID).Msg("failed to close session partCtx")
 			}
 		}
 	}()
@@ -155,34 +171,7 @@ func (s *server) SyncPart(stream clusterv1.ChunkedSyncService_SyncPartServer) er
 		sessionID = req.SessionId
 
 		if req.GetMetadata() != nil {
-			if currentSession != nil {
-				if currentSession.partCtx != nil {
-					if currentSession.partCtx.Handler != nil {
-						if finishErr := currentSession.partCtx.Handler.FinishSync(); finishErr != nil {
-							s.updateChunkOrderMetrics("finish_sync_err", currentSession.metadata.Topic)
-							s.log.Error().Err(finishErr).Str("session_id", currentSession.sessionID).Msg("failed to finish sync for previous session")
-						}
-						if closeErr := currentSession.partCtx.Close(); closeErr != nil {
-							s.log.Error().Err(closeErr).Str("session_id", currentSession.sessionID).Msg("failed to close previous session partCtx")
-						}
-					} else if closeErr := currentSession.partCtx.Close(); closeErr != nil {
-						s.log.Error().Err(closeErr).Str("session_id", currentSession.sessionID).Msg("failed to close previous session partCtx")
-					}
-				}
-			}
-			currentSession = &syncSession{
-				sessionID:      sessionID,
-				metadata:       req.GetMetadata(),
-				startTime:      time.Now(),
-				chunksReceived: 0,
-				partsProgress:  make(map[int]*partProgress),
-			}
-			if dl := s.log.Debug(); dl.Enabled() {
-				dl.Str("session_id", sessionID).
-					Str("topic", req.GetMetadata().Topic).
-					Uint32("total_parts", req.GetMetadata().TotalParts).
-					Msg("started chunked sync session")
-			}
+			currentSession = s.startOrSwitchSession(sessionID, req, currentSession)
 		}
 
 		if currentSession == nil {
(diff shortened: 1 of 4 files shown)


banyand/queue/sub/server_metrics_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	require.Equal(t, float64(0), getGaugeValue(m.activeSyncSessions))
+	require.Equal(t, float64(0), getGaugeValue(m.reorderBuffered))
+	// idempotent
+	sess.releaseMetrics(m)
+	require.Equal(t, float64(0), getGaugeValue(m.activeSyncSessions))
+	require.Equal(t, float64(0), getGaugeValue(m.reorderBuffered))
+}
+
+func TestCompletionOutcomeMetricsRecorded(t *testing.T) {
```
hanahmily, line 136: **C1 (blocker): this test is a tautology.**

It does not invoke `handleCompletion`. Lines 123–131 re-implement the same metric-recording tail inline in the test body, and the only assertion is `require.True(t, session.completed)` on line 133. None of the three outcome metrics this test claims to verify (`chunkedSyncTotalBytes`, `chunkedSyncDurationSecs`, `chunkedSyncFailedParts`) are actually asserted on, and the production function is never called. Any future refactor of `handleCompletion` (e.g., dropping the histogram observation, or changing the failed-part count semantics) will silently keep this test green.

Please change this to actually call `s.handleCompletion(...)` (or factor a helper that both production and the test call) and assert on the metrics, e.g.:

```go
require.Equal(t, 10.0, s.metrics.chunkedSyncTotalBytes.(*fakeCounter).total)
require.Equal(t, 1, s.metrics.chunkedSyncDurationSecs.(*fakeHistogram).count)
require.Equal(t, 1.0, s.metrics.chunkedSyncFailedParts.(*fakeCounter).total) // one failed part out of two
```

Without this, the central claim of the PR — "outcome metrics are recorded on completion" — is unverified. For a metrics/observability PR, that's a hard block.

banyand/queue/sub/chunked_sync.go
```diff
@@ -204,6 +193,47 @@ func (s *server) SyncPart(stream clusterv1.ChunkedSyncService_SyncPartServer) er
...
+	if dl := s.log.Debug(); dl.Enabled() {
+		dl.Str("session_id", sessionID).
+			Str("topic", req.GetMetadata().Topic).
+			Uint32("total_parts", req.GetMetadata().TotalParts).
+			Msg("started chunked sync session")
+	}
+	return newSession
+}
+
+func (s *server) cleanupPreviousSession(previousSession *syncSession) {
```
hanahmily, line 248: **M2 (major): outcome metrics are not recorded for switched-out sessions.**

`cleanupPreviousSession` calls `releaseMetrics` (decrements the saturation gauges) and `FinishSync`, but never emits `chunkedSyncTotalBytes`, `chunkedSyncDurationSecs`, or `chunkedSyncFailedParts` for the abandoned session. Combined with the same gap in the `SyncPart` `defer` (see comment on line 143), this means only sessions that reach `handleCompletion` cleanly contribute to the outcome series.

Net effect: dashboards built on `chunked_sync_total_bytes_received` / `chunked_sync_duration_seconds` will under-count, and `chunked_sync_failed_parts_total` will be biased toward sessions that completed the gRPC handshake. A reader will reasonably assume those metrics cover all sessions.

Two possible fixes:
- (a) record outcome metrics here and in the `defer` with `success=false` and partial bytes/duration, or
- (b) add an explicit `chunked_sync_aborted_total{reason}` counter (`reason` ∈ {`switch`, `stream_error`, `ctx_done`, `eof`}) so the gap is visible rather than silent.

(b) is probably the cleaner option since aborted sessions don't have meaningful `partsResults` to enumerate.

banyand/queue/sub/chunked_sync.go
```diff
@@ -121,17 +121,33 @@ type partProgress struct {
...
+	}
+	s.completed = true
+}
+
 // SyncPart implements clusterv1.ChunkedSyncServiceServer.
 func (s *server) SyncPart(stream clusterv1.ChunkedSyncService_SyncPartServer) error {
 	ctx := stream.Context()
 	var currentSession *syncSession
 	var sessionID string
 	defer func() {
```
hanahmily, line 165: **M2 (major, pair to comment on `cleanupPreviousSession`):** this defer releases the saturation gauges on stream EOF / `Recv` error / `ctx.Done`, but does not record any outcome metric for the abandoned session. See the longer note on `cleanupPreviousSession` for the proposed fix (record-on-abort, or a dedicated `chunked_sync_aborted_total{reason}` counter).

banyand/queue/sub/chunked_sync.go
```diff
@@ -121,17 +121,33 @@ type partProgress struct {
 	completed     bool
 }
 
+// releaseMetrics releases the gauges held by this session.
+func (s *syncSession) releaseMetrics(m *metrics) {
```
hanahmily, line 134: **M3 (major): field-level nil-deref risk.**

The function guards on `m == nil` but then unconditionally calls `m.activeSyncSessions.Add(...)` and `m.reorderBuffered.Add(...)`. Production `newMetrics` always initializes both fields, so this is safe in prod today — but the existing `chunked_sync_test.go` test pattern (and the new `noopGauge` workaround that this very PR had to add to keep `TestChunkOrderingMetricsAreLabeledByTopic_NotSessionID` from panicking) shows partial-init `metrics{}` structs are routine in this package.

The next test author who writes `s.metrics = &metrics{outOfOrderChunksReceived: ...}` and then runs anything that triggers session teardown is going to hit a nil deref here.

Two options:
- guard each call individually (`if m.activeSyncSessions != nil { ... }`), or
- expose a package-private `noopGauge` / `noopCounter` (already drafted in `chunked_sync_test.go`) and tell test authors to use it everywhere via a helper like `newTestMetrics()`.

The second is preferable — centralizes the gotcha in one place rather than scattering field-level guards through production hot paths.

Also, please add the idempotency contract to the doc comment, since both `handleCompletion` and the `SyncPart` defer rely on it:

```go
// releaseMetrics releases the gauges held by this session. Idempotent:
// safe to call from both handleCompletion and the SyncPart defer.
```

hanahmily (changes requested):
## Summary

Nice refactor — `startOrSwitchSession` / `cleanupPreviousSession` / `releaseMetrics` is a clean factoring, and the saturation gauges (`chunked_sync_active_sessions`, `chunk_reorder_buffered_chunks`) are correctly paired across start/buffer/drain/teardown.

A few things to address before merge — two blockers and three majors. Inline comments below cover C1, M2, M3. C2 and M1 are repo-/PR-level and noted here:

### C2 — `CHANGES.md` not updated (blocker)

PR body confirms `[ ] Update the CHANGES log`. Project convention requires a `CHANGES.md` entry for any user-visible change, and new Prometheus metrics qualify. Please add an entry under the unreleased section (likely *Observability*) listing the five new series:

- `chunked_sync_active_sessions`
- `chunk_reorder_buffered_chunks`
- `chunked_sync_failed_parts_total`
- `chunked_sync_total_bytes_received`
- `chunked_sync_duration_seconds`

### M1 — PR description over-claims `buffer_timeout` ordering change

PR body says: *"`buffer_timeout` is emitted from `checkBufferTimeout`, invoked **before** refreshing `lastActivity` in `processChunkWithReordering` so timeouts can actually fire."*

The diff doesn't contain that change — `processChunkWithReordering` (around `chunked_sync.go:286-290`) already had the comment-and-order in place pre-PR. Either trim that bullet from the description, or, if there was an intended change that got lost in a rebase, restore it. Right now this reads as a feature-claim that the diff doesn't back up.

OmCheeLin:
No need to update `CHANGES.md` in every sub-PR. I’ll summarize them at the end.

banyand/queue/sub/server_metrics_test.go
OmCheeLin: done

banyand/queue/sub/chunked_sync.go
OmCheeLin: done

banyand/queue/sub/chunked_sync.go
OmCheeLin: done

banyand/queue/sub/chunked_sync.go
OmCheeLin: done

hanahmily (approved):

PR #1113, opened 2026-05-06
feat: queue_pub observability

### LINKED ISSUE #13775: Add queue observability for liaison internal pipelines
## Problem
Liaison queue behavior is not easy to observe end-to-end during chunked sync pressure and failures.

## Current context
- Queue-sub already exposes chunk-ordering/error metrics in `banyand/queue/sub/server.go`.
- Liaison wires `sub.NewServerWithPorts(..., "liaison-server", ...)` in `pkg/cmdsetup/liaison.go`.
- Existing dashboards aggregate some queue errors (`banyandb_queue_sub_total_msg_sent_err`) in `docs/operation/observability.md` and `docs/operation/grafana-cluster.json`.

## Proposal
- Add/extend liaison-focused queue metrics (depth, retries, per-topic throughput/latency, failed-part counters).
- Add dashboard panels and alert suggestions for liaison queue health.
- Document metric meanings and troubleshooting paths.

## Acceptance criteria
- New metrics are exported and documented.
- Dashboards include liaison queue saturation/failure visibility.
- Integration/e2e validation demonstrates metric changes under injected failures.

OmCheeLin:
- Adds **`MetricsRegistry()`** to **`metadata.Service`** so the queue publisher can obtain the shared metrics factory in **`PreRun`**.
- Introduces **`queue_pub`** Prometheus metrics: send attempts/bytes/latency, **`send_err_total{reason}`**, retries/backoff, **`inflight_streams`** / **`inflight_requests`**, plus recv/server-side outcomes in **`listenBatchResponse`**.
- Adds tests for **`reason`** labels, retry exhaustion, **`listenBatchResponse`**, **`Close`** inflight handling, and **Publish** reuse-path **`inflight_requests`** balance.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13775.


hanahmily:
@OmCheeLin check copilot comments

banyand/queue/pub/batch.go
```diff
@@ -162,48 +180,65 @@ func (bp *batchPublisher) Publish(ctx context.Context, topic bus.Topic, messages
 			client:    stream,
 			ctxDoneCh: streamCtx.Done(),
 		}
+		if bp.hasMetrics() {
+			bp.pub.metrics.inflightStreams.Add(1, node)
```
hanahmily, line 190: **Bug — `cancel()` leak when `Send()` fails (just above this hunk, lines 171-178):**

```go
streamCtx, cancel := context.WithTimeout(ctx, bp.timeout)
// this assignment is for getting around the go vet lint
deferFn := cancel
stream, errCreateStream := nodeClient.client.Send(streamCtx)
if errCreateStream != nil {
    err = multierr.Append(err, fmt.Errorf("failed to get stream for node %s: %w", node, errCreateStream))
    continue
}
```

The `deferFn := cancel` aliasing only suppresses the `lostcancel` lint — neither `cancel()` nor `deferFn()` is invoked when `Send()` fails, so `streamCtx` leaks until `bp.timeout` (15 s by default) expires. Under repeated transient failures this accumulates timer goroutines and context resources per dropped stream.

```go
if errCreateStream != nil {
    cancel()
    err = multierr.Append(err, fmt.Errorf("failed to get stream for node %s: %w", node, errCreateStream))
    continue
}
```

(Anchored on the newly-added `inflightStreams.Add(1, node)` line because the cancel block sits just outside the diff hunk.)

hanahmily (commented):
Three review comments focused on bugs that affect cluster operators relying on the new observability surface. The metrics surface itself is well-scoped; these are the actionable items I'd ask to address before merge.

banyand/queue/pub/batch.go
```diff
@@ -162,48 +180,65 @@ func (bp *batchPublisher) Publish(ctx context.Context, topic bus.Topic, messages
...
+			bp.pub.connMgr.RecordFailure(curNode, errRecv)
+			bc <- batchEvent{n: curNode, e: common.NewErrorWithStatus(modelv1.Status_STATUS_INTERNAL_ERROR, errRecv.Error())}
+		}
+		return
+	}
+	if resp == nil || resp.Error == "" {
+		return
+	}
+	if bp.hasMetrics() {
+		bp.pub.metrics.sendErrTotal.Inc(1, topic, curNode, sendErrReasonServerRejected)
```
hanahmily, line 232: **Bug — `server_rejected` is counted but never propagated to the caller.**

For non-failover statuses, this code increments `send_err_total{reason="server_rejected"}` but does **not** emit a `batchEvent`, so `Close()`'s `cee` map never sees it. Operators will see the counter rise on dashboards while the publishing call site silently treats the batch as successful — a worse failure mode than before, because the new metric advertises a problem that the code intentionally hides.

Two options:

1. **Surface the rejection** by emitting a `batchEvent` for any non-empty `resp.Error`, not just failover statuses:
   ```go
   if bp.hasMetrics() {
       bp.pub.metrics.sendErrTotal.Inc(1, topic, curNode, sendErrReasonServerRejected)
   }
   ce := common.NewErrorWithStatus(resp.Status, resp.Error)
   if isFailoverStatus(resp.Status) {
       bp.pub.connMgr.RecordFailure(curNode, ce)
   }
   bc <- batchEvent{n: curNode, e: ce}
   ```
2. Keep current behavior but document it: add a Go doc comment on `listenBatchResponse` and a note in `observability.md` that `server_rejected` is observability-only and the publisher will not retry/failover on non-failover statuses.

Option 1 is preferred — leaving the metric divergent from the visible error surface will lead to confused on-call investigations.

banyand/queue/pub/batch.go
```diff
@@ -280,8 +315,14 @@ func isFailoverStatus(s modelv1.Status) bool {
 }
 
 // retrySend implements bounded retries for client streaming sends with exponential backoff and jitter.
-func (bp *batchPublisher) retrySend(ctx context.Context, stream clusterv1.Service_SendClient, r *clusterv1.SendRequest, node string) error {
+func (bp *batchPublisher) retrySend(ctx context.Context, stream clusterv1.Service_SendClient, r *clusterv1.SendRequest, node string, topic string) error {
 	var lastErr error
+	if bp.hasMetrics() {
+		start := time.Now()
+		defer func() {
+			bp.pub.metrics.sendDurationSeconds.Observe(time.Since(start).Seconds(), topic, node)
```
hanahmily: **Bug — `send_duration_seconds` observed on every outcome skews the latency signal.**

This defer records duration for **every** return path — including µs-fast `InvalidArgument` rejections (`return sendErr` after the `IsTransientError` check), context cancellations, and stream-canceled returns. The histogram will be biased left by fail-fast errors, which dilutes operator signal when watching p99 send latency.

Two viable fixes:

- **Add a `result` label** so success and each failure mode are distinguishable:
  ```go
  factory.NewHistogram("send_duration_seconds", meter.DefBuckets, "topic", "node", "result")
  ```
  Then `Observe(..., topic, node, "success" | "non_transient" | "retry_exhausted" | "canceled")` at each return path.
- **Or only observe on success and retry-exhausted paths** — the two outcomes where end-to-end send latency is operationally meaningful.

Whichever path you take, please also update the doc table in `docs/operation/observability.md` (see paired comment on the `send_duration_seconds` row).

docs/operation/observability.md
```diff
@@ -258,6 +258,64 @@ If the value is too large, it may indicate that too many data points are being i
...
+
+**Troubleshooting:** rising `chunk_reorder_buffered_chunks` or `buffer_timeouts` suggests sustained out-of-order or slow consumers. Spikes in `chunked_sync_aborted_total` with `reason=switch` often correlate with topic/hand-off changes; `stream_error` / `ctx_done` / `eof` point to RPC lifecycle issues. Use `chunked_sync_failed_parts_total` and the duration histogram to separate partial completion from healthy throughput.
+
+#### `queue_pub` — outbound batch client
+
+| Metric (suffix after `banyandb_queue_pub_`) | Type | Labels | Meaning |
+| --- | --- | --- | --- |
+| `send_success_total` | Counter | `topic`, `node` | Successful `Send` on the client stream (local write, not end-to-end ack). |
+| `send_bytes_total` | Counter | `topic`, `node` | Payload bytes on successful `Send`. |
+| `send_duration_seconds` | Histogram | `topic`, `node` | Time spent in the send path including retries. |
```
hanahmily: Paired with the `send_duration_seconds` issue in `batch.go` — this row is technically true but operationally misleading because the histogram includes fail-fast paths (non-transient rejections, ctx-canceled, stream-canceled).

If the unconditional observe is kept, please append a clarifier, e.g.:

> Time spent in the send path including retries. **Recorded for both successful and failed attempts**; filter with `send_err_total` (or use a `result` label, if added) when isolating success latency.

If you split the histogram by `result` label, update the Labels column to `` `topic`, `node`, `result` ``.

hanahmily (approved):

PR #1120, opened 2026-05-08
Fix lifecycle migration segment shorter than the configured `SegmentInterval`

mrproliu:
## What

Two related lifecycle migration bugs:

1. Inspector reports cold-tier segments shorter than the configured `SegmentInterval` after a lifecycle migration — e.g. a 3-day segment under a 15-day `SegmentInterval` for `sw_metricsHour`.
2. In non-UTC timezones (reproduced on `America/Los_Angeles` CI runners), queries against newly migrated data return empty when the source hot LA-day straddles a multi-day UTC bucket boundary used by the warm tier — e.g. `should migrate data once correctly` failing intermittently in CI on the distributed lifecycle suite.

## Why

Two independent root causes:

1. **Receiver — `segment.create()` truncated late-arriving earlier-timestamp segments.** When a part with a later `MinTimestamp` was processed first and an earlier-timestamp part arrived afterwards (e.g. concurrent chunked-sync sessions during migration), the second `create()` call saw the existing later segment as `next` and capped `end = next.Start`, producing a span shorter than `SegmentInterval`.
2. **Sender — series-index, element-index, sidx and per-part chunked-sync messages all used the source segment's `MinTimestamp` (or the part's actual `MinTimestamp`) regardless of which target bucket the iteration was meant for.** The receiver picks the warm segment via `Standard(MinTimestamp)`, so every iteration routed to the bucket containing the source `MinTimestamp`. When the source range straddled a target bucket boundary (`Num*Day`-from-epoch UTC grid vs hot's local-day alignment in non-UTC zones), only one warm segment was created; the actual data parts could land in a sibling bucket while series/element indexes ended up in the source-anchored bucket. Queries that selected the data-side bucket then found an empty index and returned nothing.

## How

- **Receiver alignment (`banyand/internal/storage/segment.go`)**: when computing the new segment's `end`, anchor `stdEnd` to the unit-aligned start *before* bumping past a legacy off-grid neighbor. The new segment uses `next.Start` as `end` only when `next` is genuinely on a previous grid bucket (the legacy off-grid case, surfaced at Info level). This keeps the segment span at the configured `Num*Unit` for the steady state and only produces a short transition segment when an unaligned legacy neighbor is still on disk; the next segment self-heals back to the global grid.
- **Sender per-target routing (`banyand/backup/lifecycle/{stream,measure,trace}_migration_visitor.go`)**: every per-target iteration now overrides `MinTimestamp`/`MaxTimestamp` on the `StreamingPartData` to the target bucket's `[start, end)` (via `getSegmentTimeRange(targetSegmentTime, targetStageInterval)`). This applies to series index, element index, trace sidx (now also iterated per target with re-opened readers), and per-part migration. As a result the receiver creates exactly one warm segment per target bucket, and series/element/sidx indexes land in the same warm segment as the data parts that belong to that bucket.
- **Sender ttl/segment_interval validation (`banyand/backup/lifecycle/steps.go`)**: pre-loop nil-validation for `ro.Ttl`, `ro.SegmentInterval`, and every `Stages[*].SegmentInterval` / `Stages[*].Ttl` — fail fast with a descriptive error instead of panicking inside a downstream log message.
- **Sender stage interval (`banyand/backup/lifecycle/segment_boundary_utils.go` + `steps.go`)**: `GroupConfig.TargetSegmentInterval` is populated from the matched stage's next stage and `getTargetStageInterval` reads it, so 3+ stage clusters use the correct target interval for bucket math and progress keys instead of always falling back to `Stages[0]`.

## Compatibility

- Existing data is untouched. Legacy short segments age out via TTL.
- During the cold-tier transition, segments produced by the receiver remain a full `Num*Unit` span on the global grid; the only short segment that can be created is the transition segment that wraps a legacy off-grid neighbor, and it ends on the next grid boundary so subsequent segments self-heal.
- Migrated data placed by the new per-target routing is read identically to non-migrated data — `selectSegments` finds the bucket via overlap and queries it normally. Indexes (series/element/sidx) are now duplicated across all target buckets the source segment covers, mirroring the data routing.

## Tests

- `TestCreateSegment_OutOfOrderArrival_NoTruncation` and `TestCreateSegment_LegacyOffGridNeighbour_TransitionThenGrid`: regression coverage for the receiver-side alignment — out-of-order arrivals retain the full configured interval, and a transition past a legacy off-grid neighbor lands on the next grid boundary so the next segment is again full-span.
- `TestGetTargetStageInterval`: table-driven coverage for 2/3/4-stage configs and every fallback path; concurrent-reader smoke test.
- `TestParseGroup_RejectsMissingIntervals`: table-driven coverage for every required-IntervalRule nil case (top-level + per-stage), asserting `parseGroup` returns a descriptive error rather than panicking.
- Local LA-timezone reproduction with a pinned `BaseTime` inside the bug window (LA-day whose UTC equivalent crosses a 3-day UTC bucket boundary): pre-fix the distributed lifecycle suite fails at `test/cases/measure/data/data.go:122` with an empty response; post-fix it passes deterministically.


banyand/internal/storage/segment.go
```diff
@@ -579,19 +579,37 @@ func (sc *segmentController[T, O]) create(start time.Time) (*segment[T, O], erro
 		}
 	}
 	options := sc.getOptions()
-	start = options.SegmentInterval.Unit.Standard(start)
+	// Anchor stdEnd to the aligned start before any bump so end stays on the
+	// global grid even when start is bumped past a legacy off-grid neighbor;
+	// subsequent segments then self-heal back to the grid.
+	alignedStart := options.SegmentInterval.Standard(start)
```
hanahmily, line 585: ### Suffix round-trip is lossy on non-UTC nodes when `Num >= 2` -> adjacent segments overlap by the local UTC offset after restart

The `alignedStart` computed here flows into `format(start)` at line 617 (called from `sc.format`, lines 491-498), which renders the in-memory bucket start with `tm.Format(dayFormat)` reading year/month/day from `t.Location()`. After a process restart, `loadSegments` reconstructs `Start` via `parseSegmentTime` (lines 502-509), which reparses with `time.ParseInLocation(dayFormat, value, time.Local)`.

With `Num=1`, `IntervalRule.Standard` falls through to `Unit.Standard` which already returns local-midnight, so the round-trip is a fixed point. With `Num >= 2`, `Standard` returns a UTC-anchored instant that is **not** a local midnight in non-UTC zones, and the round-trip drops the offset.

Repro on an LA node with `SegmentInterval = 15d`:

- Bucket start in memory: `2026-04-07 00:00 UTC` = `2026-04-06 17:00 PDT`
- `format` -> `"20260406"` (LA-local date)
- After restart, `parseSegmentTime("20260406")` -> `2026-04-06 00:00 PDT` = `2026-04-06 07:00 UTC`
- **Reloaded `Start` is 17 h earlier than the original; `End` is recovered correctly from `meta.EndTime` (RFC3339Nano carries the offset).** Every adjacent pair of segments now overlaps by exactly the local UTC offset.

Downstream effects of the overlap:

- The bump loop at lines 591-600 assumes `sc.lst` is non-overlapping; that invariant is gone after restart.
- The reverse-iteration top-guard at lines 574-580 returns the newest containing segment, so writes whose timestamps fall in the overlap window get silently routed to the next bucket's segment instead of the one their data semantically belongs to.
- `selectSegments` (line 418) produces false-positive matches at every bucket seam.

`TestIntervalRule_Standard_PreservesLocation` only checks that the bucket instant is timezone-invariant; it doesn't catch this because the bug is in the suffix encoding, not the alignment math itself.

banyand/internal/storage/storage.go
```diff
@@ -202,6 +202,41 @@ func (ir IntervalRule) NextTime(current time.Time) time.Time {
...
+	switch ir.Unit {
+	case DAY:
+		bucketWidth = time.Duration(ir.Num) * 24 * time.Hour
+	case HOUR:
+		bucketWidth = time.Duration(ir.Num) * time.Hour
+	default:
+		panic("invalid interval unit")
+	}
+	width := bucketWidth.Nanoseconds()
+	nanos := t.Sub(epochUTC).Nanoseconds()
```
hanahmily: ### `Standard` (absolute ns) and `NextTime` (calendar `AddDate`) disagree by 1 h around DST transitions -> empty-range segments and 1 h write gaps

This line computes bucket boundaries via `t.Sub(epochUTC).Nanoseconds()` divided by `Num*24*time.Hour` -- a fixed-width nanosecond grid. `IntervalRule.NextTime` for `DAY` (lines 199-200 in this file) uses `current.AddDate(0, 0, ir.Num)` -- calendar-day arithmetic that preserves wall-clock time across DST. The two metrics agree on UTC nodes but disagree by 1 h whenever a bucket spans a DST transition on a non-UTC node.

In `segmentController.create()` (`segment.go:585-586`):

```go
alignedStart := options.SegmentInterval.Standard(start)
stdEnd := options.SegmentInterval.NextTime(alignedStart)
```

`alignedStart` is a UTC-anchored grid boundary presented in `t.Location()`. `stdEnd` is then `AddDate(15)` against that local presentation, which preserves wall-clock 16:00 across the DST transition rather than landing exactly on the next UTC grid boundary.

Repro on an LA node with `SegmentInterval = 15d`, around 2026 spring-forward (DST starts March 8):

- Bucket `[2026-03-07 00:00 UTC, 2026-03-22 00:00 UTC)`.
- `alignedStart = 2026-03-07 00:00 UTC = 2026-03-06 16:00 PST`.
- `stdEnd = AddDate(15)` against `2026-03-06 16:00 PST` -> `2026-03-21 16:00 PDT` (by 03-21 we're in PDT) = `2026-03-21 23:00 UTC`.
- Real grid boundary: `2026-03-22 00:00 UTC`. **Segment is created 1 h short.**

Now a write at `2026-03-21 23:30 UTC` (in the 1 h gap):

- Top guard at `segment.go:574-580`: existing segment ends at `23:00 UTC` -> does not contain probe.
- `alignedStart = Standard(probe) = 2026-03-07 00:00 UTC` (same bucket).
- `stdEnd = 2026-03-21 23:00 UTC` (same DST bug).
- Bump loop: existing segment contains `alignedStart` -> `start = 2026-03-21 23:00 UTC`.
- Falls through to `else` branch at `segment.go:614-616`, `end = stdEnd = 2026-03-21 23:00 UTC`.
- **`start == end` -- empty-range segment is created and persisted via `sc.load(start, end, ...)` at `segment.go:639`.**

This fires every spring/fall DST transition on non-UTC nodes with `Num >= 2`, deterministically. The cold-tier groups motivating this PR (`sw_metricsHour` at 15d, etc.) match this profile exactly.

hanahmily (changes requested):

hanahmily (approved):

PR #1126, opened 2026-05-11
Support multiple types for the same tag in a single measure or stream part

ButterBright:


banyand/measure/merger.go
```diff
@@ -359,6 +366,101 @@ func mergeBlocks(closeCh <-chan struct{}, bw *blockWriter, br *blockReader) (*pa
...
+			if conflictColumns[cf] == nil {
+				conflictColumns[cf] = make(map[string]struct{})
+			}
+			conflictColumns[cf][name] = struct{}{}
+		}
+	}
+	return conflictColumns
+}
+
+func readPartColumnTypes(p *part) map[string]map[string]pbv1.ValueType {
```
hanahmily: This function is called from `collectConflictColumns` for every input part on every merge. It walks every block metadata in the part and, for each block, issues `fs.MustReadData` for each tag family and unmarshals the column-family metadata — purely to detect whether the merge has a type conflict. The very next pass (`mergeBlocks`) then re-reads the same metadata when blocks are loaded. In the common case where the schema hasn't changed, the result is an empty conflict set and the entire scan is wasted IO. Worth thinking about whether the per-part column→type summary can be persisted (the trace and sidx variants already do this — see `banyand/trace/merger.go` where `pw.p.tagType` is read directly) or at least cached on the `part` after open. Mirror in `banyand/stream/merger.go`.

banyand/measure/column.go
```diff
@@ -27,6 +29,57 @@ import (
 	"github.com/apache/skywalking-banyandb/pkg/pool"
 )
 
+const typedColumnSeparator = "#"
```
hanahmily, line 32: This introduces `#` as a reserved separator in on-disk column names. Please confirm that the schema validator forbids `#` in user-provided tag names. If a tag named `foo#int` is allowed, `decodeTypedColumn` will silently strip the suffix to `foo`, which then causes `collectConflictColumns` to manufacture a phantom conflict and the merger to rename a non-conflicting column. Same applies to `banyand/stream/tag.go`.

banyand/measure/merger.go
```diff
@@ -359,6 +366,101 @@ func mergeBlocks(closeCh <-chan struct{}, bw *blockWriter, br *blockReader) (*pa
...
+	for pmi.nextBlockMetadata() {
+		for cf, db := range pmi.block.bm.tagFamilies {
+			reader, ok := p.tagFamilyMetadata[cf]
+			if !ok {
+				continue
+			}
+			bb.Buf = bytes.ResizeExact(bb.Buf, int(db.size))
+			fs.MustReadData(reader, int64(db.offset), bb.Buf)
+			cfm := generateColumnFamilyMetadata()
+			if _, err := cfm.unmarshal(bb.Buf); err != nil {
```
hanahmily: Unmarshal failure is silently swallowed. If a part's column-family metadata block is corrupted, conflict detection skips the column, the merger doesn't rename it, and the resulting merged part ends up with the exact duplicate-typed columns this change is meant to prevent — without any signal that something went wrong. At minimum this should log a warning. Same in `banyand/stream/merger.go`.

banyand/measure/block.go
```diff
@@ -225,13 +226,24 @@ func (b *block) unmarshalTagFamily(decoder *encoding.BytesBlockDecoder, tfIndex
...
+			typedName := encodeTypedColumn(tp, schemaType)
+			for i := range cfm.columnMetadata {
+				if cfm.columnMetadata[i].name == typedName {
+					cc[j].mustReadValues(decoder, valueReader, cfm.columnMetadata[i], uint64(b.Len()))
+					cc[j].name = tp
+					continue NEXT
+				}
+			}
+		}
 		for i := range cfm.columnMetadata {
```
hanahmily, line 250: After the typed-name lookup misses, the fallback matches on the plain name without checking that the on-disk `valueType` agrees with the schema type. The four downstream output sites in this file currently filter out wrong-typed columns with `c.valueType == schemaType` guards, so the user-visible result is correct today — but the contract here is fragile: every future read site that consumes `cf.columns[i].values` must remember the rule. Compare `banyand/trace/block.go` where the projection planner gates the match on `bm.tagType[typedTag] == schemaType` once at read time, so wrong-typed columns simply don't appear in the cursor. Mirror in `banyand/stream/block.go`.

banyand/measure/block.go
```diff
@@ -956,6 +968,10 @@ func fastTagAppend(bi, b *blockPointer, offset int) error {
 				return fmt.Errorf("unexpected tag name for tag family %q: got %q; want %q",
 					bi.tagFamilies[i].name, b.tagFamilies[i].columns[j].name, bi.tagFamilies[i].columns[j].name)
 			}
+		}
+	}
+	for i := range bi.tagFamilies {
```
hanahmily, line 996: The original `fastTagAppend` performed the name check and the append inside the same loop. The change splits it into two full passes over the same tag families × columns with no functional reason — renaming happens on the source block before this function is called, so column names are stable here. The split is extra work and creates two places that must stay in sync. Mirror in `banyand/stream/block.go`.

banyand/measure/column.go
```diff
@@ -27,6 +29,57 @@ import (
...
+	for suffix := range suffixToValueType {
+		sepSuffix := typedColumnSeparator + suffix
+		if strings.HasSuffix(key, sepSuffix) {
+			return key[:len(key)-len(sepSuffix)]
+		}
+	}
+	return key
+}
+
+func hasTypeSuffix(key string) bool {
```
hanahmily: `hasTypeSuffix` (and its unit test) is added here but is never called from `banyand/measure` — same for the stream counterpart in `banyand/stream/tag.go`. The only callers live in `banyand/trace` and `banyand/internal/sidx`. Either remove the dead copies from measure/stream or wire them into the reader the way trace does.

hanahmily (commented):
Focused review on compatibility, correctness, and performance. Comments inline. Mirror issues exist in `banyand/stream/...` for items that have a stream counterpart.

banyand/measure/column.go
ButterBright: It has been implemented in api/validate/validate.go.

hanahmily (approved):

PR #1133, opened 2026-05-15
Fix incorrect counts, rate and missing trace fields in the lifecycle migration report

mrproliu:


hanahmily (commented):
## I don't agree with this design change

There are parts of this PR that are clearly wins — idempotent `Mark*Completed/Error`, all seven buckets handled in `ClearErrors`, the trace blocks in the report, the partial-failure report path, and the shard-key alignment fix in `traceMigrationVisitor.VisitShard`. The new tests pin those invariants well.

What I disagree with is removing the pre-walk and folding the denominator into `Mark*` itself, because it silently changes what the report means while keeping the same field names and schema shape.

### Before this PR
- `Counts` = pre-walk count of source items (the *planned* denominator).
- `Progress` = items the visitor finished.
- `completion_rate = Progress / Counts` was a real workload-remaining metric. Reading `progress.json` mid-flight told you "47 / 200 parts done".

### After this PR
The invariant inside `Mark*` is now:

```
Counts == Progress + len(errors)
```

So `Counts` no longer represents "what we planned to do" — it is "what we attempted so far". That has consequences I don't think this PR has reckoned with:

1. **A clean cycle always reports `completion_rate = 100%`**, regardless of how much source data was actually processed. If a future regression caused the visitor to silently skip items (never reaching `Mark*`), the report would still happily report 100 %. The old pre-walk would have caught this as `Progress < Counts`.
2. **Mid-flight observability is gone.** Because `Counts` is bumped in lock-step with `Progress`, `completion_rate` reads ~100 % even one second into a five-hour migration. The only live signal left is the per-segment log line.
3. **The number is now a success ratio, not a progress percentage** — but the field name, the report schema (`report_version=2.0`), and the surrounding tests (e.g. `assertResource(... rate, 83.33 ...)`) still read as "fraction of work done".

Concretely: a migration that successfully processes 1 part and a migration that successfully processes 10 000 parts serialize to the exact same `stream_migration.parts: {total: N, completed: N, completion_rate: 100}` shape, with no way to distinguish them from the report alone. A v2.0 schema in which `completion_rate=100%` can mean both "we migrated everything" and "we migrated one thing and then stopped" is not something I'm willing to sign off on.

The other fixes in this PR are independent of the pre-walk removal and could land on their own.

mrproliu:
Quick clarification on the trade-off before I pick a direction:

  Just restoring pre-walk and reverting Mark*++Counts doesn't fully bring the planned semantic back — for series and element_index, the visitor sets

  Counts = len(segmentFiles)

  at the top of VisitSeries / VisitElementIndex, but Mark* runs inside for targetSegments × shardIDs. When fan-out > 1 (typical hot → warm), Mark calls exceed source files and the report ends up with e.g. series.completion_rate = 266.67%,
  while element_index flips the other way to single-digit % with errors = 0. Operators in the original SRE report flagged exactly those numbers.

  So "Counts = pre-walk source" and "rate stays in [0,100]" cannot both hold for series / element_index under the current visitor shape. To avoid the 266% regression there are three options:

  1. Document & accept: keep pre-walk semantic, treat rate > 100% as "fan-out write count exceeded source count" and document it.
  2. Bump report_version to 2.1: add planned_* fields from pre-walk alongside existing fields; old fields stay [0,100], planned_completion_rate gives the mid-flight progress signal.
  3. Refactor the visitor: collapse per-source Mark* to one call (lose per-target granularity for retries, but Counts = source = mark calls becomes consistent).

  Which one do you prefer?

hanahmily:
The v2 report contract must ensure one invariant:

**`completion_rate == 100%` means "the migration finished successfully" for every resource type — parts, series, and element_index alike. Nothing less, and never more.**

Option 1 and option 2 both leave the primary `completion_rate` able to read 266 % or 5 %, which violates the invariant regardless of documentation or sibling fields.

Option 3 is the path that satisfies it. Reshape the visitor so `Counts` reflects source-cardinality and the numerator counts source items whose entire fan-out succeeded. Per-target tracking for resume granularity can stay internal — the report only needs to expose source-level numbers.

hanahmily (approved):

PR #1135, opened 2026-05-16
Fix backup restore path traversal

### LINKED ISSUE #13871: [Bug] [BanyanDB] Backup restore can write outside the target data directory
### Search before asking

### What happened

`banyand/backup/restore.go` builds local restore paths from object names returned by the remote filesystem. The code does not verify that the final cleaned path stays under `rootPath/<catalog>/data`.

Affected path:

- `restoreByName` lists remote files with `fs.List`.
- It derives `relPath` from the remote object name.
- It builds `localPath` with `filepath.Join(rootPath, catalogName, storage.DataDir, relPath)`.

If a backup source contains an object name with parent-directory segments under the expected backup prefix, restore can create a file outside the intended catalog data directory.

There is a related issue in `pkg/fs/remote/local/local.go`: the local `file://` remote filesystem adapter joins caller-supplied paths directly with `baseDir` in `Upload`, `Download`, `List`, and `Delete`. Those operations also need a containment check after path cleaning.

I verified both behaviors locally with focused Go tests.

Restore should reject any remote object path that escapes the expected backup prefix or resolves outside the target catalog data directory.

The local `file://` remote filesystem adapter should also reject paths that resolve outside its configured base directory.

### How to reproduce

Minimal restore test setup:

1. Configure or mock a backup source that returns an object name under the expected backup prefix but containing parent-directory traversal segments.
2. Run `restoreByName`.
3. Check the filesystem. The restored file is written outside `rootPath/<catalog>/data`.

For the local remote filesystem adapter:

1. Create a local remote filesystem with a temporary base directory.
2. Call `Upload` with a path containing parent-directory traversal segments.
3. The file is created outside the configured base directory.

### Anything else

Suggested fix:

- Reject absolute paths.
- Reject cleaned paths that are `..` or start with `../`.
- In the local filesystem adapter, compare the cleaned absolute target path against the cleaned absolute base directory before reading, writing, listing, or deleting.
- In restore, validate every object returned by `fs.List` before deriving the local path.
- Add regression tests for restore and the local `file://` remote filesystem adapter.

### Code of Conduct

paultanay:
## Summary

  Fixes apache/skywalking#13871.

  Backup restore previously trusted remote object names when deriving local restore paths. A crafted
  backup object containing parent-directory segments could escape the intended `<root>/<catalog>/data`
  directory. The local `file://` remote filesystem adapter had the same containment problem when joining
  caller-provided paths with its base directory.

  ## Changes
  - Validate remote object paths before using them to derive restore destinations.
  - Reject remote paths that are absolute or escape the expected backup catalog prefix.
  - Normalize the local `file://` base directory and reject any operation that resolves outside it.
  - Add regression coverage for restore traversal and local remote filesystem traversal attempts.

  ## Verification
  - `make generate`
  - `go test ./banyand/backup ./pkg/fs/remote/local`
  - `bin/golangci-lint run -v --config .golangci.yml ./banyand/backup ./pkg/fs/remote/local`
  - `bin/ginkgo -v ./banyand/trace`
  - `bin/ginkgo -v ./banyand/stream`

  ## Notes
  - Full `make test` was run twice locally and failed in different suites (`trace`, then `stream`), but
  both suites passed when rerun individually.
  - Full `make lint` is blocked locally by an unrelated UI build prerequisite: `ui/embed.go:25:12: pattern
  dist: no matching files found`.

### DIFF
--- a/banyand/backup/restore.go
+++ b/banyand/backup/restore.go
@@ -208,7 +208,7 @@ func restoreByName(fs remote.FS, timeDir, rootPath, catalogName string) error {
 	}
 
 	for _, localRelPath := range localFiles {
-		localRelPathWithCatalog := filepath.Join(catalogName, localRelPath)
+		localRelPathWithCatalog := path.Join(catalogName, filepath.ToSlash(localRelPath))
 		if !remoteRelSet[localRelPathWithCatalog] {
 			localPath := filepath.Join(localDir, localRelPath)
 			logger.Infof("found local file: %s not exist in the remote storage, so delete it", localRelPathWithCatalog)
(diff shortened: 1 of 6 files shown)


hanahmily:
@paultanay fix the failed test and comments from copilot

paultanay:
@hanahmily could you please run the ci once?

hanahmily (approved):

PR #1144, opened 2026-05-28
feat: post-trace pipeline design — native-plugin sampler + per-stage retention

hanahmily:
## Summary

Adds the design doc and API surface for a **storage-node post-trace pipeline** on BanyanDB data nodes. The pipeline runs asynchronously during storage lifecycle events; the **entire keep/drop verdict is owned by a user-supplied native Go plugin** (`SamplerPlugin`, a `.so` loaded in-process via Go's `plugin` package). The same plugin mechanism is the *only* sampler at both filter points:

- **Gating** (`TracePipelineConfig.sampler_plugin`) — runs **only in the Hot phase**, at the enabled `PipelineEvent`(s): `PIPELINE_EVENT_MERGE` (in-merge during Hot compaction) and/or `PIPELINE_EVENT_FINALIZE` (on settled Hot segments). Warm/Cold compactions stay byte-for-byte lossless.
- **Per-stage retention** (`StageRule.sampler_plugin`) — runs at each stage's **migration-out** boundary (`Hot → Warm`, `Warm → Cold`); the "rising bar" is expressed by per-stage plugin config. Cold is terminal (no internal sampling).

**No implementation code in this PR — design + proto only.** Marked draft for review.

**Design philosophy:** every claim is grounded in BanyanDB's actual storage model (event-time segments, part-count-driven LSM compaction, no "segment sealed" event, watermark-driven settling, the native columnar trace `block`) rather than a generic streaming-telemetry idea grafted on. Scenario examples use real data sampled from the `skywalking-showcase` cluster.

## What's in this PR (2 commits, 3 files)

- **`docs/design/post-trace-pipeline.md`** (new, 542 lines) — full design: architecture (§1), proto message design + targeting + uniqueness/conflict + admission validation + the native-plugin SDK contract (§2), retention-decision timing & trace completeness (§3), retention vs time-aging (§4), downstream lifecycle actions (§5), two real-data scenario configs (§6), data-flow architecture (§7), step-by-step execution (§8). Diagrams are **Mermaid**.
- **`api/proto/banyandb/pipeline/v1/trace_pipeline.proto`** (new) — `PipelineEvent` enum, `TracePipelineConfig`, `StageRule`, `SamplerPlugin`, and the full `TracePipelineRegistryService` (Create/Update/Delete/Get/List/Exist + HTTP mappings under `/v1/trace-pipeline/schema`). PGV validation on bounds.
- **`docs/api-reference.md`** — regenerated by `protoc-gen-doc` to include the new `banyandb/pipeline/v1` section.

## Key design decisions (each grounded in real codebase facts)

1. **Native Go plugin is the sole sampler.** No declarative tail-sampling vocabulary in the proto. The plugin owns the verdict over a vectorized batch and returns a boolean keep-mask. The contract (vectorized batch, projection handshake, verdict shape) lives in the pinned Go SDK module `pkg/pipeline/sdk`, described in §2.5 — not in the proto.
2. **Gating is Hot-phase only.** `PIPELINE_EVENT_MERGE`/`_FINALIZE` fire only while a segment is Hot; Warm/Cold compactions run the same merger with the hook disabled, so they remain lossless. Per-stage retention is the mechanism that decides what crosses each migration boundary.
3. **In-merge filter** integrates at `mergeBlocks` (`banyand/trace/merger.go`), wrapping both `mustWriteRawBlock` and `mustWriteBlock`. Per-`trace_id` verdicts. Drops are gated on per-trace `merge_grace` because compaction is **part-count-driven** (`getPartsToMerge`), not time-driven, so it routinely runs on the active write window (§7.1).
4. **Finalization** uses an **event-time watermark + `finalize_grace`**, not a "segment sealed" event — because BanyanDB has none: `segmentController.create` pre-creates the next segment up to `creationGap` (1h) *before* the current one ends, and event-time routing keeps late data landing in old segments (§7.3).
5. **Pre-migration rewrite** runs as a standalone reduced-part producer between `generateAllPartData` and `streamPartToTargetShard`; the migration byte-copy stays lossless (§7.2).
6. **`merge_grace` (per-trace) vs `finalize_grace` (per-segment)** — distinct knobs for distinct correctness problems (intra-trace span spread vs segment-wide late arrival). Both `Duration` with PGV `gt: 0s` and documented engine defaults (30s / 5m).
7. **Vectorized, native-layout batch with optional spans (§2.5).** The batch reuses BanyanDB's native trace `block` (`banyand/trace/block.go`), not Apache Arrow. `Project()` returns `Projection{ Tags []string; SpanIDs bool; Spans bool }`: tags are projected by name, and the spans stream (span ids and bodies, which share one encoded data block) is **opt-in, default off** — only `trace_id`/`MinTS`/`MaxTS` are free intrinsics. A plugin that needs none of them stays on the merge raw fast path.
8. **Read-only batch contract + fail-open (§2.5).** The keep-mask is the plugin's only output channel; the engine writes retained traces from its own untouched block data, so a buggy plugin cannot corrupt stored data. `Decide` panics/errors/bad-length verdicts fail **open** (retain all).
9. **Plugin config is a `google.protobuf.Struct`** set directly in the pipeline config — the engine serializes it to JSON and hands the bytes to `NewSampler([]byte)`, keeping the `.so` boundary stdlib-only (survives Go plugin's toolchain version-lock) while staying inspectable on the wire.
10. **Conflict policy uses two disjoint keys (§2.3):** gating `(Group, Schema, Event)` and retention `(Group, Schema, Source-Stage)`. At most one active pipeline per key; the registry rejects conflicting writes. Runtime fail-safe: on any ambiguous match, the engine logs and **retains** (never a destructive drop on ambiguity).
11. **Cluster-aware admission (§2.4):** the liaison validates static fields at write time (flag, trusted-dir path, abi_version, well-formed Struct); the `.so` load / symbol / ABI match / `NewSampler` validation happens on each **data node** (where plugins live) at apply time. PGV covers `enabled_events` (defined-only, non-`UNSPECIFIED`), `schema_names` (non-empty), `stage`/`path` (non-empty), `abi_version` (≥1); server-side rules add event-set dedup, stage uniqueness, non-terminal-stage, and regex-compile checks.

## Operational constraints (knowingly accepted)

Native Go `.so` is the **only** sampler path by deliberate choice. Its constraints — Linux/macOS only, no unload (config change ⇒ node restart), a panic is contained with `recover()` (fail-open), and the irreducible toolchain/build-flag version-lock — are documented in §2.5 and accepted as the trade-off for in-process, zero-copy, fully-programmable verdicts. The SDK is the single pinned surface operators build against.

## What's intentionally NOT here

- **No implementation** — design + proto only. Follow-up PRs add the merge-filter hook, pre-migration rewrite, finalization scheduler, the `pkg/pipeline/sdk` module, the registry service implementation, and the required **trace-ID-aware sidx rewrite** (§7.1 point 5 — today's `sidx.Merge` is part-ID-based and must be extended so dropped traces don't leave dangling index entries).
- **No weighted scoring model** — an earlier weighted-score proposal was dropped; under the plugin model, scoring like that would live inside a sampler plugin, not as engine-native fields.

## Review

Reviewed with an external critic (Codex) across two passes; all findings resolved or explicitly accepted (e.g. the `.so`-only operational risk is an accepted product decision).

## Local CI status

`make generate / build / lint / check` pass on this branch; `buf lint` clean; `go build` of the generated `banyandb/pipeline/v1` package succeeds. `make test*` not run — pure design + proto changes have no Go test impact.

## Test plan

- [ ] Review the design doc end-to-end; flag any storage-model assumption that doesn't match current code (`banyand/trace/merger.go`, `banyand/trace/block.go`, `banyand/internal/storage/rotation.go`, `banyand/backup/lifecycle/trace_migration_visitor.go`).
- [ ] Review `trace_pipeline.proto` field-by-field: the `PipelineEvent` enum, `SamplerPlugin` (path/symbol/abi_version/config), `StageRule.sampler_plugin`, the PGV bounds, and the `TracePipelineRegistryService` shape vs the existing `database/v1` registries.
- [ ] Confirm the §2.5 SDK contract is acceptable as an implementation requirement (vectorized native-block batch, `Project()` projection incl. optional spans, read-only contract, fail-open).
- [ ] Confirm the two-key uniqueness/conflict policy (§2.3) and the split static-vs-data-node admission flow (§2.4).
- [ ] Confirm the two scenario configs (§6.1 `sw_trace`, §6.2 `sw_zipkinTrace`) match intended retention behavior.
- [ ] Run `make generate && make build && make lint` locally — codegen produces clean `.pb.go` / `.pb.validate.go` / `.pb.gw.go` for the new package.

### DIFF
--- a/api/proto/banyandb/pipeline/v1/trace_pipeline.proto
+++ b/api/proto/banyandb/pipeline/v1/trace_pipeline.proto
@@ -1,0 +1,271 @@
+// Licensed to Apache Software Foundation (ASF) under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Apache Software Foundation (ASF) licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+syntax = "proto3";
+
+package banyandb.pipeline.v1;
+
+import "banyandb/common/v1/common.proto";
+import "google/api/annotations.proto";
+import "google/protobuf/duration.proto";
+import "google/protobuf/struct.proto";
+import "protoc-gen-openapiv2/options/annotations.proto";
+import "validate/validate.proto";
+
+option go_package = "github.com/apache/skywalking-banyandb/api/proto/banyandb/pipeline/v1";
+option java_package = "org.apache.skywalking.banyandb.pipeline.v1";
+option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {base_path: "/api"};
+
+// PipelineEvent identifies a pipeline-wide event that can be independently
+// enabled. Per-stage retention (StageRule.sampler_plugin) fires implicitly at
+// the stage's migration-out boundary and is not toggleable via this enum.
+enum PipelineEvent {
+  PIPELINE_EVENT_UNSPECIFIED = 0;
+  // In-merge filter during Hot-phase LSM compaction merges (Warm/Cold
+  // compactions stay lossless). Per-trace drops are gated by `merge_grace` so
+  // partial traces are not destroyed prematurely. Cheap, runs often; verdicts
+  // wait for trace maturity (see §7.1).
+  PIPELINE_EVENT_MERGE = 1;
+  // Tail-sampling gate at Hot-phase segment finalization, after the segment
+  // has settled (event-time watermark past `segment.End + finalize_grace`).
+  // Heavy but authoritative; sees the complete trace (see §7.3).
+  PIPELINE_EVENT_FINALIZE = 2;
+}
+
+// TracePipelineConfig is the root configuration for a storage-node trace pipeline.
+// It reuses existing catalog identifiers (group via metadata, stage names, schema
+// names) for targeting instead of declaring a parallel metadata model.
+//
+// The pipeline has up to three filter points:
+//   1. PIPELINE_EVENT_MERGE — in-merge filter during Hot-phase LSM compaction (default).
+//   2. PIPELINE_EVENT_FINALIZE — tail-sampling gate at Hot-phase finalization.
+//   3. Per-stage retention via StageRule.sampler_plugin, applied at the
+//      stage's migration-out boundary (when the segment migrates to the next
+//      stage). Always implicit when any StageRule carries a sampler_plugin.
+// Events 1 and 2 are toggleable via `enabled_events`. The gating policy those
+// events evaluate is a native Go plugin (`sampler_plugin`).
+message TracePipelineConfig {
+  // Identity and revision tracking; metadata.group is the Group this pipeline
+  // lives in and applies to, consistent with every other schema resource.
+  // Required: every config needs a name/group for registry handling.
+  common.v1.Metadata metadata = 1 [(validate.rules).message = {required: true}];
+  // Active status of the pipeline.
+  bool enabled = 2;
+  // Per-stage retention rules: which lifecycle stages this pipeline acts on,
+  // with the retention sampler_plugin for each. Each rule fires at its stage's
+  // migration-out boundary regardless of `enabled_events`. Empty means the
+  // only filters are the `enabled_events` events (no per-stage drop).
+  repeated StageRule stages = 3;
+  // Explicit schema names to target within the Group (exact match on Metadata.name).
+  // Each entry must be non-empty; cross-element uniqueness is enforced server-side.
+  repeated string schema_names = 4 [(validate.rules).repeated.items.string.min_len = 1];
+  // RE2 regular expression matched against schema names. A schema is targeted if it
+  // is listed in schema_names OR matches this pattern. When both are empty, every
+  // schema in the Group is targeted.
+  string schema_name_regex = 5;
+  // Gating policy: a user-supplied native Go plugin (.so) loaded in-process
+  // that owns the entire keep/drop verdict, evaluated by any enabled event
+  // (PIPELINE_EVENT_MERGE and/or PIPELINE_EVENT_FINALIZE). Unset means the only
+  // retention is the per-stage StageRule sampler_plugin(s) at migration-out.
+  SamplerPlugin sampler_plugin = 6;
+  // Pipeline-wide events to run. Empty defaults to [PIPELINE_EVENT_MERGE] —
+  // the in-merge filter is on, the finalization gate is off. To enable the
+  // finalization gate, include PIPELINE_EVENT_FINALIZE; to disable the merge
+  // filter, list only [PIPELINE_EVENT_FINALIZE]; the explicit empty default
+  // value is also acceptable to mean "merge only". Each element must be a
+  // defined, non-UNSPECIFIED value; duplicates are normalized to a set
+  // server-side.
+  repeated PipelineEvent enabled_events = 7 [(validate.rules).repeated.items.enum = {
+    defined_only: true
+    not_in: [0]
+  }];
+  // Per-trace maturity window for the in-merge filter (§7.1). A trace is
+  // eligible for dropping during an LSM compaction merge only once its latest
+  // span timestamp is older than `now - merge_grace`; younger traces pass
+  // through the merge unchanged. Bounds the expected intra-trace span arrival
+  // spread (typically seconds). Used iff `enabled_events` contains
+  // PIPELINE_EVENT_MERGE. Strictly positive if set; engine default 30s if unset.
+  google.protobuf.Duration merge_grace = 8 [(validate.rules).duration = {
+    gt: {seconds: 0}
+  }];
+  // Per-segment settling window for the scheduled finalization pass (§7.3). A
+  // segment is treated as settled, and the authoritative final filter runs,
+  // once the event-time watermark exceeds `segment.End + finalize_grace`.
+  // Bounds segment-wide late arrival (typically minutes). Used iff
+  // `enabled_events` contains PIPELINE_EVENT_FINALIZE. Strictly positive if
+  // set; engine default 5m if unset.
+  google.protobuf.Duration finalize_grace = 9 [(validate.rules).duration = {
+    gt: {seconds: 0}
+  }];
+}
+
+// StageRule binds the pipeline to one lifecycle stage of the targeted Group
+// and declares that stage's retention sampler. The rule fires at the stage's
+// migration-out boundary (i.e. when a segment migrates from this stage to the
+// next stage); routine compaction is governed by PIPELINE_EVENT_MERGE on
+// TracePipelineConfig, not by StageRule.
+//
+// Per-stage retention uses the SAME native-plugin mechanism as gating: each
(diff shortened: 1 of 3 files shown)


hanahmily:
Addressed all 9 Copilot review comments. Branch was force-pushed; commit count stays at 3 (proto / design / backlog) as originally planned, with the fixes folded into the relevant commits.

**Proto validation (4 comments):**
- `TracePipelineConfig.metadata` — now `[(validate.rules).message = {required: true}]` so empty configs are rejected at parse time rather than reaching the registry.
- `StageRule.stage`, `TagSamplingRule.tag_key`, `TagMatcher.tag_key` — all now `[(validate.rules).string.min_len = 1]`, matching the project convention (e.g. `common.v1.LifecycleStage.name`). An empty stage name or tag key would have been silently un-matchable.

**Backlog doc inconsistency (1 comment):**
- The "field 2 and field 7 are reserved" claim was stale: those reservations were dropped during the "remove all scoring references" pass earlier in the branch, and the gaps were closed (StageRule field 2 is now `apply_at`; TracePipelineConfig field 7 is now `merge_grace`). Updated the backlog doc to say so explicitly — any future revival must pick new field numbers. The renumber is wire-safe because the proto has never been released.

**Section numbering (3 comments):**
- Jumped from §4.2 straight to §6.1 because the original §5 was a one-paragraph intro that got folded into §6 during an earlier rewrite, leaving subsection labels off by one. Renumbered to keep the sequence contiguous: §6.x → §5.x, §7.x → §6.x, §8.x → §7.x — applied to both subsection headings and every cross-reference (in main doc, proto comments, and backlog doc). No `§8.x` or "Scenario 7/8" left anywhere.

**Finalization composition semantics (1 comment):**
- The §7.3 design-decision blockquote previously said "traces failing **both** the gate and any FINALIZE `StageRule` predicates are dropped", which could be read as `StageRule` being able to override a tail-sampling drop. Rewrote it to be explicit: `tail_sampling` is the **absolute gate**, evaluated first — traces failing it are dropped immediately and never reach any `StageRule`. Surviving traces are then evaluated against any FINALIZE-targeted `StageRule`; failing those is also a drop. To survive finalization a trace must pass **both**. This now matches §1.1's "decides whether the trace survives at all" framing.

`make generate / build / lint / check` all pass locally; CI was clean on the previous push, the regenerated `docs/api-reference.md` picked up the new validation comments only.

hanahmily:
Force-pushed a simplification: **`StageEvent` enum and `apply_at` field removed**; per-stage filtering now fires only at migration-out. The `merge_grace` field, the §7.1 in-merge filter chapter, and the §5.1 compaction-rewrites section are gone as natural consequences.

**Two-anchored-points architecture (down from three):**
- **Tail-sampling gate at finalization** (`tail_sampling`, gated by `finalize_grace` watermark) — decides whether a trace is retained at all once its segment is settled.
- **Per-stage retention at migration-out** (`StageRule` predicates — `min_duration` / `keep_errors` / `keep_tag_rules`) — decides what migrates to the next stage.

Routine LSM compaction is now intentionally outside the pipeline's scope; it stays byte-for-byte lossless. This removes the entire merge-time complexity (the `merge_grace` per-trace maturity gate, the in-merge filter hook in `mergeBlocks`, the lossy-merge correctness rules) — none of it was load-bearing once stage filtering is anchored to migration boundaries (a segment is already ≥ stage TTL old at migration, so completeness concerns are already satisfied without a separate merge-time grace).

**Proto deltas** vs the previous PR push:
- Removed `enum StageEvent`, `StageRule.apply_at`, `TracePipelineConfig.merge_grace`.
- Renumbered `StageRule` fields (`stage`=1, `min_duration`=2, `keep_errors`=3, `keep_tag_rules`=4) and `TracePipelineConfig.finalize_grace` (7).
- `buf lint` + `buf build` pass.

**Doc deltas:**
- §5 collapses from 3 subsections to 2 (`§5.1` Migration + `§5.2` Eviction); the former §5.1 Compaction Rewrites is gone.
- §7 collapses from 3 subsections to 2 (`§7.1` Pre-Migration Rewrite + `§7.2` Scheduled Finalization); the former §7.1 LSM Compaction Merge Phase Loop is gone.
- §1.2, §2.1, §2.2, §2.3 (tuple now `(Group, Schema, Stage)`, no `StageEvent`), §2.4 (no `merge_grace` bullet), §3.1 (no "Provisional during LSM merge" bullet), §4.2, §6.x JSONs (no `apply_at` lines, no `merge_grace`), §7.3-now-§7.2 (single-pass tail_sampling, no FINALIZE-targeted StageRule composition), §8 step-by-step (gating then migration-out only), and backlog all updated.
- Section numbering renumbered to stay contiguous: §1-§7 with no gaps.

**Net diff:** 3 commits, 4 files, +876 / 0 vs `apache/main` (down from +971 — ~95 lines removed by the simplification). `make generate` / `buf lint` / `buf build` all pass.

3 commits on the PR:
- `767588e7` Add post-trace pipeline design (per-stage retention + tail-sampling)
- `c8ea4da2` Add trace_pipeline.proto for banyandb.pipeline.v1
- `ca0e9d4d` Preserve deferred weighted-scoring model under design/backlog

hanahmily:
Force-pushed another iteration: **`finalize` and `merge` events are now independently toggleable via `enabled_events`**, with `[PIPELINE_EVENT_MERGE]` as the default. The in-merge filter (with per-trace `merge_grace`) is restored as `PIPELINE_EVENT_MERGE`; the finalization gate (with `finalize_grace`) is `PIPELINE_EVENT_FINALIZE` and opt-in. Both share the same `tail_sampling` keep-rules — one configured set, evaluated wherever events fire.

**Proto change (Option 3 from the design discussion):**

```protobuf
enum PipelineEvent {
  PIPELINE_EVENT_UNSPECIFIED = 0;
  PIPELINE_EVENT_MERGE       = 1;
  PIPELINE_EVENT_FINALIZE    = 2;
}

message TracePipelineConfig {
  // ...
  TailSampling tail_sampling                = 6;  // shared keep-rules
  repeated PipelineEvent enabled_events     = 7;  // default = [MERGE]
  google.protobuf.Duration merge_grace      = 8;  // used iff MERGE enabled
  google.protobuf.Duration finalize_grace   = 9;  // used iff FINALIZE enabled
}
```

Why these defaults: out-of-box behavior is the in-merge filter (with deferred drops via `merge_grace`) — cheap, runs often, reclaims space incrementally during routine LSM compaction. Operators who want the heavier per-segment-after-settle gate explicitly add `PIPELINE_EVENT_FINALIZE` to the list. Per-stage retention (`StageRule`) is always implicit at migration-out when a StageRule carries a predicate, and is not in the enum.

**Doc deltas:**

- Restored §5.1 *Compaction Rewrites* and §7.1 *LSM Compaction Merge Phase Loop* (gated on `PIPELINE_EVENT_MERGE`). Both describe the per-trace `merge_grace` gate that prevents partial-trace drops during compaction.
- §1.1 / §1.2 reframed: the gate fires at whichever events are enabled; routine compaction is now governed by the toggle, not "always lossless."
- §2.1 paragraph: three anchored events (merge, finalize, migration-out) instead of two.
- §2.2 message summary: `TracePipelineConfig` bullet lists `enabled_events`, `merge_grace`, `finalize_grace`; new `PipelineEvent` bullet describes the enum.
- §2.4 admission: updated to handle the default = `[MERGE]` semantics (config with `tail_sampling` set is valid without explicit `enabled_events`).
- §3.1 retention timing: bullets now cover all three filter points (merge / finalize / migration-out).
- §6.1 / §6.2 scenarios: scenario 1 explicitly sets `enabled_events: [PIPELINE_EVENT_MERGE, PIPELINE_EVENT_FINALIZE]` and includes both graces; scenario 2 relies on the default to demonstrate the minimal config form.
- §7 data-flow diagram: three pathways again; the conditional gating is called out per entry.
- §8 step-by-step: three phases (A in-merge, B finalization, C migration-out) walking the same real trace through each.

`buf lint`/`buf build` and `make generate` both pass. **3 commits, +991/0 vs `apache/main`** (up ~115 lines from the previous push, mostly the restored §5.1 and §7.1 chapters).

PR #1145, opened 2026-05-29
Fix lifecycle migration of non-multiple segment intervals via row-replay

mrproliu:
## Why

  The lifecycle migration moves expired data from one stage to the next (hot → warm → cold). Each stage may define a **different `segment_interval`**. The original migration copied whole part files into
   a single target segment chosen from the part's timestamp (the "chunk-sync" path). That is only correct when every source part falls entirely inside one target segment.

  When the source and target segment intervals are **not multiples of each other** (their day-grids don't align — e.g. a 5-day source interval migrating into 7-day target segments), a single source
  part's rows can **straddle a target segment boundary**. Chunk-copying such a part dumps **all** of its rows into one target segment, so the rows that belong to the adjacent segment end up in the wrong
   place: the migrated part's `[minTimestamp, maxTimestamp]` escapes its target segment's `[start, start+interval)` bounds. This produces mis-segmented data and breaks per-segment assumptions for later
  queries and retention.

  ## What

  Detect source parts that cross a target segment boundary and migrate them with **row-level replay** instead of a file copy:

  - **Part fully inside one target segment** → unchanged fast **chunk-sync** (copies the part file).
  - **Part spanning ≥2 target segments** → **row-replay**: re-publish each row so the receiving node routes it to the target segment that matches the row's own timestamp. No row lands in the wrong
  segment.

  ### Supporting changes
  - **Per-part flush-and-confirm**: each migrated part is flushed and its server-side acks confirmed before it is marked complete (the batch publisher is closed/reopened per part), so a part is only
  recorded as migrated once it is durably delivered; on failure it is recorded as an error instead of silently lost.
  - **Source-segment completion key**: migration progress is keyed on the *source* segment instead of the target, fixing a collision where multiple source segments mapping to one target segment could be
   wrongly treated as already-migrated.
  - **`raw_element_id` consistency (stream)**: a new `raw_element_id` is carried through row-replay so an element migrated via both chunk-sync and row-replay keeps one storage element id; the warm-stage
   query dedup then converges to a single row instead of duplicating it.
  - **Migration report** now breaks down `chunk_sync` vs `row_replay` parts/rows per group so the path taken is observable.

  ## Tests
  - New `row_replay_*` unit tests for the per-type replayers and progress tracking.
  - Distributed e2e (`TestPropertyLifecycle`): added measure / stream / trace cross-segment specs that seed boundary-straddling data, run the migration, and verify (a) the **physical** layout — every
  migrated part stays within its own target segment, segments split correctly, exact segment count + contiguous boundaries — and (b) the **gRPC query** results from the warm stage (per-timestamp,
  same-segment, and cross-segment aggregate). Stream specs also assert `raw_element_id` dedup; trace adds a `cross_segment_timestamp` index rule so warm-stage queries are verifiable.
  - `make lint`, lifecycle unit tests, and the distributed e2e (5/5) all pass.

### DIFF
--- a/banyand/backup/lifecycle/row_replay_measure.go
+++ b/banyand/backup/lifecycle/row_replay_measure.go
@@ -1,0 +1,336 @@
+// Licensed to Apache Software Foundation (ASF) under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Apache Software Foundation (ASF) licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package lifecycle
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"strconv"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"github.com/apache/skywalking-banyandb/api/common"
+	"github.com/apache/skywalking-banyandb/api/data"
+	commonv1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/common/v1"
+	databasev1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/database/v1"
+	measurev1 "github.com/apache/skywalking-banyandb/api/proto/banyandb/measure/v1"
+	"github.com/apache/skywalking-banyandb/banyand/internal/dump"
+	dumpmeasure "github.com/apache/skywalking-banyandb/banyand/internal/dump/measure"
+	"github.com/apache/skywalking-banyandb/banyand/metadata"
+	"github.com/apache/skywalking-banyandb/banyand/metadata/schema"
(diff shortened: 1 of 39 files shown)


banyand/backup/lifecycle/row_replay_measure.go
```diff
@@ -0,0 +1,336 @@
...
+	r.irCacheMu.Lock()
+	defer r.irCacheMu.Unlock()
+	if ir, ok := r.irCache[segmentPath]; ok {
+		return ir, nil
+	}
+	ir, err := dump.NewIndexResolver(segmentPath, dump.DefaultIndexCacheSize, r.mergedRuleToTag)
+	if err != nil {
+		return nil, fmt.Errorf("open index resolver for %s: %w", segmentPath, err)
+	}
+	r.irCache[segmentPath] = ir
```
hanahmily: `irCache` accumulates one `dump.IndexResolver` (bluge index reader + its cache) per source segment and only releases them in `Close()`, which for the measure/stream replayers is the whole group migration. A migration that spans many source segments therefore keeps every segment's index reader open simultaneously, so peak memory grows with the number of cross-segment source segments rather than staying bounded.

Since segments are visited in order and an `IndexResolver` is only needed while that segment's parts are being replayed, could the resolver for a segment be closed/evicted once its last part has been replayed (or capped with a small LRU)? That would keep resident memory bounded regardless of migration span. The same applies to `streamRowReplayer.irCache`.

(Not a correctness issue — the replayers run against snapshot dirs, so there's no live-serving conflict — purely a resource-footprint concern on large-group migrations.)

hanahmily (changes requested):

banyand/backup/lifecycle/row_replay_measure.go
mrproliu: Since all shards should walk from the same segment, I delete the map reference to a single reference to reduce the memory usage.

hanahmily (approved):

PR #1154, opened 2026-06-04
Fix migration failure when row replay from large data

mrproliu:
## Why

  Tiered-storage lifecycle migration falls back to **row-replay** (re-publishing each row through the Write API) for any source part whose segment spans more than one target segment — the case for non-multiple/coprime stage intervals (e.g.
  `sw_metricsHour` hot **5d** → warm **7d**, where a `[05-27, 06-01)` segment straddles the warm `05-28` boundary).

  For large parts this failed with:

  ```
  code = DeadlineExceeded desc = context deadline exceeded
  ... confirm row-replay measure part .../seg-20260527/shard-0/...: 1 node error(s)
  file-based measure migration failed ... measure parts incomplete
  ```

  **Root cause:** row-replay sent the *entire part* on a single client-streaming batch publisher. That stream's context deadline (the `30s` batch timeout) was set once at stream-open and had to cover the whole part's send. A real part of
  **425,940 rows** takes ~42s to build/route/marshal/publish, so the single 30s window expired mid-part → the receiver returned `DeadlineExceeded` → the part was marked incomplete and the group migration "partially completed", retrying
  with the same failure every cycle.

  ## Solution

  Reworked row-replay so the timeout is **per batch**, not per part, and extracted the machinery into one shared layer:

  - **Per-batch timeout window** — each 2000-row batch is published on its *own* publisher with a fresh timeout and rotated in immediately. No single deadline ever spans the whole part, so arbitrarily large parts complete.
  - **Bounded confirm pipeline (depth 2)** — a batch's receiver-side confirmation overlaps with building/sending the next batch (double-buffering), cutting wall time while capping in-flight memory.
  - **Shared `batchSender` + `confirmPipeline`** — the per-type replayers (measure/stream/trace) now just build messages and enqueue; the publisher lifecycle (open/rotate/close), bounded overlap, and confirmation ordering live in one place
  (removes ~350 lines of triplicated logic).
  - **Standard `error` instead of a custom outcome struct** — replayers return a plain `error` (`*nodeReplayError`) that still distinguishes per-node delivery failures (`cee`) from global send/build failures.
  - **Per-part timing instrumentation** — each part logs `build_send` vs `confirm_wait` to pinpoint sender- vs receiver-bound migrations.

  On abort, in-flight batches are still drained and the un-flushed residual is discarded, so a failed part is re-replayed whole on resume (no duplicate/partial commits).

  ## Validation

  Restored real production data and ran the migration end-to-end:

  - `sw_metricsHour` hot→warm: **23 parts row-replayed successfully, 0 `DeadlineExceeded`**, incl. part `5820` (**425,940 rows, ~41.8s**) and part `17658` (**484,798 rows, ~61.5s**) — both far beyond the old 30s limit.
  - Timing showed `build_send ≈ 41.7s` vs `confirm_wait ≈ 0.1s`, i.e. sender-bound (the pipeline keeps the receiver from being the bottleneck).

  ## Tests

  Added unit tests covering the pipeline contract: in-flight depth bound, per-batch publisher/timeout rotation, every-row-sent across batch boundaries + tail, abort-drains-and-discards (build error / iterator error / node error),
  partial-failure reporting, and the `error` type's message/unwrap behavior. `make lint` and the package `-race` suite are green.


banyand/backup/lifecycle/row_replay_pipeline.go
```diff
@@ -0,0 +1,357 @@
...
+// captured publisher and channel.
+type batchSender struct {
+	client    queue.Client
+	publisher queue.BatchPublisher
+	pipeline  *confirmPipeline
+	batch     []bus.Message
+	topic     bus.Topic
+	timeout   time.Duration
+	batchSize int
+	mu        sync.Mutex
```
hanahmily: `batchSender` is documented as "driven by a single goroutine" yet carries a `sync.Mutex`. The replay loop calls `take()`/`enqueue()` exclusively, and background confirm goroutines only touch their own captured `pub`/`ch` — they never access `s.batch`. The mutex is not protecting against any actual concurrent access.

Either remove it (and `take()`'s lock/unlock), or add a comment explaining what concurrent scenario it guards against.

banyand/backup/lifecycle/row_replay_pipeline.go
```diff
@@ -0,0 +1,357 @@
...
+}
+
+// newBatchSender builds a sender for one data type. batchSize and timeout are
+// per-type knobs each replayer (measure/stream/trace) supplies from its own
+// constants; they coincide at 2000/30s today, which is why unparam is silenced.
+//
+//nolint:unparam
```
hanahmily: Suppressing `unparam` because all three callers today pass the same `batchSize`/`timeout` constants is a code smell. If the parameters are truly always the same, make them constants inside `newBatchSender` and drop the parameters entirely. If the intent is to support different values per type in the future, a brief `// reserved for per-type tuning` comment is clearer than a linter suppression that silently hides the fact.

banyand/backup/lifecycle/row_replay_pipeline.go
```diff
@@ -0,0 +1,357 @@
...
+		cee, closeErr := pub.Close()
+		err := pubErr
+		if err == nil {
+			err = closeErr
+		}
+		out = newNodeReplayError(cee, err)
+	}()
+	// Abort promptly: an earlier in-flight batch's failure (via the bound) wins,
+	// otherwise this batch's own synchronous send error.
+	if awaited := s.pipeline.add(ch); awaited != nil {
```
hanahmily, line 219: When `Publish` fails (`pubErr != nil`), the receiver never saw the batch, so `cee` from the subsequent `pub.Close()` is meaningless — the close-side per-node errors reflect rows the receiver processed, which is none here. Passing a potentially non-nil `cee` alongside `pubErr` to `newNodeReplayError` can produce a confusing error that reports both a global send failure and node-specific rejections.

Consider zeroing `cee` when `pubErr != nil`:
```go
err := pubErr
if err == nil {
    err = closeErr
} else {
    cee = nil // publish failed — receiver saw nothing, per-node cee is noise
}
out = newNodeReplayError(cee, err)
```

banyand/backup/lifecycle/row_replay_trace.go
```diff
@@ -141,40 +112,8 @@ func (r *traceRowReplayer) replayPart(ctx context.Context, partPath string) (int
...
-			Int("row_idx", pos.RowIdx).
-			Int("rows_published", rowCount).
-			Msg("trace row-replay aborted mid-part; will retry on resume")
-		return rowCount, iterErr
-	}
-	if r.counter != nil {
-		atomic.AddUint64(r.counter, 1)
-	}
-	return rowCount, nil
+	return r.sender.replay(ctx, r.logger, r.group, partPath, r.counter, it,
```
hanahmily: The old per-type abort log included `Str("trace", r.traceName)`. The shared `warnAbort` closure in `batchSender.replay()` only logs `group` and `part`, so this field is silently dropped for trace replays. If monitoring queries or alerts filter on `trace=`, this is a quiet observability regression.

Consider embedding the trace name in the `part` label (e.g. `r.traceName+"/"+partPath`) or adding an optional label parameter to `replay()`.

banyand/backup/lifecycle/row_replay_test.go
```diff
@@ -24,15 +24,18 @@ import (
...
 	"path/filepath"
 	"strconv"
+	"sync/atomic"
 	"testing"
 	"time"
 
 	"github.com/onsi/gomega"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	gomock "go.uber.org/mock/gomock"
```
hanahmily: Project convention uses explicit aliases only for the project's protobuf packages. `go.uber.org/mock/gomock` is already named `gomock` by its package declaration — the alias is redundant. Use the bare import:
```go
"go.uber.org/mock/gomock"
```

hanahmily (changes requested):
Thank you for this well-structured fix. The core change (per-batch publisher rotation so no single stream deadline spans the whole part) is correct, and the extracted `batchSender`/`confirmPipeline` layer is a clean refactor. Five items to address below — one is a correctness edge case, the rest are observability, style, and maintenance.

**Required before merge:** Please update `CHANGES.md` — the PR description has this as an unchecked checkbox. For a bug fix that affected production migrations, a changelog entry is important for operators tracking what changed between releases.

banyand/backup/lifecycle/row_replay_pipeline.go
mrproliu: removed mutex.

banyand/backup/lifecycle/row_replay_pipeline.go
mrproliu: since the const may changed from each data type, so keep the parameter and remove the `nolint` lable.

banyand/backup/lifecycle/row_replay_pipeline.go
mrproliu: done

banyand/backup/lifecycle/row_replay_trace.go
mrproliu: done

banyand/backup/lifecycle/row_replay_test.go
mrproliu: done

mrproliu:
> Required before merge: Please update CHANGES.md — the PR description has this as an unchecked checkbox. For a bug fix that affected production migrations, a changelog entry is important for operators tracking what changed between releases.

Row replay is the new feature in this milestone, and it already added in the changes.md, so no need to re-add again.

hanahmily (approved):

PR #1179, opened 2026-06-16
fix: remove mcp from release signing process

ButterBright:


wu-sheng:
Why do you remove the mcp from release sign?

ButterBright:
MCP is packaged and signed along with BanyanD, the signing process fails on my machine when that line is added.

hanahmily (approved):

PR #1184, opened 2026-06-19
feat: implement global tag type metadata for stream parts

ButterBright:
# Stream Global Tag Type Design

## Goal

Add a part-level `tag.type` metadata file to stream storage. The file records the tag types present in a stream part while the existing type field in
each block's tag metadata remains unchanged.

This change prepares stream storage to access tag types globally in a future query-path update. The current query and block-decoding paths will continue
to use the existing block metadata.

## Scope

The change covers:

- in-memory part creation and file flushing;
- file-part creation during merge;
- deterministic serialization of the global tag type metadata;
- part synchronization and handoff so remote parts retain the new file;
- unit tests for serialization, writing, merging, and synchronization.

The change does not cover:

- reading `tag.type` when opening a part;
- using global tag types during queries;
- removing `valueType` from `tagMetadata`;
- resolving conflicting types for the same family and tag name.

## Data Model And File Format

Each stream part may contain a `tag.type` file. The in-memory representation is hierarchical so tag family and tag name are not concatenated:

```go
type tagType map[string]map[string]pbv1.ValueType
```

The outer key is the tag family name. The inner key is the tag name.

The binary format is:

1. tag family count;
2. for each tag family, its encoded name and tag count;
3. for each tag, its encoded name and one-byte `pbv1.ValueType`.

Families and tags are sorted by name before serialization. This makes the output deterministic without relying on Go map iteration order. Names are
length-prefixed with the repository's existing encoding helpers, matching the conventions used by trace `tag.type` metadata.

If no tag types were collected, no `tag.type` file is written. Type conflicts are currently outside scope; assigning an already known family/tag key
replaces its previous value.

## Write Path

`blockWriter` owns the part-level collector because all blocks written to a part pass through it. As each block is written, the writer visits its tag
families and records each tag's `valueType` in the hierarchical map.

`blockWriter.Flush` exports the collected map together with `partMetadata`. The existing `tagMetadata.marshal` output is not changed, so every block
continues to carry its original type information.

`memPart` stores the collected map. When it is flushed to disk, it writes `tag.type` atomically after the core data files, following the trace
implementation's durability pattern. Reset and pool reuse clear the map to prevent metadata leaking between parts.

## Merge Path

The stream merge path must create a new `tag.type`; copying an input file would be incorrect because a merged part may contain tags from multiple source parts.

Merge already decodes source blocks and rewrites them through `blockWriter`. The same collector therefore observes every tag written to the merged
destination. `mergeBlocks` returns both `partMetadata` and the collected `tagType`, and `mergeParts` writes the resulting `tag.type` before opening the
new part.

No conflict detection or renaming is added. If repeated occurrences of the same family/tag use different types, the last observed type is stored
globally while each block retains its exact original type.

## Synchronization

Part synchronization must include `tag.type`; otherwise a part written on a receiving node would differ from its source.

The sender adds the file to the chunked part file list when it exists and is non-empty. The receiver recognizes the `tag.type` stream, buffers its
chunks, and flushes it atomically during part finalization. Cleanup clears the buffer on success and failure.

This applies to the existing stream synchronization path used for transferring file parts. It does not make the receiver or query path parse the file.

## Error Handling And Durability

Serialization returns errors for truncated names or missing type bytes. Filesystem read and write behavior follows the trace helpers and existing
stream panic-on-storage-corruption conventions.

`tag.type` is written atomically. The final atomic metadata write preserves the existing directory durability ordering. Synchronization finalizes core
writers before writing sidecar metadata.


hanahmily (commented):
A few additional notes beyond the automated review (these are separate from the points Copilot raised).

**Test coverage (serializer):** Beyond the chunked-sync receiver gap noted elsewhere, the serializer itself lacks direct coverage — there's no standalone `marshal`↔`unmarshal` round-trip test, and no negative/truncation test. The current tests only validate the happy path indirectly (reading the file back after write/merge). A small table test exercising round-trip plus a few truncated/corrupt inputs would lock in both the format and the error-handling behavior.

Nice, well-scoped change overall — it mirrors the trace `tag.type` precedent closely and correctly leaves the read path untouched.

banyand/stream/block_writer.go
```diff
@@ -173,6 +174,7 @@ type blockWriter struct {
 
 func (bw *blockWriter) reset() {
 	bw.writers.reset()
+	bw.tagType.reset()
```
hanahmily: `reset()` calls `bw.tagType.reset()` (i.e. `clear()`) directly, without guarding against a nil map. This is safe today only because `generateBlockWriter` always initializes `tagType`, but `copyFrom` would panic if a `blockWriter` ever reached use with a nil map. The merged trace implementation guards this — could we match it here for consistency and robustness?

```go
if bw.tagType == nil {
    bw.tagType = make(tagType)
} else {
    bw.tagType.reset()
}
```

banyand/stream/part.go
```diff
@@ -232,6 +239,7 @@ func (mp *memPart) mustFlush(fileSystem fs.FileSystem, path string) {
 	}
 
 	mp.partMetadata.mustWriteMetadata(fileSystem, path)
+	mp.tagType.mustWriteTagType(fileSystem, path)
```
hanahmily, line 241: `tag.type` is written *after* `metadata.json` here, which raises two points:

- The comment just below ("…covering the dirent changes for all data files written above") is now slightly stale, since `metadata.json` is no longer the final write.
- `metadata.json` is the part's completion marker, so a crash between the two writes yields a "complete" part with no `tag.type`. Harmless now (read path is out of scope), but the future query-path reader will need to tolerate a missing `tag.type`.

Note also that `FinishSync` in `write_data.go` writes them in the opposite order (`tag.type` before `metadata.json`), so the ordering is inconsistent between the flush/merge paths and the sync path. Worth aligning, or at least documenting the intended invariant.

banyand/stream/part_metadata.go
```diff
@@ -107,6 +112,96 @@ func (pm *partMetadata) mustWriteMetadata(fileSystem fs.FileSystem, partPath str
...
+		sort.Strings(tagNames)
+		for _, tagName := range tagNames {
+			dst = encoding.EncodeBytes(dst, convert.StringToBytes(tagName))
+			dst = append(dst, byte(tags[tagName]))
+		}
+	}
+	return dst
+}
+
+func (tt tagType) unmarshal(src []byte) error {
```
hanahmily, line 159: The PR description states this "returns errors for truncated names or missing type bytes," but the decode is only partially defensive: `encoding.BytesToVarUint64` and `DecodeBytes` return `0`/empty *without* an error on several truncation shapes. A corrupt `familyCount`/`tagCount` would therefore spin the loop (inserting empty entries) or panic in `make(..., tagCount)` rather than returning a clean error. Since `unmarshal` is test-only today this is latent, but before the read path lands I'd suggest bounds-checking the decoded counts against `len(remaining)` so corruption fails fast.

(This is the same function Copilot flagged, but a distinct concern — its compile claim is incorrect, as `uint64` is a valid `make` size hint.)

banyand/stream/part.go
```diff
@@ -422,6 +430,18 @@ func CreatePartFileReaderFromPath(partPath string, lfs fs.FileSystem) ([]queue.F
 		if e.IsDir() {
 			continue
 		}
+		if e.Name() == tagTypeFilename {
```
hanahmily, line 433: Minor: this `if e.Name() == tagTypeFilename` branch doesn't `continue` afterward, so control falls through to the `filepath.Ext` checks below. It's harmless (`.type` matches none of those extensions), but adding `continue` would match the structure of the sibling branches and avoid the redundant comparisons.

ButterBright: This is not necessary and the sibling branches don't have `continue`.

banyand/stream/part_metadata.go
ButterBright: Perhaps we should fix it by revising `BytesToVarUint64` in a future PR.

hanahmily (approved):

PR #1195, opened 2026-06-27
Fix block metadata reset before unmarshal for stream and trace

ButterBright:


hanahmily:
@ButterBright, fix the comments from Copilot.

hanahmily (approved):

PR #1196, opened 2026-06-27
feat: implement global tag type metadata for measure parts

ButterBright:


banyand/measure/merger.go
```diff
@@ -261,6 +261,7 @@ func (tst *tsTable) mergeParts(fileSystem fs.FileSystem, closeCh <-chan struct{}
 		return nil, err
 	}
 	pm.mustWriteMetadata(fileSystem, dstPath)
+	mergedTagType.mustWriteTagType(fileSystem, dstPath)
```
hanahmily, line 263: Nit: write ordering is inconsistent with `memPart.mustFlush`, which writes `tag.type` *before* `metadata.json`, whereas here it goes *after*. Functionally harmless since `WriteAtomic` fsyncs the parent dir on each rename, but worth aligning the two paths (and it keeps the "covers the dirent changes for all data files written above" comment accurate) — suggest writing the tag type before `mustWriteMetadata` here too.

hanahmily (approved):

PR #1198, opened 2026-06-30
feat(trace): dynamic sampler registration via group-scoped pipeline config

hanahmily:
### Dynamic sampler registration via group-scoped pipeline config

Register/update/remove in-merge trace retention samplers at runtime **without restarting the data node**, replacing the static `--trace-pipeline-config` boot loader added in #1194 (now removed). This is the dynamic follow-up that the static in-merge filter PR explicitly deferred.

#### How it works
- The sampler pipeline config is carried on `common.v1.Group` (new optional `pipeline` field). `TracePipelineConfig`/`SamplerPlugin` were relocated from `pipeline/v1` into the leaf `common/v1` to avoid a `common`→`pipeline` proto import cycle.
- Config is delivered through the **existing `KindGroup` schema watch** (property-backed registry) — no new RPC, no etcd.
- The data node rebuilds its group-keyed sampler registry on `OnAddOrUpdate(KindGroup)` / `OnDelete(KindGroup)`, **fail-open** and gated to `ROLE_DATA` + native-pipeline-enabled.
- Plugin cache keyed by `(path, symbol, configHash)` with an ABI re-check on every miss; retired sampler instances are **never `Close`d** (Go plugins are immortal).
- Per-group `merge_grace`; `validate.GroupForNonProperty` validates the embedded config and rejects stage-only configs.
- Metrics: `sampler_{active_count,register,update,remove,load_failed}`.

#### Notes
- **Dynamic-only**: the static `--trace-pipeline-config` flag + boot loader are removed (`--trace-pipeline-native-plugin-enabled` / `--trace-pipeline-trusted-plugin-dir` retained).
- Generated `*.pb.go` are gitignored; run `make generate` after checkout.

- [x] Tests (including UT, IT, E2E) are added to verify the new feature. 


hanahmily:
Thanks @copilot — addressed 4 of the 5 in `37284ce1` (CI green):

- **#1 `BatchWaitSec` truncation** — `schemaserver` now derives the index batch wait via `batchWaitSeconds()` (ceil-to-seconds, floored at 1), so a sub-second `--schema-server-flush-timeout` yields ≥1s instead of 0.
- **#2 `enabled_events` ignored** — `reconcilePipeline` now installs merge samplers only when `PIPELINE_EVENT_MERGE` is enabled (empty list defaults to MERGE for back-compat); added `TestReconcilePipeline_MergeEventNotEnabled` locking the FINALIZE-only → no-samplers behavior.
- **#3 `sampler_load_failed` cardinality** — the `reason` label now maps to a small, stable set of codes (`abi_mismatch`, `path_escape`, `open_failed`, `ctor_failed`, `symbol_missing`, `config_marshal`, `config_invalid`, `panic`, `load_error`); the full error is still logged.
- **#4 hard-coded `abi_version == 1`** — the validator now compares against `sdk.ABIVersion` (single source of truth).

**#5 `TracePipelineRegistryService` shape** — leaving as-is for now. This service is currently unwired (no `Register…Server` call) and is an intentional planned-API stub (the doc comment references §2.3/§2.4 of the schema-consistency design). This PR delivers config via the group-scoped `common.v1.Group.pipeline` field over the existing `KindGroup` watch rather than this RPC, so reconciling the Create/Update (name-less) vs Get/Delete (`{group}/{name}`) shape is a deliberate API decision better made when/if the RPC is actually implemented. Happy to follow up separately if maintainers prefer to either drop the service or make it group-scoped now.

mrproliu (approved):

PR #1201, opened 2026-07-01
Add the codex/claude plugin to bydbQL

JophieQu:
### What this PR does
Adds a Claude/Codex plugin that packages the BanyanDB MCP server so a host model
can generate, validate, and run **read-only** BydbQL over STREAM/MEASURE/TRACE/PROPERTY.
It ships:
- the `bydbql` skill (`skills/bydbql/`) with syntax/examples/safety references;
- MCP tools `list_groups_schemas`, `get_generate_bydbql_prompt`, `validate_bydbql`, `list_resources_bydbql`;
- a parse-only Go validator (`mcp/tools/bydbql-parse`) reused by `validate_bydbql`;
- plugin manifests (`.claude-plugin/`, `.codex-plugin/`) + `.mcp.json`.

### Install / build
    cd mcp
    npm install
    npm run build:all   # tsc dist/ + Go validator binary tools/bin/bydbql-parse
Load the plugin from the repo root. See docs/operation/mcp/plugin.md.

### Safety
Two-layer safety: TS read-only first gate (validateBydbQL: SELECT/SHOW TOP only,
rejects DELETE/;/-- comments/over-length) + Go parse-only validation. The skill
promotes get_generate_bydbql_prompt for any sorting/ranking so ORDER BY only uses
indexed fields (validate_bydbql is parse-only and can't catch that).

### Tests / CI
- Go: mcp/tools/bydbql-parse/main_test.go + testdata/{valid,invalid}.txt
- TS (vitest): src/query/validation.test.ts (safety layer) + parse-validator.test.ts (spawn integration)
- Gated in CI via `make -C mcp test` in the `check` job.
- NL→BydbQL generation quality is validated out-of-band (not a PR gate).

- [x] Update the CHANGES log.


mcp/src/query/parse-validator.ts
```diff
@@ -0,0 +1,148 @@
...
+  args: string[];
+  cwd: string;
+};
+
+function mcpRootDir(): string {
+  const modulePath = fileURLToPath(import.meta.url);
+  return path.resolve(path.dirname(modulePath), '..', '..');
+}
+
+function validatorCommands(rootDir: string): ValidatorCommand[] {
```
hanahmily: **[major] `validate_bydbql` shells out to `go run` at query time — the Go toolchain becomes a runtime dependency.** If `tools/bin/bydbql-parse` isn't prebuilt, every validation call runs `go run ./tools/bydbql-parse`, which (a) requires a Go toolchain in the plugin's *runtime* environment and (b) does a full compile — easily exceeding the 10s timeout on a cold cache, producing a spurious "validation timed out".

Could we build the binary at install time (nothing in `.claude-plugin/plugin.json` / `.codex-plugin/plugin.json` currently runs `build:validator`) and have the fallback fail with an actionable message ("run `npm run build:validator`") instead of silently invoking `go run`? The same class of issue affects `.mcp.json` pointing at `./mcp/dist/index.js` while `dist/` is git-ignored — the server won't start unless `tsc` build ran first.

skills/bydbql/SKILL.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Tool Use
+
+Use the existing BanyanDB MCP tools:
+
+- `list_groups_schemas`: discover groups and resource names for streams, measures, traces, and properties.
+- `validate_bydbql`: run read-only safety checks and parse-only BydbQL syntax validation.
+- `list_resources_bydbql`: execute a validated BydbQL statement when data retrieval is requested.
+
+The Codex host model should perform natural language to BydbQL translation using these references and live schema discovery. Use `get_generate_bydbql_prompt` only when its live schema prompt text would materially help.
```
hanahmily: **[major] Demoting `get_generate_bydbql_prompt` drops the only index-rule `ORDER BY` safety.** That prompt (`mcp/src/query/llm-prompt.ts`) is the only component that injects the indexed-field list and enforces field-similarity substitution/omission for `ORDER BY`. `validate_bydbql` is parse-only, so a host model following this skill can emit `ORDER BY <non-indexed-field>` that validates `true` and then fails at execution.

Recommend promoting `get_generate_bydbql_prompt` to a normal step whenever ranking/sorting is involved (or whenever schema was already fetched), or otherwise carrying the index-rule check into the skill flow.

skills/bydbql/references/syntax.md
```diff
@@ -0,0 +1,374 @@
...
+
+Use this reference to generate one read-only BanyanDB Query Language statement from natural language. Keep dynamic names such as groups, resource names, tags, fields, and sortable fields aligned with live BanyanDB schema discovered through MCP.
+
+## Core Rules
+
+- Statement forms: `SELECT ... FROM ...` or `SHOW TOP ... FROM MEASURE ...`.
+- Resource types: `STREAM`, `MEASURE`, `TRACE`, `PROPERTY`.
+- `FROM <RESOURCE_TYPE> <resource_name> IN <group>[, <group>...]` is required.
+- Parentheses around group lists are optional: `IN default, staging` and `IN (default, staging)` are both valid.
+- `TIME` is required for `STREAM`, `MEASURE`, `TRACE`, and `SHOW TOP` queries. `PROPERTY` queries do not use `TIME`.
```
hanahmily: **[minor] Doc ↔ grammar inconsistency: "TIME is required".** The grammar makes `TIME` optional (`GrammarSelectStatement.Time` is `@@?`), and `references/examples.md` example #2 omits it — and it parses clean, so `validate_bydbql` returns `valid=true` and execution may then fail. Either add `TIME` to that example, or soften this to "required at execution / strongly recommended" and note the parser doesn't enforce it.

.claude-plugin/plugin.json
```diff
@@ -0,0 +1,37 @@
+{
```
hanahmily, line 1: **[minor] Duplicated manifest — drift risk.** `.claude-plugin/plugin.json` and `.codex-plugin/plugin.json` are ~95% identical (only `version` differs). A short note on how they're kept in sync — or generating one from the other — would avoid divergence over time.

.mcp.json
```diff
@@ -0,0 +1,15 @@
+{
+  "mcpServers": {
+    "banyandb": {
```
hanahmily, line 3: **[minor] Assumes install location.** `cwd: "."` + `args: ["./mcp/dist/index.js"]` assume the plugin is loaded from the repo root. Can we confirm the plugin loader resolves this relative to the plugin dir in a real install (not just the repo checkout)?

hanahmily (commented):
Thanks for packaging the MCP server as a Claude/Codex plugin — the two-layer safety (TS prefix/comment checks + Go parse validation) and the skill references are a solid foundation. Before merge I'd like to resolve a few items; most are inline. Two that can't be anchored to the diff:

**`CHANGES.md` not updated (ASF hygiene).** The template checkbox is unchecked and there's no entry under `## 0.11.0`. Please add one, e.g. under Features: "Add a Claude/Codex plugin packaging the BanyanDB MCP server with a parse-only `validate_bydbql` tool and a BydbQL skill."

**PR description is the unfilled template.** A short summary of what the plugin adds plus how to install/build it (and which CI covers the new Go tool + TS) would help review and future readers.

**Non-issue (documenting so it isn't re-raised):** I checked the unquoted hyphenated identifiers in the docs (`IN datacenter-1`, `sw_trace`) — the lexer `Ident` pattern is `[a-zA-Z_][a-zA-Z0-9_-]*`, so these parse fine. No change needed.

hanahmily (changes requested):
Converting my prior review to a formal request for changes. The two **[major]** items should be addressed before merge:

1. **`validate_bydbql` runtime dependency** — the `go run` fallback + git-ignored `dist/` make the plugin depend on a Go toolchain and a `tsc`/`build:validator` step at runtime, with no build wired into the plugin manifests. (see inline on `mcp/src/query/parse-validator.ts`)
2. **Skill drops index-rule `ORDER BY` safety** — demoting `get_generate_bydbql_prompt` removes the only indexed-field validation; `validate_bydbql` is parse-only, so invalid-but-parseable `ORDER BY` queries pass and fail at execution. (see inline on `skills/bydbql/SKILL.md`)

Also please address the ASF hygiene items (CHANGES.md entry + filled-in PR description) and the minor doc/manifest notes from the previous review. Happy to re-review once these are in.

mcp/tools/bydbql-parse/main.go
```diff
@@ -0,0 +1,90 @@
...
+
+type validateResponse struct {
+	Valid      bool     `json:"valid"`
+	Message    string   `json:"message"`
+	QueryType  string   `json:"queryType,omitempty"`
+	SyntaxOnly bool     `json:"syntaxOnly"`
+	Warnings   []string `json:"warnings,omitempty"`
+}
+
+func main() {
```
hanahmily, line 38: **[blocking][Layer 1+2] Add a unit test for the validator tool.** The validation logic lives inside `main()`, so it isn't callable from a test. Please extract it:
```go
func validate(query string) validateResponse {
    grammar, parseErr := bydbql.ParseQuery(query)
    if parseErr != nil {
        return validateResponse{Valid: false, SyntaxOnly: true,
            Message: fmt.Sprintf("failed to parse BydbQL: %v", parseErr)}
    }
    return validateResponse{Valid: true, SyntaxOnly: true, Message: "valid BydbQL syntax",
        QueryType: queryType(grammar),
        Warnings: []string{"parse-only validation does not verify group, resource, tag, field, or index-rule existence"}}
}
```
Then add a table-driven `main_test.go` covering: valid SELECT per resource type (STREAM/MEASURE/TRACE/PROPERTY) → `Valid=true` + correct `QueryType`; `SHOW TOP` → `QueryType=TOPN`; and rejection of `DELETE`, trailing `;`, `--` comments, two statements, and empty input. This tool is in the root module, so `go test ./mcp/tools/...` runs under the existing `make test` with no CI changes.

**(Layer 2)** Also add a curated `testdata/valid.txt` + `testdata/invalid.txt` corpus read by the test — do not scrape the markdown docs, since they contain placeholders (`<stream_name>`) and clause fragments that intentionally don't parse. The corpus becomes the regression net for future bugs.

mcp/package.json
```diff
@@ -5,6 +5,7 @@
   "type": "module",
   "main": "dist/index.js",
   "scripts": {
```
hanahmily, line 7: **[blocking][Layer 3] No test tooling in `mcp/`.** Please add a test runner so the TypeScript safety layer can be covered:
```jsonc
"scripts": { "test": "vitest run", ... },
"devDependencies": { "vitest": "^2.x", ... }
```
Then add `src/query/validation.test.ts` asserting the first-gate `validateBydbQL` behavior that `validate_bydbql` relies on: accepts `SELECT`/`SHOW TOP`; rejects non-read-only (`DELETE`), `;`/`--` comments, and over-length (>4096) input. These are pure functions — no spawn needed — so the test is fast and hermetic. Please also wire `npm test` into whatever CI runs the `mcp/` lint step.

mcp/src/query/parse-validator.ts
```diff
@@ -0,0 +1,148 @@
...
+  return {
+    ...process.env,
+    GOCACHE: process.env.GOCACHE || path.join(tmpdir(), 'banyandb-mcp-go-build-cache'),
+    GOMODCACHE: process.env.GOMODCACHE || path.join(tmpdir(), 'banyandb-mcp-go-mod-cache'),
+    GOPATH: process.env.GOPATH || path.join(tmpdir(), 'banyandb-mcp-go-path'),
+    GOFLAGS: process.env.GOFLAGS || '-mod=readonly',
+  };
+}
+
+export async function validateBydbQLSyntax(query: string): Promise<BydbQLParseValidationResult> {
```
hanahmily, line 47: **[recommended][Layer 4] Integration test for the spawn path.** Once `vitest` is in place, add a test that runs `validateBydbQLSyntax()` against the built binary (`npm run build:validator` first) to cover the JSON round-trip and binary-vs-`go run` selection, e.g. a valid query returns `{valid:true, queryType:'STREAM', syntaxOnly:true}` (use a ~20s vitest timeout; skip gracefully if the binary is absent). This is also the natural place to assert the timeout/error-to-`{valid:false}` behavior — which ties directly to the cold-`go run` timeout concern in my other comment on this file.

skills/bydbql/SKILL.md
```diff
@@ -0,0 +1,37 @@
...
+3. Read `references/examples.md` when mapping natural language to BydbQL.
+4. Identify the BanyanDB resource type: `STREAM`, `MEASURE`, `TRACE`, or `PROPERTY`.
+5. If the group, resource name, or resource type is missing or ambiguous, use `list_groups_schemas`.
+6. Generate exactly one read-only BydbQL statement.
+7. Call `validate_bydbql` before execution.
+8. Execute with `list_resources_bydbql` only when the user asks to run, query, fetch, show, or inspect raw BanyanDB data.
+9. If `validate_bydbql` returns `valid=false`, fix the query and validate again before any execution.
+10. If execution fails due to a missing group, resource, or schema, use `list_groups_schemas` to discover the correct names and retry only when the correction is clear.
+
+## Tool Use
```
hanahmily, line 24: **[note, non-blocking] Document that generation quality is validated out-of-band.** Since the host model performs NL→BydbQL, its output can't be unit-tested deterministically. Consider a `make eval-bydbql` target (or scheduled job, not PR CI) that runs a curated `{nl_prompt → [acceptable BydbQL]}` corpus through a real model and scores each output with two LLM-free checks: (a) does it parse (reuse the Layer-1 `validate()`), and (b) does it normalize-match an accepted form. Keep it non-blocking so CI stays model-independent.

hanahmily (changes requested):
**Blocking: this PR ships no tests, and the new code is testable without an LLM in the loop.** The skill's only non-deterministic step is NL → BydbQL generation (done by the host model); everything that step depends on — the Go parser tool, its JSON contract, the read-only safety layer, and the spawn/timeout wrapper — is deterministic and belongs under CI. The LLM's generation quality should be validated by a separate, out-of-band eval corpus that is NOT a PR gate.

Requesting the layers below. Layer 1 needs no CI wiring (the tool is in the root Go module, so `make test` picks it up); Layers 3–4 need `vitest` added to `mcp/`.

| Layer | Scope | Gate? |
|---|---|---|
| 1 | Go tool unit test (`main_test.go`) — parse contract, `queryType`, reject writes/`;`/comments/multi-stmt/empty | required |
| 2 | Curated valid/invalid golden corpus (`testdata/`) | required |
| 3 | TS safety-layer unit test (`validateBydbQL`) via vitest | required |
| 4 | TS spawn integration vs the built binary (timeout/fallback) | recommended |
| — | NL→BydbQL eval corpus, model-run, human-reviewed | never in blocking CI |

.mcp.json
hanahmily, line 3: **[minor] Please confirm install-location resolution.** This still uses `cwd: "."` + `args: ["./mcp/dist/index.js"]`. The new `.agents/plugins/marketplace.json` (`source.path: "./"`) suggests the repo root is the plugin root, which would make this correct — but could you confirm the plugin loader resolves `./mcp/dist/index.js` relative to the installed plugin dir (not just a repo checkout)? A one-line note in `docs/operation/mcp/plugin.md` would settle it.

.claude-plugin/plugin.json
hanahmily, line 1: **[minor] Duplicated manifest still unaddressed.** `.claude-plugin/plugin.json` and `.codex-plugin/plugin.json` remain ~95% identical (only `version` differs). Either generate one from the other, or add a short comment/README note on how they're kept in sync, to avoid future drift. Fine as a follow-up if you'd rather not touch it here.

hanahmily (commented):
Thanks for the thorough follow-up — re-reviewed the delta since `32d8ea0`. **All blocking items are resolved and CI-wired**, and I ran the Go validator tests locally (against generated protos): `ok github.com/apache/skywalking-banyandb/mcp/tools/bydbql-parse`.

Confirmed fixed:
- **`go run` runtime dependency** — fallback removed entirely; prebuilt binary required with an actionable error; `tools/bin/` git-ignored. 
- **Skill `ORDER BY` index-rule safety** — `get_generate_bydbql_prompt` promoted to a required step for sort/rank; documented as the only index-rule source.
- **Tests** — `main_test.go` (contract + `queryType` + rejections), curated `testdata/valid.txt`/`invalid.txt` golden corpus, `validation.test.ts` (read-only first gate), and `parse-validator.test.ts` (spawn integration, skips gracefully without Go). Wired via `make -C mcp test` in CI. 
- **Docs** — `TIME` "required" reworded to "optional in grammar / expected at execution"; `examples.md` #2 now carries `TIME`; `CHANGES.md` entry added.

Nice extras too (`build:all`, `npm ci`, the `license-dep` lockfile-backup fix, and the plugin/build docs).

Two **minor** items remain before I approve — see inline. Neither is a big lift.

hanahmily (approved):

PR #1209, opened 2026-07-07
Introduce parameter binding (`?` placeholders) into BydbQL

### LINKED ISSUE #13628: [Feature] Introduce parameters binding into BydbQL
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/skywalking/issues?q=is%3Aissue) and found no similar feature requirement.

### Description

After discussing the use of BydbQL in OAP, rather than using raw YAML requests, to increase code readability, there is a typical feature gap in QL: it lacks parameter binding. 

We need to use QL like this
```sql
SELECT trace_id, message
FROM STREAM logs in group1
TIME > ?
WHERE message MATCH(?);
```
with parameter binding
- idx0 = -30m
- idx1 = `error`
To finalize the query, in order to get rid of QL injection in the product environment.

### Use case

_No response_

### Related issues

_No response_

### Are you willing to submit a pull request to implement this on your own?

- [ ] Yes I am willing to submit a pull request on my own!

### Code of Conduct

mrproliu:
### Introduce positional parameter binding into BydbQL

This PR introduces positional parameter binding (`?` placeholders) into BydbQL so that untrusted values never need to be concatenated into the query string, eliminating QL injection by construction. `bydbql.v1.QueryRequest` gains a `params` field (`repeated model.v1.TagValue`, reusing the existing common value type instead of introducing a new message) whose entries are bound to the `?` placeholders by their order of appearance.

**How it works.** Binding is implemented as a standalone AST pass (`pkg/bydbql/binder.go`) that runs between the parser and the transformer: the lexer gains a `?` token, value nodes in the grammar gain a placeholder branch, and `BindParams` fills the placeholders from the request parameters so that a bound AST is indistinguishable from one parsed with literal values — the 1700-line transformer needs zero placeholder awareness. Since a bound parameter is always a value and can never alter the query structure, injection is impossible by design. As defense in depth, `Transform` rejects any grammar that still contains unbound placeholders, and a reflection-based tripwire test guards that every `@Param` position in the grammar is wired into the binder.

**Supported positions.** Placeholders are accepted in value positions only: TIME values (`str` relative/RFC3339/`now`, or a typed `timestamp` which is validated at bind time), scalar comparisons (`str`/`int`/`null` for all six operators), IN/NOT IN/MATCH/HAVING value lists (`str`/`int`/`null` plus `str_array`/`int_array` which expand in place while the container keeps its literal single-vs-list form), and LIMIT/OFFSET/TOP N counts (`int` only, range-checked to `[0, MaxInt32]`), matching mainstream SQL prepared-statement conventions. Identifier and structural positions (resource/group/stage names, columns, ORDER BY/GROUP BY, MATCH analyzer) remain syntax errors — unlike some SQL implementations where `ORDER BY ?` silently degrades to a no-op, BydbQL rejects it outright. Binding errors (count mismatch, type mismatch, out-of-range) are reported as `InvalidArgument` with the 1-based position of the offending parameter.

**Consumers.** The embedded web UI's BydbQL console gains a parameter input panel (typed rows sent as protojson TagValues), and the error path now surfaces the server's detailed message instead of the bare HTTP status text. The MCP server's `list_resources_bydbql` tool accepts a typed `params` array so LLM-generated queries can pass untrusted values out of band, with the prompt guidance updated accordingly.

BydbQL console with parameter binding: a ? placeholder in the query bound from the typed parameter row below the editor

**Testing.** `pkg/bydbql` grows to 544 specs: a full acceptance matrix (every parameter type x every placeholder path), an equivalence suite asserting that a parameterized query and its literal-valued twin produce byte-identical native query requests through the real transformer (with an `expectError` marker so error cases cannot pass vacuously), parse-time rejection tests for every forbidden position, and injection-payload tests. All six integration test suites (stream/measure/trace/property/topn/tracepipeline) share a new `helpers.ExtractQL` that reads an optional `#!params:` directive from `.ql` files, and each gets an end-to-end `params_bind` case covering a different parameter shape, including `SHOW TOP ?` and `LIMIT ?`. Documentation (`docs/interacting/bydbql.md` §2.6) documents the complete supported/unsupported position tables and type rules.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#13628. 


pkg/bydbql/binder.go
```diff
@@ -0,0 +1,310 @@
...
+)
+
+// BindParams binds positional parameters to the `?` placeholders in the parsed grammar,
+// in their order of appearance in the query text. After a successful call, the grammar is
+// indistinguishable from one parsed with literal values, so the transformer needs no
+// placeholder awareness. Callers must invoke it before Transform even when params is empty,
+// so a query holding unbound placeholders is rejected instead of silently misinterpreted.
+// Parameter positions in error messages are 1-based. On error the grammar may be partially
+// bound and must be discarded; binding is not retryable on the same grammar.
+func BindParams(g *Grammar, params []*modelv1.TagValue) error {
```
hanahmily, line 39: **HIGH — `BindParams` mutates the parsed AST in place, so the parse-once/bind-many reuse this feature is meant to unlock isn't possible. Please track the immutable-bind redesign as a follow-up PR and guard the leak here.**

**Why reuse is the goal.** Positional parameter binding has two payoffs, and this PR banks only one. The first is injection safety (fully delivered). The second — the classic reason prepared statements exist — is amortizing parse/plan cost across many executions of the same query shape with different values: lex + parse + grammar-build happens once, and each request only binds fresh params and transforms. Our own wire contract already models this — `QueryRequest{ query, params }` is a prepared statement — so consumers (high-QPS dashboards re-issuing the same query every refresh, the MCP/LLM loop re-running one query shape with varying values, any client caching parsed queries keyed by query text) will reasonably expect to parse once and rebind. Without reuse we pay full lex+parse on every request and capture only half the feature's value.

**Why it's blocked today.** Binding writes values straight into the grammar nodes and clears the placeholder markers — `bindScalarValue`/`bindTimeValue` do `v.String = … ; v.Param = false` (lines 291, 308), `collectIntSlot` does `*value = … ; *param = false` (line 118), and `expandLists` rewrites the IN/MATCH/HAVING slices via `list.assign(rebuilt)` (line 269). Since `collect` only registers a slot while the marker is set, the parsed AST is one-shot:
- **Rebind the same grammar** → 2nd call collects 0 slots → `parameter count mismatch: query contains 0 placeholder(s) but N parameter(s) provided`.
- **Silent cross-request leak** → if a cached grammar is rebound with empty `params`, the count check passes (0==0), `Transform` sees no unbound placeholders, and the query runs **with the previous request's values still baked in** — no error, wrong/other-tenant results.

**Ask.** This need not block the PR — the injection-safety win is real and self-contained.
1. **This PR:** make the one-shot contract safe and explicit — document the invariant loudly on `Grammar`/`ParseQuery`/`BindParams`, so nobody adds a query cache before reuse lands and trips the silent-leak path.
2. **Follow-up PR:** implement reusable-AST binding so the template is never mutated. Preferred is a positional value **overlay** the transformer reads instead of `GrammarValue` fields (immutable template, true reuse — accepting the transformer gains minimal placeholder awareness); a per-request **deep clone** is the simpler fallback but re-introduces per-request tree allocation, which partly defeats the point. Let's open a tracking issue and link it here.

pkg/bydbql/transformer.go
```diff
@@ -101,6 +101,11 @@ func NewTransformer(registry metadata.Repo) *Transformer {
 
 // Transform transforms a Grammar into a native query request.
 func (t *Transformer) Transform(ctx context.Context, grammar *Grammar) (*TransformResult, error) {
+	// Defense in depth: an unbound placeholder would otherwise transform into an
+	// empty string or a zero count silently instead of failing loudly.
+	if unbound := countUnboundParams(grammar); unbound > 0 {
```
hanahmily: **MEDIUM (perf) — the whole grammar is walked twice per query.** `BindParams` already walked the tree via `collect`; `Transform` then builds a second `binder` in `countUnboundParams` and reruns the same recursion, allocating `[]func(...)`/`listContainer`s it immediately discards — paid on every query including zero-placeholder ones. Give `countUnboundParams` a closure-free counting walk (increment on `Param==true`), or have `BindParams` stamp the grammar "bound" so `Transform` can skip the re-walk. Note this interacts with the reuse redesign — an overlay/clone approach likely removes this walk entirely.

pkg/bydbql/grammar.go
```diff
@@ -319,13 +323,15 @@ type GrammarOrderByWithIdent struct {
 // GrammarLimitClause represents LIMIT clause.
 type GrammarLimitClause struct {
 	Limit string `parser:"@'LIMIT'"`
-	Value int    `parser:"@Int"`
+	Value int    `parser:"( @Int"`
+	Param bool   `parser:"| @Param )"`
```
hanahmily, line 332: **MEDIUM — literal count positions aren't range-checked, so the new `?` path is stricter than a literal.** `BindParams` correctly bounds a bound count to `[0, MaxInt32]` (binder.go:112), but the literal path casts unchecked in the transformer: `uint32(statement.Limit.Value)` / `uint32(statement.Offset.Value)` (transformer.go:184,187) and `int32(statement.N)` / `int32(topn.N)` (transformer.go:590,1565). So `LIMIT ?`=-5 is rejected cleanly, yet `LIMIT -5` wraps to a huge `uint32` and `SHOW TOP 3000000000` wraps negative. Pre-existing, but this PR establishes the correct guard and touches these count fields — good moment to apply the same `[0, MaxInt32]` check to the literal path via a shared helper so the two paths can't diverge.

mcp/src/query/validation.ts
```diff
@@ -119,12 +135,83 @@ export function normalizeQueryHints(args: unknown): QueryHints {
...
+    throw new Error(`params[${position}]: value exceeds ${maxParamValueLength} characters`);
+  }
+  return value;
+}
+
+function validateParamInteger(value: unknown, position: number): string {
+  if (typeof value === 'number' && Number.isSafeInteger(value)) {
+    return String(value);
+  }
+  if (typeof value === 'string' && value.length <= maxIntParamLength && integerPattern.test(value)) {
```
hanahmily, line 161: **MEDIUM — `maxIntParamLength = 20` admits int64-overflowing values, deferring the failure to an opaque gateway error.** A 20-char string like `"99999999999999999999"` passes `integerPattern` + the length cap and is emitted as `{ int: { value: "…" } }`, then fails late in protojson int64 decoding on the server instead of returning a clean `params[i]` message. Range-check against int64 bounds (e.g. a `BigInt` compare) here so the error is client-side and specific.

mcp/src/query/validation.ts
```diff
@@ -119,12 +135,83 @@ export function normalizeQueryHints(args: unknown): QueryHints {
...
+}
+
+/**
+ * Validate BydbQL parameters and convert them to protojson TagValue form
+ * accepted by the BanyanDB HTTP API.
+ *
+ * The server-side `timestamp` variant is deliberately not exposed here:
+ * an RFC3339 `str` is equivalent for TIME positions.
+ */
+export function toTagValueParams(params: BydbQLParam[]): TagValueParam[] {
```
hanahmily, line 190: **MEDIUM — new param validation ships with zero unit tests.** `git diff --stat` shows `validation.ts +90` (this function plus `validateParamString`/`Integer`/`Array` and the new limits) but `validation.test.ts` is untouched, despite the existing MCP test convention (#1036). This is the untrusted, LLM-facing entry point. Add cases for: param-count cap, each type's happy path, type mismatch, empty/oversized array, int overflow (ties to the comment above), and the `{ null: null }` shape. Contrast the Go side's 544 specs.

hanahmily (commented):
Review of the parameter-binding implementation. Overall the injection-safety design is sound and self-contained. Findings below (1 High + 4 Medium); none need block this PR.

- **High** — the parsed AST is mutated in place during bind, so the parse-once/bind-many reuse this feature is meant to unlock is not yet possible (and a cached grammar could silently leak values across requests). Suggest documenting the one-shot invariant in this PR and implementing reusable-AST binding in a follow-up.
- **Medium** — grammar walked twice per query; literal count positions not range-checked (the `?` path is stricter); MCP `maxIntParamLength=20` admits int64 overflow; new MCP param validation ships without unit tests.

See inline comments for detail and suggested fixes.

pkg/bydbql/binder.go
mrproliu: Added a `paramBound` bool for this for check the parameter already bound or not.

pkg/bydbql/transformer.go
mrproliu: Added `countUnboundParams`.

pkg/bydbql/grammar.go
mrproliu: Add a checker after parameter bound.

mcp/src/query/validation.ts
mrproliu: Add a checker after parameter bound.

mcp/src/query/validation.ts
mrproliu: Done

hanahmily (approved):

PR #1213, opened 2026-07-09
Introduce reusable `Prepared` QL binding

mrproliu:
## Background

This is the follow-up requested in review on the parameter-binding PR: https://github.com/apache/skywalking-banyandb/pull/1209#discussion_r3541263665

The existing `BindParams` binds parameters by **mutating the parsed grammar in place**, so a parsed query can be bound exactly once — it cannot deliver the prepared-statement "parse once, bind many" payoff. This PR adds a prepared-query API built on that preferred **positional value overlay**: `Prepare` numbers the `?` placeholders on an immutable template, `Bind` resolves parameters into a per-request overlay, and `TransformBound` reads values from that overlay. The template is never written after `Prepare`, so one `*PreparedStatement` can be cached and bound concurrently.

### Prepare once, bind many

```go
// Parse and number placeholders once — cache this and reuse it.
ps, err := bydbql.Prepare(
    "SELECT * FROM STREAM sw IN default WHERE service_id = ? AND duration > ? LIMIT ?")
if err != nil {
    return err
}

// Bind fresh parameters per request; the template is left untouched.
bq, err := ps.Bind([]*modelv1.TagValue{
    strValue("web"),  // service_id = ?
    intValue(100),    // duration  > ?
    intValue(20),     // LIMIT ?
})
if err != nil {
    return err
}

// Transform reads the values from the per-request overlay.
result, err := transformer.TransformBound(ctx, bq)
```

The same `ps` can be `Bind`-ed again with different parameters, including concurrentl no re-parsing, no locking.

Value sets fold into a single placeholder: `WHERE id IN (?)` bound to a `str_array`/` so one template serves any number of values (no `IN (?, ?, …)` and no per-sizetemplate churn).

### One-shot path is unchanged

Callers that do not reuse a query keep the existing in-place path:

```go
g, err := bydbql.ParseQuery(query)
// ...
err = bydbql.BindParams(g, params) // mutates g in place; binds exactly once
result, err := transformer.Transform(ctx, g)
```

## Liaison query-path cache

The liaison now caches prepared statements on the gRPC BydbQL query path, so a repeatate is parsed only once instead of on every request (the participle lex/parse dominates per-request cost). `Query` looks the raw query text up in the cache: on a hit it reuses the cached `*PreparedStatement` and only binds the request's parameters; on a miss it prepares the statement,
stores it, and returns it.

Only parameterized queries (those containing `?`) are cached. A literal query bakes ievery distinct value would be a unique key that never hits and only evicts usefultemplates — literal queries therefore bypass the cache; they are counted separately as `bypass` and excluded from the hit ratio, which stays meaningful. The cache is a concurrency-safe LRU (built on
`hashicorp/golang-lru`) bounded by **both** an entry count and the estimated in-memornts (the query text plus a reflection-measured footprint of the parsed AST); whicheverbound is reached first evicts least-recently-used entries. It is enabled by default and never needs invalidation, because a parsed statement depends only on the query text, not on schema.

### Configuration

Flags on the `liaison` / `standalone` command:

- `--bydbql-prepared-cache-size int` — max number of cached prepared statements; `0` disables the cache (default `2000`).
- `--bydbql-prepared-cache-max-bytes int` — max total estimated size (bytes) of the cs the byte bound (default `10485760`, i.e. 10MiB).
- `--bydbql-slow-query-threshold duration` — end-to-end latency above which a query is counted as slow and tracked; `0` disables slow-query tracking (default `1s`).
- `--bydbql-topk-log-interval duration` — how often to log the top-10 cache-miss and he top-K log (default `5m`).

### Cache cost & observability

A cache hit is effectively free — a lookup costs ~25 ns and 0 allocations. A miss re-st scales with query complexity; for a wide OAP trace/log query (~13-column projection, long `WHERE`) a miss is about four orders of magnitude more expensive:

| lookup | ns/op | B/op | allocs/op |
|----|----|----|----|
| **hit**  | ~25 | 0 | 0 |
| **miss** (re-parse) | ~220,000 | ~147,000 | ~2,100 |

So the cache's value is entirely in keeping the hit ratio high: a low hit ratio pushePU and short-lived heap garbage (GC pressure) — back onto every request.

### Metrics

Metrics (under the liaison gRPC scope):

- `bydbql_prepared_cache_total{result="hit"|"miss"|"bypass"}` — lookups by outcome (`terized queries that are never cached; excluded from the hit ratio).
- `bydbql_prepared_cache_hit_ratio` — live cumulative hit ratio (over cacheable queries only).
- `bydbql_prepared_cache_count` / `bydbql_prepared_cache_bytes` — current entries and
- `bydbql_slow_query_total` — queries whose end-to-end latency exceeded `--bydbql-slow-query-threshold`.

### Finding un-cached & slow queries

The query text behind misses and slow responses is surfaced through logs, not Prometheus labels — query text is high-cardinality and can be several KB, so it must never become a label value.
Prometheus therefore carries only the counters above; the specific queries are logged

- **Access-log tags** — with `--enable-query-access-log`, each entry in the `bydbql-qe outcome via its service field: `bydbql-hit`, `bydbql-miss`, `bydbql-bypass`.Filtering for `bydbql-miss` / `bydbql-bypass` yields exactly the queries that did not hit the cache, with their full text.
- **Slow queries** — a query whose end-to-end latency exceeds `--bydbql-slow-query-thlow_query_total`.
- **Top-10 logs** — every `--bydbql-topk-log-interval` the liaison logs the top-10 cache-miss queries and the top-10 slow queries (ranked by occurrence count, each with its max latency), using a
bounded Space-Saving heavy-hitters tracker (O(1), fixed memory).

Together these let SkyWalking OAP spot un-parameterized queries (parameterize the varue sets with `IN (?)` bound to a `str_array`/`int_array`) and slow templates, withoutany high-cardinality metric.

- [x] Update the [`CHANGES` log](https://github.com/apache/skywalking-banyandb/blob/m


pkg/bydbql/binder.go
```diff
@@ -336,40 +318,115 @@ func (b *binder) expandLists() {
...
-	case *modelv1.TagValue_Null:
-		v.Null = true
+	case *modelv1.TagValue_StrArray:
+		if len(val.StrArray.GetValue()) == 0 {
+			return nil, fmt.Errorf("array parameter must not be empty")
+		}
+		expanded := make([]*GrammarValue, 0, len(val.StrArray.GetValue()))
+		for _, s := range val.StrArray.GetValue() {
+			expanded = append(expanded, &GrammarValue{String: &s})
+		}
```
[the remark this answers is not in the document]
mrproliu: These are false positives. The module targets go 1.25.8 (see go.mod), and since Go 1.22 the range loop variable is scoped per-iteration, so &s / &i take a distinct address each iteration. This is also covered by the equivalence tests — str_array param expands in IN and int_array param expands in HAVING bind ("a","b") / (200,500) and assert a proto-equal result against the literal query; an aliasing bug would collapse them to the last element and fail those tests.

pkg/bydbql/binder.go
```diff
@@ -336,40 +318,115 @@ func (b *binder) expandLists() {
...
+		return expanded, nil
+	case *modelv1.TagValue_IntArray:
+		if len(val.IntArray.GetValue()) == 0 {
+			return nil, fmt.Errorf("array parameter must not be empty")
+		}
+		expanded := make([]*GrammarValue, 0, len(val.IntArray.GetValue()))
+		for _, i := range val.IntArray.GetValue() {
+			expanded = append(expanded, &GrammarValue{Integer: &i})
+		}
```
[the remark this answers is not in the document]
mrproliu: These are false positives. The module targets go 1.25.8 (see go.mod), and since Go 1.22 the range loop variable is scoped per-iteration, so &s / &i take a distinct address each iteration. This is also covered by the equivalence tests — str_array param expands in IN and int_array param expands in HAVING bind ("a","b") / (200,500) and assert a proto-equal result against the literal query; an aliasing bug would collapse them to the last element and fail those tests.

banyand/liaison/grpc/bydbql.go
```diff
@@ -143,6 +166,79 @@ func (b *bydbQLService) Query(ctx context.Context, req *bydbqlv1.QueryRequest) (
...
+}
+
+func (d *topKDumper) close() {
+	if d != nil && d.cancel != nil {
+		d.cancel()
+	}
+}
+
+func (d *topKDumper) dump() {
+	d.logTopK(d.miss, "top bydbql cache-miss queries", func(s topKSlot) string {
```
hanahmily: **Requested change — the cumulative miss top-K conflates cold-start with thrashing.**

This list tracks misses cumulatively since process start. But every parameterized template necessarily misses exactly once on its first lookup (cold cache), so a perfectly healthy, always-hitting template still appears here at `count=1` and stays for the life of the process. That makes *presence* in the log non-actionable — only an elevated/growing miss count (a template repeatedly evicted and re-parsed → thrashing) is a real signal.

An operator reading a log titled "cache-miss queries" is likely to misread benign `count=1` cold-starts as misconfigured/uncacheable queries.

Suggest surfacing only *problematic* misses — see the tracker note on `topk.go`; per-interval windowing fixes this too.

The slow-query top-K and the `bydbql_prepared_cache_hit_ratio` gauge don't have this issue — this is specifically the cache-miss list's cumulative semantics.

banyand/liaison/grpc/topk.go
```diff
@@ -0,0 +1,94 @@
...
+		if dur > s.maxDur {
+			s.maxDur = dur
+		}
+		return
+	}
+	if len(t.slots) < t.k {
+		t.slots[key] = &topKSlot{key: key, count: 1, maxDur: dur}
+		return
+	}
+	// Full: evict the least-frequent entry and let the new key inherit its count.
```
hanahmily, line 74: **Requested change — the Space-Saving tracker is correct/thread-safe, but the reported numbers aren't actionable in the scenario this feature targets.** Ranked:

1. **(primary) Add windowing.** Both trackers are cumulative for the whole process lifetime. Reset — or exponentially decay — the maps at each dump/`snapshot`. This one change fixes items 2 and 3 below and largely neutralizes 4, and makes the numbers track the *current* workload instead of ancient history.

2. **`count=` is a Space-Saving over-estimate, not a true count.** The newcomer inherits `minCount+1`, so its printed count includes frequency donated by other evicted keys, and no per-entry error term (ε) is tracked to bound it. `count=812` does not mean the query missed 812 times. If you keep it cumulative, track ε so `count` can be reported as a bounded range; windowing makes this mostly moot.

3. **Inflation is worst exactly during thrashing** — the case the miss log exists to diagnose. In a high-cardinality stream (mostly-unique templates) nearly every `observe` hits this eviction branch; the min-count floor rises ~1 every k inserts, so after N observations all k slots converge to ≈ N/k with near-identical counts. Result: the top-10 becomes 10 arbitrary queries with indistinguishable counts, and the real repeat-offenders don't stand out.

4. **`k = 10` is a weak guarantee** — Space-Saving with k=10 only reliably surfaces items exceeding ~10% of the stream, which a many-template workload rarely hits. Bump to ~100–200; the O(k) min-scan and map stay trivially cheap.

5. **`maxDur` is silently lost on eviction.** A slow query that is evicted and later re-observed re-enters as a fresh slot with `maxDur = dur`, discarding its earlier peak → `max_latency` under-reports for churning keys. Windowing moots this within a window.

6. **Slow queries are ranked by frequency only.** `snapshot` sorts by `count` desc, so a query that was catastrophically slow once (huge `max_latency`, `count=1`) gets buried/evicted under frequently-mildly-slow ones. Rank or additionally emit the slow tracker by `max_latency`.

7. **(polish) Non-deterministic tie-break.** Equal-count entries order by map iteration + unstable sort, so output order flaps between dumps. Sort by `(count desc, maxDur desc, key)`.

Recommendation: do 1 + 4 (and 6 for the slow tracker) — small, high-leverage, and they resolve the rest.

hanahmily (changes requested):
Two requested changes on the query top-K logging — inline notes on (1) the cache-miss dump and (2) the Space-Saving tracker algorithm. Both are about making the logs actionable; no correctness/race bug in the tracker itself.

banyand/liaison/grpc/topk.go
mrproliu: fixed 1+4+6.

banyand/liaison/grpc/bydbql.go
mrproliu: I have changed the top-K to `128` and the logger filters only with `count >= 2`.

hanahmily (approved):

PR #1244, opened 2026-08-04
add cluster schema consistency check in FODC lifecycle endpoint

mrproliu:
## What

Adds an **FODC schema consistency check**: a cluster-wide, read-only health signal that detects when a node's schema cache or runtime view has drifted from the metadata registry.

For every group, the checker compares three fingerprints per object:
- **registry** — the source of truth, read once from a schema-serving node
- **cache** — what each node has cached
- **runtime** — what each node is actually serving (materialized schema + index rules)

The verdict is exposed object-centrically on `GroupLifecycleService.InspectAll` and aggregated by fodc-proxy at `GET /cluster/lifecycle`.

## Why

Schema propagation is asynchronous; a stale cache or an un-applied index rule on a single node can silently corrupt reads. Until now there was no way to observe this from outside. This gives operators a steady-state "is every node serving the schema the registry thinks it should" signal.

## How it works

- **Fingerprinting** (`pkg/schema/consistency`): deterministic proto marshal + xxhash over each object plus its derived index-rule set.
- **Verdict** (`Checker`): registry vs. per-node cache/runtime, rolled up to `CONSISTENT` / `INCONSISTENT` / `UNKNOWN`. A divergence must survive **two consecutive rounds** before it is reported, filtering in-flight propagation and torn reads. An incomplete roster downgrades to `UNKNOWN` rather than passing a silent node as consistent.
- **Collection**: node evidence rides on the existing `CollectDataInfo`/`CollectLiaisonInfo` inspect path, folded in behind a new `include_schema_state` request flag. Only the FODC path sets it; the public `GroupRegistryService.Inspect` RPC honors the caller's flag (default `false`) and group-delete never sets it, so neither pays for the per-object fingerprinting or the extra payload.
- **Merge** (fodc-proxy): each agent reports its own verdict; the proxy keeps the **most severe** status and the **deduped union** of issues and per-object node fingerprints, so one healthy node can never mask another node's finding.
- **Error surfacing**: registry-build, node-roster, and liaison-collection failures are reported on `GroupLifecycleInfo.errors`, not just logged.

## Changes

| Area | Change |
| --- | --- |
| `api/proto/.../database/v1` | `ObjectSchemaState`, `include_schema_state` on inspect request |
| `api/proto/.../fodc/v1` | `SchemaConsistency` object-centric verdict on `GroupLifecycleInfo` |
| `pkg/schema/consistency` | fingerprint, derive, object sort, suppression-aware `Checker` |
| `banyand/queue/sub` | registry fingerprint build + per-group verdict wiring |
| `banyand/{measure,stream,trace}` | schema-state collection folded into the collectors |
| `fodc/proxy` | verdict merge + `/cluster/lifecycle` exposure |
| `test/e2e-v2/cases/fodc` | steady-state consistency guard |


hanahmily:
## 🚫 Requested changes

### Move `xxhash` fingerprint compute out of `banyandb` core into a FODC sidecar agent

The current PR embeds fingerprint generation, suppression state, and registry fingerprint building into the data path (`banyand/queue/sub/schema_consistency.go`, `pkg/schema/collect_state.go`, the per-shard `metadata.go` files, the new `pkg/schema/consistency/` package, plus the `include_schema_state` flag on `Inspect`). This is observability logic in the data path. Before merge, please invert the design so the data path returns its existing protos and the FODC agent fingerprints on receive.

#### Why this blocks

**Decoupling of observability from the data path.** FODC's hash algorithm, suppression policy, and classification rules should be able to ship independently of `banyandb`. Today they're pinned.

**Footprint in `banyandb` is unnecessary.** ~150 LOC of fingerprint code in the data path plus 4 new proto fields on an existing RPC. None of it is needed if the agent owns the compute.

**Same wire integrity guarantees.** The trust boundary moves one hop; its posture is unchanged. There's no defensible reason to keep the compute inside `banyandb`.

#### Proposed architecture

```mermaid
sequenceDiagram
    autonumber
    actor Op as Operator / dashboard
    participant FP as fodc-proxy
    participant A as fodc-agent (sidecar)
    participant L as Liaison (schema-serving)
    participant DN as Data nodes

    Op->>+FP: GET /cluster/lifecycle
    FP->>+A: collect(group)
    A->>+L: fetch registry view (existing RPC)
    L-->>-A: registry proto
    A->>A: registry_fp = xxhash(deterministic-marshal + derived rules)
    par one server-stream per data node
        A->>+DN: StreamGroupSchemaState(group)
        DN-->>A: SchemaSnapshotEvent.object (N times)
        DN-->>-A: SchemaSnapshotEvent.trailer{node, version, count, errors}
    and
        A->>+DNo: StreamGroupSchemaState(group)
        DNo-->>-A: ...stream...
        DNo-->>-A: trailer
    end
    A->>A: Checker (two-strike, streaks in agent)
    A-->>-FP: SchemaConsistency{verdict, objects, issues}
    Note over FP: merge · most severe + deduped union
    FP-->>-Op: 200 OK · aggregated GroupLifecycleInfo
```

#### Required

The `pkg/schema/consistency/` package moves to the FODC agent binary (e.g. `fodc/agent/consistency/`), replacing the new package in `banyandb`.

`banyand/queue/sub/schema_consistency.go` is deleted. The registry fingerprint is built by the agent from the registry RPC response rather than inside queue/sub.

The per-shard `metadata.go` files for measure, stream, and trace revert to no fingerprint code.

`pkg/schema/collect_state.go` is deleted; collection lives in the agent now.

The `include_schema_state` flag on `GroupRegistryServiceInspectRequest` is removed.

The `ObjectSchemaState` field on `DataInfo` and `LiaisonInfo` is removed.

The per-shard `metadata_consistency_test.go` files (×3) are deleted; equivalent coverage moves into `fodc/agent/consistency_test.go`.

#### Preserved (no change)

- `Checker` algorithm (two-strike suppression, classify).
- `xxhash` + `clearVolatile` fingerprint format.
- `ObjectConsistency` / `SchemaIssue` / `SchemaConsistency` wire shape exposed by the proxy.
- Proxy merge logic (most-severe + deduped union), including the `len(nodes) < expectedNodes → UNKNOWN` rule from `pkg/schema/consistency/checker.go:144`.
- Existing per-catalog fingerprint identity semantics in `pkg/schema/consistency/fingerprint.go:92-107`.

Companion comment below has the streaming proto API the agent talks to — they're a pair.

hanahmily:
## 🚫 Requested changes

### Replace `ObjectSchemaState` (uint64 fingerprints) with a streaming `NodeSchemaStateService` that returns proto bodies

The companion comment above moves fingerprint compute to the agent. To make that work, the data path must return each node's schema **proto body**, not uint64 fingerprints. This PR's `ObjectSchemaState` (uint64s) and `include_schema_state` flag must be removed; in their place, add `NodeSchemaStateService.StreamGroupSchemaState`, which streams the actual snapshot objects.

Before merge, please make this change.

#### Why this blocks

The current `ObjectSchemaState` is a dead end for the agent-side design. Once fingerprint code lives in the agent, shipping pre-hashed uint64s from `banyandb` is wasted work on the hot path.

`Inspect` cannot stream; this RPC must. At ~3 KB per schema object (proto body + bound index rules), a unary RPC hits gRPC's default 4 MB ceiling at ~1k objects and breaks past ~5k. Streaming with bounded per-message size sidesteps the limit and lets the agent fan out to N data nodes in parallel — total wall-clock is the slowest single node, not the sum.

`SchemaInfo` returns names, not protos. This is the only clean way to get proto bodies without N+1 `GetSchema` calls.

#### Required: new RPC

```proto
service NodeSchemaStateService {
  // StreamGroupSchemaState returns this node's full snapshot for one
  // group. The server holds a snapshot lock for the duration of the
  // stream so every yielded object comes from the same view; mid-stream
  // watch events queue behind the stream's release.
  //
  // Empty stream (zero object events) is a normal termination — the
  // agent treats the absence of objects as "this node doesn't cache
  // this group", not as an error.
  rpc StreamGroupSchemaState(NodeSchemaStateRequest)
      returns (stream SchemaSnapshotEvent);
}

message NodeSchemaStateRequest {
  string group = 1;

  // internal_version correlates the snapshot with what the agent
  // asked for. Zero means "no version pinning, best-effort" — the
  // agent then treats the stream as torn-read suspect and lets the
  // two-round suppression absorb it.
  uint64 internal_version = 2;
}

// SchemaSnapshotEvent is the streaming element. Wire-extensible — new
// event kinds add oneof variants without breaking older agents.
// Stream terminates either with a SchemaSnapshotTrailer (clean) or a
// non-EOF status (transport error).
message SchemaSnapshotEvent {
  oneof event {
    SchemaObjectSnapshot  object  = 1;
    SchemaSnapshotTrailer trailer = 2;
  }
}

// One schema object's body, generated from the server's snapshot.
// The server's snapshot_version is identical for every object event
// in a single stream; the agent rejects streams with version drift.
message SchemaObjectSnapshot {
  uint64 snapshot_version = 1;
  ObjectSnapshot object = 2;
}

// Final message of a successful stream. Always sent before EOF.
message SchemaSnapshotTrailer {
  // node echoes the responding node so the agent can detect misrouted
  // responses in mixed-version clusters (older nodes may field
  // requests on behalf of newer ones).
  string node = 1;

  // snapshot_version mirrors the value tagged on every object event
  // in this stream — agent verifies first-event and trailer agree.
  uint64 snapshot_version = 2;

  // object_count is a sanity check the agent can verify against the
  // number of objects it received. Mismatch ⇒ torn stream, retry.
  uint32 object_count = 3;

  // errors are failures observed during snapshot collection. They
  // flow up to GroupLifecycleInfo.errors but do NOT cause the stream
  // to fail — partial successes are still useful.
  repeated string errors = 4;
}

// ObjectSnapshot is one schema object's body plus the index rules
// bound to it. The oneof keeps catalog-specific types well-typed
// without forcing the agent to switch on `kind` strings.
message ObjectSnapshot {
  string kind = 1;   // schema.Kind.String()
  string name = 2;
  oneof payload {
    banyandb.database.v1.Group            group_payload              = 10;
    banyandb.database.v1.Stream           stream_payload             = 11;
    banyandb.database.v1.Measure          measure_payload            = 12;
    banyandb.database.v1.Trace            trace_payload              = 13;
    banyandb.database.v1.IndexRule        index_rule_payload         = 14;
    banyandb.database.v1.IndexRuleBinding index_rule_binding_payload = 15;
    banyandb.database.v1.TopNAggregation  topn_payload               = 16;
  }
  // bound_index_rules is the set the proto is hashed against. The
  // agent uses these directly — no re-derivation, no extra RPC.
  repeated banyandb.database.v1.IndexRule bound_index_rules = 20;
}
```

#### Required: removed proto surface

The `bool include_schema_state = 2` field on `GroupRegistryServiceInspectRequest` (`api/proto/banyandb/database/v1/rpc.proto:482-493`) is removed.

The `repeated ObjectSchemaState schema_objects` field on `DataInfo` (`api/proto/banyandb/database/v1/rpc.proto:533-534`) is removed.

The `repeated ObjectSchemaState schema_objects` field on `LiaisonInfo` (`api/proto/banyandb/database/v1/rpc.proto:626` area) is removed.

The `message ObjectSchemaState` definition itself (`api/proto/banyandb/database/v1/rpc.proto:537-551`) is removed.

#### Acceptance criteria

`StreamGroupSchemaState` is implemented on every data node and liaison that owns a group cache. The server holds a snapshot lock for the stream's duration; every yielded event shares one `snapshot_version`.

Default gRPC message-size limits are not raised. Streaming with bounded per-message `ObjectSnapshot` is the path; unary fallback is not in scope.

The agent opens one stream per data node in parallel; total wall-clock is bounded by the slowest single node, not the sum.

The agent verifies `object_count` on `SchemaSnapshotTrailer` against the received-object count. A mismatch triggers a single retry, then degrades to UNKNOWN for that node.

Wire-compat is preserved end-to-end: new RPC + new messages + new fields are non-breaking additions, and the removed fields are documented in `CHANGES.md` under breaking changes.

Tests cover: empty group, partial failure, large group (≥ 5,000 objects verifying default message-size limit is not exceeded), mid-stream watch event (verifies version-pinning surfaces as a torn-stream that the agent rejects), and misrouted response (verifies the `node` echo on the trailer matches the requested endpoint identity).

This is a cohort ship with the first comment above — the architectural shift only works if this RPC exists.

hanahmily (changes requested):
Requesting changes on PR #1244 before this can land.

The current shape puts fingerprint compute and the cache-fingerprint proto fields inside `banyandb` core. Two changes need to land together for this feature to work cleanly:

1. **Move fingerprint compute to the FODC sidecar agent** — `xxhash`, suppression state, classify, merge — out of `banyandb` core. The data path returns what it always returns; the agent fingerprints on receive. Detailed rationale and proposed topology in the first comment.

2. **Replace `ObjectSchemaState` and `include_schema_state` with a streaming `NodeSchemaStateService.StreamGroupSchemaState` RPC** that returns each node's cache and materialized runtime as proto bodies. Detailed proto in the second comment.

These are a pair — the architectural shift only works if the streaming RPC exists, and vice versa. With both in place the `banyandb` diff for this feature should be net negative LOC.

mrproliu:
Add `NodeSchemaStateService` in each liaison/data node to collect the cached schema. Then all the fingerprints are generated on the FODC `proxy/agent` side.

wu-sheng:
Remember to update grafana setup and later oap horizon setup update.

hanahmily (approved):

PR #1247, opened 2026-08-04
Fix flaky TestTraceWriteRoundTrip by stopping modules before reading trace parts

hanahmily:
## Summary

`TestTraceWriteRoundTrip` intermittently opened a trace part as soon as `metadata.json` appeared, while the trace service could still flush or merge the part. That allowed the dump reader to combine metadata and data from different file generations and panic on an invalid byte-block type.

The test now stops the modules after publishing and before discovering or reading part directories. Cleanup is guarded with `sync.OnceFunc`, preserving deferred cleanup on early failures while making shutdown the explicit flush/quiescence barrier.

No timeout, polling window, sleep, retry, skip, or assertion was changed. The fixed test passed 50 race-enabled executions in five batches with `GOMAXPROCS=4`, plus a final race-enabled smoke check after restoring the fix.

Measure and stream round-trip tests were left unchanged because the reported failure and exact local reproduction were trace-specific.

Flaky-Test: github.com/apache/skywalking-banyandb/banyand/internal/dump/trace.TestTraceWriteRoundTrip

### DIFF
--- a/banyand/internal/dump/trace/suite_test.go
+++ b/banyand/internal/dump/trace/suite_test.go
@@ -23,6 +23,7 @@ import (
 	"io/fs"
 	"path/filepath"
 	"strconv"
+	"sync"
 	"testing"
 	"time"
 
@@ -84,8 +85,8 @@ func TestTraceWriteRoundTrip(t *testing.T) {
 		"--trace-root-path=" + rootPath,
 		"--trace-flush-timeout=200ms",
 	}
-	moduleDefer := test.SetupModules(flags, pipeline, metaSvc, traceSvc)
-	defer moduleDefer()
+	stopModules := sync.OnceFunc(test.SetupModules(flags, pipeline, metaSvc, traceSvc))
+	defer stopModules()
 
 	registerRoundTripTrace(t, metaSvc)
 	require.Eventually(t, func() bool {
@@ -139,6 +140,7 @@ func TestTraceWriteRoundTrip(t *testing.T) {
 		req.NoError(errPub)
 	}
 	req.Empty(bp.Close())
+	stopModules()
 
 	var partDirs []string
 	require.Eventually(t, func() bool {


hanahmily:
# Review: 20260804-130827-fixthenextfl

## Verdict

All four judgments pass. The fix is a clean root-cause fix with no masking, a credible reproduction, and a properly recorded `Flaky-Test:` line.

## 1. Root cause addressed

The CI failure (`artifacts/reports/ci-failure.md` lines 19–21, 133–142) is a panic in `decodeDefaultTagValues` caused by `BytesBlockDecoder` receiving a block with byte `0x80` instead of the supported `0`/`1`. The panic stack trace shows `PartReader.decodeBlock` at `iterator.go:77` reading a trace part whose metadata and data blocks no longer match — i.e. the dump reader combined metadata from one file generation with data from another.

The CI log (lines 1369–1370) shows the trace module was still running when the panic happened (`stopping module=STANDALONE-TEST name=trace ... elapsed=380.203073`). The trace module had not quiesced before the dump reader opened the part, so the race was real.

The diff (`banyand/internal/dump/trace/suite_test.go:88,143`) inserts an explicit `stopModules()` between `bp.Close()` and `partDirs` discovery, turning shutdown into a hard flush/quiescence barrier. `sync.OnceFunc` (Go 1.21+, available in this repo's `go1.25.12` toolchain) keeps the `defer stopModules()` at line 88 idempotent so cleanup still runs even if the explicit call never reached. This is exactly the fix the triage handoff (`root_cause: "the test treated metadata.json visibility as a flush-completion barrier and read a multi-file trace part while the active service could still mutate it"`) calls for.

## 2. Masking absent

The diff touches only:
- one new import (`sync`, line 26)
- wrapping `SetupModules` return in `sync.OnceFunc` (line 88)
- explicit `stopModules()` call before part discovery (line 143)

No timeout was raised. No `Eventually`/`Consistently` window widened. No `time.Sleep` added. No retry or `--repeat` bumped inside the test. No `t.Skip`. No assertion deleted or loosened. The fix adds a synchronization barrier instead of a wait-and-pray delay — the opposite of masking.

## 3. Reproduction real

Verification command from the handoff:
```
GOMAXPROCS=4 bin/ginkgo --race -ldflags "..." --focus TestTraceWriteRoundTrip --repeat 9 ./banyand/internal/dump/trace/...
```

- `--race` is enabled — a non-`-race` run could not have demonstrated a race was fixed.
- `GOMAXPROCS=4` matches the CI runner that produced the original failure.
- `--repeat 9` per execution × 5 batches (`pr-body.md` line 7) = 45+ runs, and `verification_runs: 50` is reported.
- The exact panic signature from the CI failure (`cannot decode string lengths: cannot decode bytes block: unexpected block type: 128`) would surface under `--race`/repeated runs, and 50 race-enabled executions is meaningful evidence that the window is closed.

## 4. No regression / scope creep

The diff is confined to `banyand/internal/dump/trace/suite_test.go`. Measure and stream round-trip tests are intentionally left alone (`pr-body.md` line 9: "Measure and stream round-trip tests were left unchanged because the reported failure and exact local reproduction were trace-specific.") — consistent with the CI log showing `TestMeasureWriteRoundTrip` and `TestStreamWriteRoundTrip` passed in the same ginkgo process.

## 5. No recurrence in packet

`recurrence` is not present / not > 0, so judgment 5 does not apply.

## pr-body.md check

Line 11 of `artifacts/contracts/pr-body.md`:
```
Flaky-Test: github.com/apache/skywalking-banyandb/banyand/internal/dump/trace.TestTraceWriteRoundTrip
```
Matches the triage `flaky_test_key` literally. Next run can discover this test's history.

## Issues

None.

hanahmily:
Closing this in favour of a cleaner submission of the same fix.

PR #1248, opened 2026-08-04
Add latest parameter in slow QL log

mrproliu:
## What

The BydbQL slow-query top-K log reports the query template but drops the bound parameters, so an operator can see *which query* is slow but never *which values made it slow* — the time window, the `LIMIT`, or the one service that is always the slow one. This PR adds them back as a `last_params` field, behind a redaction mode, because parameters are client-supplied and can carry user data.

## Why a mode instead of just logging them

Parameters bound to `?` placeholders are exactly the values a user does not want in an operational log: service names, instance ids, trace ids. Logging them unconditionally is not acceptable; dropping them entirely is what we have today and it costs real diagnostic power. So the amount of parameter detail that reaches the log is a deployment choice:

| `--bydbql-topk-param-mode` | `last_params` | Reveals |
|---|---|---|
| `none` | field is absent | nothing |
| `fingerprint` (default) | `str(len=12):fp=1a2b3c4d` | the length, and *whether two slow queries used the same value* — never the value |
| `raw` | `"checkout-svc"` | the value itself |

Invalid values fail at startup rather than silently degrading the log. An unset value (a `server` built without a parsed `FlagSet`) resolves to the default, because "unset" is not the same thing as "invalid".

## Worked examples

Every cell below is real output, generated by calling the redaction path directly rather than written by hand. Digests are stable across runs, restarts and nodes, so these exact values are reproducible.

| Parameter type | Bound value | `none` | `fingerprint` (default) | `raw` |
|---|---|---|---|---|
| `null` | `NULL` | *(field omitted)* | `null` | `null` |
| `int` | `100000` | *(field omitted)* | `100000` | `100000` |
| `timestamp` | `2026-08-04T10:00:00Z` | *(field omitted)* | `2026-08-04T10:00:00Z` | `2026-08-04T10:00:00Z` |
| `int_array` | `[1 2 3]` | *(field omitted)* | `int[n=3]:[1 2 3]` | `int[n=3]:[1 2 3]` |
| `str` | `checkout-svc` | *(field omitted)* | `str(len=12):fp=b700d1a3897dbfc7` | `"checkout-svc"` |
| `str_array` | `[checkout-svc payment-svc]` | *(field omitted)* | `str[n=2]:fp=[b700d1a3897dbfc7 f81a40ce4e7ea65e]` | `str[n=2]:["checkout-svc" "payment-svc"]` |
| `binary_data` | 3 bytes | *(field omitted)* | `bytes(len=3):fp=743e13ee0c4ee5a5` | `bytes(len=3):fp=743e13ee0c4ee5a5` |

Reading the table: the four non-string rows are identical across `fingerprint` and `raw` — those types are never redacted, because they are what explains the latency and carry nothing user-identifying. `binary_data` is identical too, but for the opposite reason: it is never revealed, not even under `raw`. Only `str` and `str_array` actually differ between the two modes, and those are exactly the types that can carry user data. Note also that `str_array` digests each element with the same function the scalar path uses, so one value stays correlatable whether it arrives alone or inside an `IN (?)` set.

The same query, as the log line an operator actually sees, for `SELECT * FROM STREAM sw IN default WHERE service_id = ? AND TIME > ? LIMIT ?` bound to `checkout-svc` / `2026-08-04T10:00:00Z` / `100000`:

```
# --bydbql-topk-param-mode=none
"SELECT * FROM STREAM sw IN default WHERE service_id = ? AND TIME > ? LIMIT ?" count=3 max_latency=8.3s max_latency_at=2026-08-04T10:05:00Z last_seen=2026-08-04T10:09:00Z

# --bydbql-topk-param-mode=fingerprint   (default)
"SELECT * FROM STREAM sw IN default WHERE service_id = ? AND TIME > ? LIMIT ?" count=3 max_latency=8.3s max_latency_at=2026-08-04T10:05:00Z last_seen=2026-08-04T10:09:00Z last_params=[str(len=12):fp=b700d1a3897dbfc7, 2026-08-04T10:00:00Z, 100000]

# --bydbql-topk-param-mode=raw
"SELECT * FROM STREAM sw IN default WHERE service_id = ? AND TIME > ? LIMIT ?" count=3 max_latency=8.3s max_latency_at=2026-08-04T10:05:00Z last_seen=2026-08-04T10:09:00Z last_params=["checkout-svc", 2026-08-04T10:00:00Z, 100000]
```

Under `none` the field is absent entirely rather than present-and-empty, so a `none` deployment shows no vestigial `last_params=[]`. Note what the default already buys you without revealing anything: the `LIMIT` is `100000` and the time bound is a fixed instant — that is usually enough to explain the 8.3s on its own, and the service name is only needed to answer "is it always the same one", which the digest answers too.

Array capping, which only appears past the 8-element limit — a 12-element `IN (?)`:

```
# fingerprint
str[n=12]:fp=[469389b2fc486795 d73062237a2ddf2e 2f7bf6ce89498d0e 2783c5209bba2ee 84e187903785f429 ff041ccc16c439b6 949e1127929df714 cdd3ba8f8642d95c +4 more]

# raw
str[n=12]:["svc-00" "svc-01" "svc-02" "svc-03" "svc-04" "svc-05" "svc-06" "svc-07" +4 more]
```

`n=12` is the true element count and `+4 more` states what was dropped, so a capped list is never mistaken for a complete one.


banyand/liaison/grpc/paramredact.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// redactParam renders a single parameter. Numeric, timestamp and null values pass through
+// verbatim in every mode above none: they are what explains why a query is slow (window
+// width, LIMIT, thresholds) and they carry no user-identifying content. Only str and
+// str_array are subject to the mode; binary_data is always digested, never rendered.
+func redactParam(mode paramMode, param *modelv1.TagValue) string {
+	switch value := param.GetValue().(type) {
+	case *modelv1.TagValue_Null:
+		return "null"
```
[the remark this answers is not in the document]
mrproliu: The protobuf will take care of the `nil` value, so no need to worry.

banyand/liaison/grpc/paramredact.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// redactParams renders one query's bound parameters for the top-K log under mode, or ""
+// when there is nothing to render, which keeps the field out of the log line entirely.
+//
+// It runs at observe time rather than at dump time, so the tracker never holds a raw
+// parameter value in memory: whatever mode forbids does not merely go unprinted, it is
+// never retained. The cost is irrelevant because only queries already over the slow
+// threshold reach this path.
+func redactParams(mode paramMode, params []*modelv1.TagValue) string {
```
hanahmily, line 70: Consider adding a `BenchmarkRedactParams` (one per mode × one per type) here. This runs in the slow-query hot path — `fingerprint` calls `xxhash.Sum64String` for every string parameter, and `joinCapped` allocates per element past the 8-element cap. A one-time benchmark gives future contributors a regression baseline when they extend the type switch (e.g. add a new `modelv1.TagValue` variant) or switch digests, instead of having to construct one ad hoc.

banyand/liaison/grpc/paramredact.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// fingerprint is a stable, UNSALTED 64-bit digest of a parameter value. Unsalted is a
+// deliberate tradeoff: it keeps a value's digest identical across restarts and across
+// liaison nodes, which is what lets an operator tell "one particular service is always
+// slow" from "every service is slow". The cost is that a low-cardinality value — a
+// service name, an instance id — can be recovered by digesting a candidate dictionary.
+// This hides values from a casual reader of the log; it is not a cryptographic guarantee,
+// and it must not be described as one.
+func fingerprint(value string) string {
```
hanahmily, line 138: Same benchmark note — `fingerprint` is the inner loop. Capturing the per-call ns/op for a representative service-name string (~12 bytes) and a long value (1 KiB) would pin the cost of any future swap to a salted HMAC or a SHA-256.

banyand/liaison/grpc/paramredact_test.go
```diff
@@ -0,0 +1,359 @@
...
+func TestRedactParamsCapsLongArraysAndReportsTheRemainder(t *testing.T) {
+	values := make([]string, 0, 20)
+	for idx := 0; idx < 20; idx++ {
+		values = append(values, "svc-"+string(rune('a'+idx)))
+	}
+	got := redactParams(paramModeFingerprint, []*modelv1.TagValue{strArrayParam(values...)})
+
+	assert.Contains(t, got, "str[n=20]", "the full length must survive the cap")
+	assert.Contains(t, got, "+12 more", "the dropped remainder must be stated, not silently truncated")
+	assert.Equal(t, maxRenderedArrayElems, strings.Count(got, " ")-1,
```
hanahmily: The cap-counting assertion `assert.Equal(t, maxRenderedArrayElems, strings.Count(got, " ")-1, ...)` works but is indirect — a reader has to mentally reconstruct the `str[n=20]:fp=[a b c d e f g h +12 more]` layout to verify the math. A direct check (`strings.Count(got, ":fp=")` and `assert.Contains(got, "+12 more")`) would be clearer and would catch the same regressions.

hanahmily (changes requested):
Nice PR overall — fail-closed redaction, observe-time stamping so the tracker never holds raw values, and `strconv.Quote` against log injection. One non-blocking ask: a small benchmark would let future contributors reason about the cost when extending the type switch. Detailed inline comments below.

mrproliu:
# `redactParams` benchmark baseline

## Environment

| | |
|---|---|
| OS / arch | `darwin/arm64` |
| CPU | Apple M3 Max (16 logical) |
| Package | `github.com/apache/skywalking-banyandb/banyand/liaison/grpc` |
| Method | `-count=3`, median reported; run-to-run spread ≤ 3% except the `none` row (~1.4 ns, noise-dominated at that scale) |

```sh
GOTOOLCHAIN=auto go test ./banyand/liaison/grpc/ -run '^$' -bench 'BenchmarkRedactParams|BenchmarkFingerprint' -benchmem -count=3
```

**Read these as a per-slow-query cost, not a per-request one.** Rendering is reached only by queries that already exceeded `--bydbql-slow-query-threshold` (default `1s`), so the most expensive cell below is ~0.00005% of the query it annotates.

## Latency — ns/op

| Parameter type | `none` | `fingerprint` (default) | `raw` |
|---|---|---|---|
| `null` | 1.35 | 5.17 | 5.16 |
| `int` | 1.37 | 19.98 | 19.76 |
| `timestamp` | 1.39 | 36.47 | 36.46 |
| `int_array` (3) | 1.36 | 149.30 | 149.00 |
| `binary` (3 B) | 1.39 | 111.00 | 108.80 |
| **`str`** (12 B) | 1.35 | **106.20** | **68.60** |
| **`str_array`** (2) | 1.35 | **196.20** | **276.10** |
| **`str_array`** (64, over cap) | 1.35 | **527.30** | **568.50** |

## Allocations — B/op · allocs/op

| Parameter type | `none` | `fingerprint` (default) | `raw` |
|---|---|---|---|
| `null` | 0 · 0 | 0 · 0 | 0 · 0 |
| `int` | 0 · 0 | 8 · 1 | 8 · 1 |
| `timestamp` | 0 · 0 | 24 · 1 | 24 · 1 |
| `int_array` (3) | 0 · 0 | 104 · 4 | 104 · 4 |
| `binary` (3 B) | 0 · 0 | 64 · 3 | 64 · 3 |
| **`str`** (12 B) | 0 · 0 | **64 · 3** | **16 · 1** |
| **`str_array`** (2) | 0 · 0 | **192 · 6** | **176 · 6** |
| **`str_array`** (64, over cap) | 0 · 0 | **600 · 13** | **408 · 13** |

## `fingerprint` in isolation

The inner loop of the default mode: one call per string parameter, and one per array element up to the cap.

| Value | ns/op | Throughput | B/op | allocs/op |
|---|---|---|---|---|
| service name, 12 B | 25.13 | 478 MB/s | 16 | 1 |
| long value, 1 KiB | 79.96 | 12.8 GB/s | 16 | 1 |

An 85× larger input costs 3.2× more — the digest is dominated by fixed overhead, not by hashing. That is the baseline to measure any future swap to a salted HMAC or SHA-256 against.

## Array capping — cost is bounded by the cap, not by the array

Measured separately with `fingerprint` and a `str_array` of increasing length:

| Elements | ns/op | B/op | allocs/op |
|---|---|---|---|
| 8 (exactly at cap) | 448.1 | 592 | 12 |
| 64 | 511.0 | 600 | 13 |
| 1,024 | 538.5 | 656 | 15 |
| 10,000 | 542.2 | 656 | 15 |

**1,024 and 10,000 elements cost the same** — 540 ns, 656 B, 15 allocs. `joinCapped` iterates to `min(n, maxRenderedArrayElems)`, so work is O(cap), not O(n); the only thing that grows past the cap is the decimal width of the `+N more` count. A pathological `IN (?)` bound to a 10,000-entry `str_array` costs the same as a 64-entry one.

## What the numbers say

**`none` is free.** ~1.35 ns and zero allocations across every type, because `redactParams` returns before touching the parameter slice at all. Turning the feature off costs nothing measurable, so the flag is a real off switch rather than a cheaper rendering path.

**The mode only gates strings — the benchmark proves it.** `null`, `int`, `timestamp`, `int_array` and `binary` are within run-to-run noise of each other across `fingerprint` and `raw` (e.g. `int_array` 149.30 vs 149.00). Only `str` and `str_array` diverge. That is the same claim the docs make, verified by measurement rather than by reading the type switch.

**The safe default is not the cheap one, and that is fine.** `fingerprint`/`str` (106 ns, 3 allocs) is ~55% slower than `raw`/`str` (69 ns, 1 alloc), because it hashes and formats a length prefix where `raw` does a single `strconv.Quote`. It is still ~100 ns on a query that took over a second.

**Arrays invert that.** `fingerprint`/`str_array` (196 ns) is *cheaper* than `raw` (276 ns): a 16-hex-digit digest is shorter to build and join than a quoted string, so the digest path moves fewer bytes.

**Worst realistic case is ~570 ns / 13 allocs** — an over-cap `str_array` under `raw`. Against a >1 s query that is noise, and it does not grow with input size.

hanahmily (approved):

PR #1267, opened 2026-08-09
ci: skip tests on main pushes

owenwillison-vajra:
## What changed

- add one pull-request-only test gate
- make unit, integration, FODC, and end-to-end jobs depend on that gate, so they skip transitively on `push` events to `main`
- keep pull request test behavior unchanged
- keep preparation, consistency checks, vulnerability scanning, FODC change detection, and the final CI result on `main`
- allow the final result job to accept the skipped gate on `main` while still requiring every test to pass on pull requests

## Why

Pull request checks are mandatory and direct pushes to `main` are prohibited, so running the complete test matrix again after every merge duplicates the already-required validation. The previous main CI run consumed approximately 277 runner-minutes.

The gate keeps the workflow maintainable: the event condition is declared once rather than repeated on every test job.

## Validation

- `actionlint .github/workflows/ci.yml`
- YAML parsing
- `git diff --check`
- real push-event validation of the gate on a temporary fork branch: [run 31287490378](https://github.com/owenwillison-vajra/skywalking-banyandb/actions/runs/31287490378)
  - Prepare, Check, Vulnerability Scan, FODC detection, and Continuous Integration passed
  - Enable Tests and all eight test/E2E jobs were skipped

The temporary test branch was deleted after validation. Only `.github/workflows/ci.yml` is changed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,48 +150,61 @@ jobs:
           fi
 
           echo "should_run=${should_run}" >> "${GITHUB_OUTPUT}"
+  test-gate:
+    name: Enable Tests
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo 'Tests enabled'
   test-banyand:
     name: Test Banyand
-    needs: [prepare]
+    needs: [prepare, test-gate]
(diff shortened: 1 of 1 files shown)


wu-sheng:
I don't think this is a right call. Main branch test is always needed.

hanahmily (approved):

PR #1295, opened 2026-08-23
fix(test): await schema barrier before standalone writes

hanahmily:
## Summary
- After standalone schema preload, wait for stream/measure/trace schemas and block on `SchemaBarrierService.AwaitSchemaApplied` (with min revisions) so liaison caches are ready before tests write.
- Assert every measure/stream/trace write ack is `STATUS_SUCCEED` so schema races fail loudly instead of silently dropping points and breaking count-based queries (e.g. bydbctl `Measure Data Query` / `relative end` flake on #1294).

## Test plan
- [x] `go test ./bydbctl/internal/cmd/ -ginkgo.focus='Stream Data Query|Trace Data Query|Measure Data Query'`
- [x] `make pre-push`
- [ ] CI bydbctl + integration suites green

Made with [Cursor](https://cursor.com)


hanahmily:
Authenticated standalone CI fails because schema readiness registry calls omit username and password metadata. Pass the existing credentials through `waitForSchemaKind` and attach them to both group and per-kind registry list contexts in `pkg/test/setup/setup.go:824`.

hanahmily:
Superseded by the new Owen-authored PR (includes write-stream deadlines from Copilot review).

PR #1296, opened 2026-08-23
fix(test): await schema barrier before standalone writes

hanahmily:
## Summary
- After standalone schema preload, wait for stream/measure/trace schemas and block on `SchemaBarrierService.AwaitSchemaApplied` (with min revisions) so liaison caches are ready before tests write.
- Pass auth credentials through `waitForSchemaKind` / registry list calls so `StandaloneWithAuth` does not fail with `Unauthenticated`.
- Assert every measure/stream/trace write ack is `STATUS_SUCCEED`, and open write streams with `context.WithTimeout(flags.EventuallyTimeout)` so a stalled `Recv` cannot hang forever.

Supersedes #1295.

## Test plan
- [x] Local auth integration focus: `Query service_cpm_minute with authentication`
- [x] Local bydbctl measure/stream/trace data query suites
- [x] `make pre-push` (prior iteration)
- [ ] CI bydbctl + integration standalone remainder (incl. auth) green

Made with [Cursor](https://cursor.com)


hanahmily:
Closing: will resubmit from @owenwillison-vajra so the PR author is Owen.

PR #1302, opened 2026-08-24
Advertise BanyanDB API version 0.11

hanahmily:
## Summary

- advertise BanyanDB API version 0.11 for the schema revision fields and `SchemaBarrierService` operations added in this release
- use the same API version for external clients and internal BanyanDB node communication
- document the coordinated upgrade and rollback procedure

## Compatibility

BanyanDB 0.11 nodes require API 0.11 for internal communication. Mixed BanyanDB 0.10/0.11 clusters are intentionally unsupported, so this upgrade requires a maintenance window rather than a rolling mixed-version phase.

## Testing

- `go test -p=1 ./banyand/liaison/grpc ./banyand/queue/...`
- `bin/buf breaking api/proto --against '.git#tag=v0.10.3,subdir=api/proto'`
- `make license-check`
- `make -C api generate`


mrproliu (approved):

hanahmily:
Closing this PR so it can be replaced by an equivalent PR from the hanahmily fork.

PR #1315, opened 2026-08-28
ci: publish release images to Docker Hub

hanahmily:
## Summary
- retain SHA-tagged GHCR publishing for each `main` commit
- publish release images to Docker Hub only through a manual `publish-docker` run against an annotated tag
- publish default, slim, Canopy, MCP, FODC agent, and FODC proxy release images

## Validation
- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.7 .github/workflows/publish-docker.yml`
- `make -n` for all six release image targets

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -27,9 +27,14 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: Release tag to publish, for example v0.11.0
+        required: true
+        type: string
 
 env:
-  HUB: ghcr.io/apache
   SEGMENT_DOWNLOAD_TIMEOUT_MINS: 5
 
 jobs:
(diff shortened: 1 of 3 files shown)


hanahmily:
This enables the already-published v0.11.0 release to be backfilled through `workflow_dispatch`, then makes future GitHub releases publish Docker Hub images automatically. The PR workflow is validating both default/slim and plugin image builds without publishing.

wankai123 (dismissed):

hanahmily:
Updated after comparing Docker Hub release 0.10.3: its exact release set is only the default and `-slim` tags. This revision deliberately does not publish plugin, FODC, MCP, bydbctl, testing, or rolling `latest` tags to Docker Hub.

wankai123 (dismissed):

hanahmily:
Added the missing Canopy release image. The release set is now the default, `-slim`, and `-canopy` tags; plugin/FODC/MCP/bydbctl/testing variants remain GHCR development-only images.

wankai123 (dismissed):

hanahmily:
Comparison with 0.10.0 found the historical MCP and FODC agent/proxy images, which were absent from 0.10.3. The release workflow now publishes those three in addition to default, slim, and Canopy. It still excludes development-only testing, bydbctl, and plugin images.

hanahmily:
Fixed the failed `publish-docker` PR build: Canopy does not define a default `PLATFORMS` value, so its PR-time local build invoked Buildx with an empty platform. The workflow now explicitly builds it as `linux/amd64`; release publishing already uses `linux/amd64,linux/arm64`.

hanahmily:
Updated trigger semantics: a `main` commit continues to publish SHA-tagged development images to GHCR. Docker Hub publishing has no automatic release trigger; it runs only when a release manager manually dispatches `publish-docker` with an annotated `v<version>` tag.

wankai123 (approved):

mrproliu (approved):

PR #1316, opened 2026-08-28
feat(index): count committed generations with deletions in the native reader

### LINKED ISSUE #14009: [NIDX-01B] Count committed generations with deletions
# [NIDX-01B] Count committed generations with deletions

Parent workstream: #14002
Blocked by: #14008

## Summary

Extend the live native `inverted.ReadOnlyDocCount` cutover from NIDX-01A to multiple segments, snapshot deletion masks, newest-valid generation selection, and concurrent legacy publication. This slice adds no document visitation or query surface.

## Blocked by

NIDX-01A (#14008) must be merged before this work starts. An open pull request is insufficient.

## Pre-agreed test seam

Test only:

```go
func inverted.ReadOnlyDocCount(path string) (int64, error)
```

Generation enumeration, snapshot records, roaring deletion masks, and segment accounting remain private.

## Independent fixture

Extend the legacy corpus with:

- valid generation `10`, referencing two segments containing five physical documents;
- one snapshot deletion, for `doc-22`;
- invalid newer generation `11`, referencing a missing segment; and
- arbitrary reserved CRC32 bytes.

The declared selected generation is `10`; the visible count is the literal `4`.

## Requirements

R1. `ReadOnlyDocCount` selects the newest structurally complete generation and returns `4`. It never combines records from generations `10` and `11`.

R2. Count is computed from segment counts and deletion masks without decoding stored fields or materializing document results.

R3. While a retained legacy writer publishes a complete generation adding `doc-26`, concurrent count calls return only `4` or `5`. They never observe a partial generation, create an index-local runtime file, or fail because the writer is open.

R4. Missing segments, invalid deletion ordinals, truncated roaring data, and arithmetic overflow produce bounded typed errors. Reserved CRC32 values remain ignored.

- Boundary RED: the NIDX-01A implementation is exercised against generation `10`/`11`; it must fail until multi-segment deletion and generation selection are implemented.
- E2E RED: keep the legacy writer open, publish `doc-26` while repeated production counts run, and assert the observed set is a subset of `{4,5}` and eventually contains `5`; directory readers never expose an intermediate count.

## Acceptance criteria

- The extended fixture and provenance manifest declare generation IDs, segment membership, the deleted ordinal, file hashes, and expected count `4` independently of native code.
- `go test ./pkg/index/inverted/... -run 'Test(ReadOnlyDocCountCommittedGenerations|E2EReadOnlyCountDuringPublication)' -count=1` passes, including under `-race`.
- Before/after inventories prove the reader changes no bytes, mtimes, directory entries, or runtime files.
- `make test-ci PKG=./pkg/index/inverted/...`, `make build`, and `make pre-push` pass.

## Scope

Packages: `pkg/index/inverted` and its private native snapshot/deletion reader.

Out of scope: stored fields, dictionaries, term postings, document walk, doc values, sort/search-after, writers, and merge.

## Compatibility and rollback

This slice emits no index bytes. Roll back `ReadOnlyDocCount` to its retained legacy reader without a writer drain or directory conversion.

## Lexical non-regression

## Design

hanahmily:
## What this changes

BanyanDB reports a closed segment's series-index document count by opening the
index directory read-only, without reopening the writable index or taking its
exclusive lock. That read-only path now goes through our own ICE v3 / snapshot
v3 reader, but so far it only understood the simplest directory there is: one
committed generation over one segment with nothing deleted.

Real directories are not like that. A series index accumulates several segments,
records deletions as per-segment masks in the snapshot manifest, and — because
the persister streams each file into its final name rather than renaming an
atomically completed one — spends part of its life holding a newest generation
that is not yet whole. A reader that always pins the highest-numbered manifest
sees those moments as damage. It reports a corrupt index for a directory that is
perfectly healthy and merely busy.

This change teaches the read-only counter to walk committed generations
newest-first and pin the newest one that is *structurally complete*: every
segment the manifest references is present and parses. A generation that fails
that check is skipped in favour of the one behind it, which is the generation a
reader would have seen a moment earlier and is by construction a consistent
view. The typed corruption error is now reserved for what it was always meant to
mean — there is no complete generation to read at all. The count itself is the
sum of the referenced segments' recorded document counts less the documents
their deletion masks mark; no stored field is decoded and no document is
materialised on the way to an integer.

```mermaid
flowchart TD
    A["ReadOnlyDocCount(path)"] --> B[List committed manifests]
    B --> C{Any manifest?}
    C -- no --> D["ErrNoCommittedIndex"]
    C -- yes --> E[Take newest not yet tried]
    E --> F{Manifest parses and<br/>every referenced<br/>segment is present?}
    F -- "no: publication in flight,<br/>or damaged" --> G{Older generation left?}
    G -- yes --> E
    G -- no --> H["ErrCorruptIndex"]
    F -- yes --> I[Pin this generation]
    I --> J["sum(segment doc counts)<br/>- sum(deletion masks)"]
    J --> K[Visible count]
```

The practical effect is that `SeriesIndexStats` keeps reporting a number while a
segment is being written, instead of intermittently logging that the index is
unreadable, and that the number it reports honours deletions.

## Compatibility

Nothing here writes index bytes, so there is no format change and no migration.
Directories written by earlier releases open unchanged and report the counts they
always reported; reverting is a code change alone, with no writer drain, no
replay and no directory conversion. The reserved CRC32 fields stay carried and
unchecked — they take no part in the count or in classifying corruption.

## Verification

The contract is pinned against a checked-in corpus produced through the store
boundary itself: two segments holding five physical documents, one snapshot
deletion, and a newer generation whose referenced segment is deliberately absent
from the directory. A provenance manifest declares the generation identifiers,
segment membership, physical counts, the deleted ordinal and every file hash, so
the expected answers come from the corpus rather than from the reader being
tested. Restoring the withheld segment moves the answer forward a generation,
which is what distinguishes "pick the newest valid one" from "pick an old one
and hope".

Alongside that, an end-to-end test keeps a writer open and publishing while
counts run in a separate process, and asserts every observation is one of the
two legitimate answers and never an intermediate or an error. Bounded, typed
failures are checked for a missing segment with nothing behind it, an
out-of-range deletion ordinal, a truncated deletion payload and an unsatisfiable
document count, each in a child process so that a panic, an abort or a hang
fails as an assertion rather than taking the suite down.

Fixes apache/skywalking#14009


hanahmily:
The FODC UTC check was cancelled on its first attempt after an unrelated proxy integration test failed and hung during cleanup; this PR only changes native inverted-index code. Please rerun the cancelled check before treating it as a PR failure.

wu-sheng (approved):

PR #1322, opened 2026-09-01
Bound the vectorized stream Top-N merge by limit+offset

SammyUrfen:
`SortedMerge` in the vectorized stream pipeline (`scan -> merge -> distinct -> limit`) buffered a row reference for every scanned row, stable-sorted the entire set in `Finalize`, and applied the distinct-`ElementID` cap only afterwards. For a query with `LIMIT 20` over a wide time range, both the merge state and the sort cost scaled with the number of rows scanned rather than with the limit. The row path does not behave this way: its heap merge stops once it has collected N distinct elements.

### Approach

`Consume` now prunes the buffer as it fills. Once it grows past twice the cap it sorts in place and re-applies the cap, keeping only the first row of each retained `ElementID`.

Dropping the later duplicates instead of passing them to the downstream `Distinct` is what makes the bound hold. The cap can only fire when a new distinct `ElementID` appears, so on a corpus whose distinct cardinality sits just above the cap, one prune removes the boundary row and it is not re-established for thousands of rows. Every duplicate in that window would otherwise accumulate, and each one pins its source `RecordBatch`. Measured over 100k rows with cap 20 and cardinality 21, carrying the duplicates retained 987 references pinning 94 of 98 batches; compacting retains 20 references pinning 1.

Since every prune now leaves at most `maxRows` rows, the threshold no longer needs a backoff and that logic is removed.

### Results

100k rows, batch size 1024, `-benchtime=20x -count=6`, medians:

| Benchmark | Before | After |
| --- | --- | --- |
| `SortedMergeTopN` (LIMIT 20) | 13,011,532 B/op, 178 ms/op | 170,722 B/op, 64 ms/op |
| `SortedMergeTopNDuplicateHeavy` | 16,750,484 B/op, 67 ms/op | 168,706 B/op, 35 ms/op |
| `SortedMergeUncapped` (control) | 13,090,242 B/op, 183 ms/op | 13,090,243 B/op, 181 ms/op |

The uncapped control is unchanged byte for byte, which is the check that the filtered path has not been affected.

### Requirements covered

The size audit on the issue classifies it as a tracking parent whose first deliverable is R1/R2 for criteria-less queries, with R3 as a follow-up leaf. This PR is that first deliverable, not the whole issue.

**R1 — delivered for criteria-less queries**, subject to the limitation below. Merge state is O(L + O) row references instead of O(N), and the element set and order are unchanged.

**R2 — delivered.** `SortedMerge` no longer sorts N rows when `mergeCap > 0`. It sorts at most `maxRows + batchSize` rows at a time, so total work is O(N log(maxRows + batchSize)). Stable tie-breaking on the arrival sequence and the distinct-`ElementID` cap semantics are preserved.

**R3 — not delivered, and not feasible as currently scoped.** Index-order queries wrapped by a tag filter continue to merge uncapped. Their filter runs at egress, after the merge, so a cap applied here would starve it: the filter's selectivity is unknown, so the top `limit + offset` rows before it can contain arbitrarily few surviving rows, while the row path keeps pulling until it has enough. Bounding that path requires columnar tag-filter pushdown ahead of the merge, which the issue lists as out of scope. The parity constraint is now stated in `localIndexScan.ExecuteVectorized`, where the decision is made. This remains the follow-up leaf.

**R4 — green.** `TestSortedMergeCapKeepsInOrderTopN`, `TestSortedMergeCapCountsDistinctElementIDs` and `TestSortedMergeHugeCapDoesNotOverAllocate` pass unchanged, as do the stream vectorized parity and integration tests.

The RED test the size audit asks for is `TestSortedMergePruneBoundsRetainedState`. It asserts retained state on a synthetic pipeline with N far greater than the cap, and fails against the previous implementation, where `len(SortedMerge.rows)` reaches N before the cap is applied.

On the callers listed in the issue boundary: `localIndexScan.ExecuteVectorized` is the only non-test caller of `BuildStreamMergePipeline` and is updated here. `banyand/query/processor.go` reaches the merge only through it, and its `applyStreamTagFilter` egress sits on the uncapped path, which is unchanged. The distributed liaison merge path is covered by `test/integration/distributed/query`.

### Testing

`TestSortedMergePruneMatchesFullSort` is a differential oracle: over a corpus large enough to prune repeatedly, seeded with colliding timestamps and repeated `ElementID`s, the capped pipeline must return exactly what an uncapped merge returns truncated to the first `mergeCap` distinct elements. It sweeps three seeds, four caps, two cardinalities and both sort directions.

`TestSortedMergePruneBoundsRetainedState` sweeps distinct-`ElementID` cardinality across `cap+1`, `cap+5`, `cap/2` and all-distinct. Its corpus is shuffled deliberately: with the scan already in sort order every survivor comes from the newest batch, and the assertion would hold at 1 regardless of what the implementation does. It fails against the previous implementation at cardinalities 21, 25 and 10.

Also run locally: `go test -race` over `pkg/query/vectorized/...`, `banyand/stream/...` and `pkg/query/logical/stream/...`; `test/integration/standalone/query` and `test/integration/distributed/query`; `golangci-lint` over the project-scoped paths CI lints; `license-eye header check`; and the format configuration.

There are no on-disk format or wire protocol changes.

### Known limitation

The bound is on merge state. Each retained reference still pins its whole source `RecordBatch`, so worst-case retained rows are `O(min(N, maxRows * batchSize))` rather than `O(maxRows)` — one batch when the scan arrives in sort order, up to `maxRows` batches when scan order is uncorrelated with it. Tightening that further means copying surviving rows into a merge-owned batch, which is a larger change and is not required to remove the growth in N that this issue is about. I am happy to follow up with it if reviewers would rather have the stricter bound.

Addresses R1, R2 and R4 of apache/skywalking#14050. R3 stays open as the follow-up leaf that issue's size audit identifies, so this does not close it.


pkg/query/vectorized/stream/sorted_merge.go
```diff
@@ -74,15 +75,36 @@ func NewSortedMerge(schema *vectorized.BatchSchema, desc bool, batchSize int) *S
...
+// The cap is enforced INCREMENTALLY: Consume prunes the buffer back to the
+// in-order top-N whenever it grows past 2*maxRows, so merge state stays O(maxRows)
+// rows — plus the batch being consumed — and sort work is O(N log maxRows) rather
+// than O(N log N).
+//
+// This yields the same QUERY result as capping after one full sort (the emitted
+// rows are the deduped top-N rather than a duplicate-carrying prefix of it, which
+// the downstream Distinct makes indistinguishable). A row dropped by a prune
+// already sits behind maxRows distinct ElementIDs, and rows consumed later can
+// only insert AHEAD of it (pushing it further past the cap), never behind it;
```
hanahmily: This proof is not quite accurate: rows consumed later can sort either ahead of or behind a dropped row. The monotonic argument is that later rows ahead can only worsen the dropped row's distinct rank, while later rows behind cannot improve it. Since this is the correctness justification for destructive pruning, could you reword it accordingly?

pkg/query/vectorized/stream/sorted_merge_test.go
```diff
@@ -208,6 +211,129 @@ func TestSortedMergeHugeCapDoesNotOverAllocate(t *testing.T) {
...
+)
+
+// runPruneParityCase asserts one (seed, cap, cardinality, direction) of the
+// prune differential oracle.
+func runPruneParityCase(t *testing.T, seed int64, mergeCap, idCardinality int, desc bool) {
+	t.Helper()
+	name := fmt.Sprintf("seed%d/cap%d/ids%d/desc%v", seed, mergeCap, idCardinality, desc)
+	rng := rand.New(rand.NewSource(seed)) //nolint:gosec // fixed seed: a failing case must reproduce exactly
+	rows := make([]testRow, pruneParityRows)
+	for i := range rows {
```
hanahmily: The repository's AGENTS.md asks for descriptive range-loop names rather than single-letter indices. Could this be rowIdx? The same applies to i at line 309, the i/j shuffle indices at line 313, and b at line 316.

pkg/query/vectorized/stream/sorted_merge_bench_test.go
```diff
@@ -0,0 +1,133 @@
...
+	benchTopN      = 20
+)
+
+// benchBatches builds a fixed corpus of benchRows rows, shuffled so the merge
+// cannot exploit an already-ordered input, split into benchBatchSize batches.
+// The corpus is built once per benchmark and replayed every iteration so the
+// reported allocations are the merge pipeline's, not the fixture's.
+func benchBatches(schema *vectorized.BatchSchema) []*vectorized.RecordBatch {
+	rows := make([]testRow, benchRows)
+	for i := range rows {
```
hanahmily: Please use descriptive indices here as required by AGENTS.md, for example rowIdx. The same applies to the shuffle callback's i/j and the duplicate-heavy fixture loop at line 86.

pkg/query/vectorized/stream/sorted_merge_bench_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+// BenchmarkSortedMergeUncapped is the uncapped control: the same corpus with
+// mergeCap 0 (the filtered index-order path), which must keep buffering every
+// row. It bounds how much of the capped case's cost is inherent to the scan.
+func BenchmarkSortedMergeUncapped(b *testing.B) {
+	schema := tsSchema()
+	batches := benchBatches(schema)
+	b.ReportAllocs()
+	b.ResetTimer()
+	for range b.N {
```
hanahmily: The three benchmarks repeat pipeline construction, draining, result validation, and cleanup. Could we extract a small benchmark helper parameterized by merge cap and expected row count so these paths do not drift?

hanahmily (changes requested):

pkg/query/vectorized/stream/sorted_merge.go
```diff
@@ -116,11 +139,19 @@ func (s *SortedMerge) Consume(_ context.Context, batch *vectorized.RecordBatch)
...
+			s.arriveSeq++
+		}
 	}
-	for _, sel := range batch.Selection {
-		s.rows = append(s.rows, mergeRowRef{batch: batch, row: int(sel), arrive: s.arriveSeq})
-		s.arriveSeq++
+	// Every prune leaves at most maxRows rows (capByDistinctElementID compacts to
+	// one row per retained ElementID), so the buffer never exceeds maxRows plus
+	// one batch and the threshold needs no backoff: prunes stay O(maxRows +
+	// batchSize) work each, amortized O(N log(maxRows + batchSize)) overall.
```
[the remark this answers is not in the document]
SammyUrfen: Correct, thanks. The threshold is only checked after a whole batch has been appended, so when batchSize is smaller than maxRows the buffer keeps growing across several batches before a prune fires. Measured at cap 1000 with batch size 64: the peak is 1984, against a 2*maxRows+batchSize bound of 2064. Comment corrected in 3ccf4b0b.

pkg/query/vectorized/stream/sorted_merge.go
SammyUrfen: Reworded in 3ccf4b0b. It now says that rows consumed later may sort to either side of a dropped row, and that neither side rescues it: rows sorting ahead only add to the distinct ElementIDs preceding it, and rows sorting behind cannot remove any, so its distinct rank never improves.

pkg/query/vectorized/stream/sorted_merge_test.go
SammyUrfen: Renamed in 3ccf4b0b: rowIdx for the two fixture loops, leftIdx/rightIdx for the shuffle callback, and batchIdx for the consume loop. I also renamed the two single-letter indices in sorted_merge.go that this change already touches. The pre-existing loops at lines 63 and 142 are left alone to keep the diff scoped, happy to include them if you would rather they were consistent.

pkg/query/vectorized/stream/sorted_merge_bench_test.go
SammyUrfen: Done in 3ccf4b0b, rowIdx and leftIdx/rightIdx throughout. The duplicate-heavy fixture no longer has its own loop, it now goes through the shared corpus builder described in the next thread.

pkg/query/vectorized/stream/sorted_merge_bench_test.go
SammyUrfen: Extracted in 3ccf4b0b: runMergeBenchmark(b, schema, batches, mergeCap, wantRows) drives the pipeline, and benchCorpus(schema, cardinality, shuffle) builds the fixture. Each benchmark is now two lines and the file is 35 lines shorter. Benchmark numbers are unchanged after the refactor (170,722 / 168,712 / 13,090,244 B/op).

SammyUrfen:
Pushed 3ccf4b0b with all five points addressed, rebased on the current branch head.

One change beyond the review comments. Since the general buffer bound is 2*maxRows + batchSize, the mergeCap+batchRows assertion in TestSortedMergePruneBoundsRetainedState is only exact for that test's configuration, where batchRows exceeds mergeCap so every Consume prunes. I kept the tighter assertion, because relaxing it to the general bound stops the test catching the cardinality-25 case, and added a comment recording why it is configuration-specific.

Green on the rebased branch: the package tests under -race, banyand/stream and pkg/query/logical/stream under -race, golangci-lint on the pkg scope, revive, license-eye, gofmt, and both test/integration/standalone/query and test/integration/distributed/query.

hanahmily (approved):

hanahmily:
Thanks for landing R1/R2/R4 here.

R3 (bounded merge for **filtered** index-order Stream vec queries) is tracked as apache/skywalking#14056 — it needs pre-merge / columnar tag-filter pushdown, which was correctly out of scope for this PR.

@SammyUrfen if you are interested in owning a follow-up, that issue is the place to start (suggested first leaf is sketched in the size audit). No pressure either way.

PR #1325, opened 2026-09-02
feat(index): select stored documents by exact term and cut schema walks over to it

### LINKED ISSUE #14011: [NIDX-01D] Cut over exact-term schema walks
# [NIDX-01D] Cut over exact-term schema walks

Parent workstream: #14002
Blocked by: #14010

## Summary

Extend the native stored-document visitor with exact term and exact-term OR selection, then move `metadata/schema/reader.WalkShard` and `WalkDocs` to it. This slice adds the ICE term dictionary and postings intersection needed by schema kinds; it does not add ranges, prefix/wildcard expansion, analyzers, or sorting.

## Blocked by

NIDX-01C (#14010) must be merged before this work starts. Before applying `Backlog`, update identifier names in this body if NIDX-01C's merged visitor uses different names; do not change the behavioral seam.

## Pre-agreed test seam

The product seam is the existing:

```go
func reader.WalkShard(shardPath string, visit func(reader.Doc) error, kinds ...schema.Kind) error
func reader.WalkDocs(schemaRoot string, visit func(reader.Doc) error, kinds ...schema.Kind) error
```

The native visitor may gain exactly one bounded “field equals any of these literal terms” input. Do not expose a general query language, FST type, posting iterator, or analyzer. Boundary tests observe `WalkShard`/`WalkDocs` documents and errors, not dictionary internals.

## Independent fixture

The legacy schema corpus contains:

- `p1`, kind Stream, revisions `1` and `2`;
- `p2`, kind Measure, revision `1`;
- `p3`, kind Stream, revision `1` followed by a tombstone at revision `2`;
- `p4`, kind Group, revision `1`; and
- `p5`, kind Measure, with deliberately malformed stored `_source` bytes.

For kinds `{Stream}`, the result is only `p1@2`. For `{Stream, Group}`, results are `p1@2` and `p4@1`. `p3` is absent. A Stream-only walk must not decode `p5`; a Measure walk must return
`reader.ErrMalformedPropertyDocument` for it. Because the ICE container remains valid, the failure must not be classified as native index corruption.

## Requirements

R1. Exact lookup resolves only the requested field/term. Exact-term OR unions the requested postings, applies the selected snapshot's deletions, and visits each live document once.

R2. `WalkShard` and `WalkDocs` return exactly the fixture results above, preserving `_source` bytes, latest-revision selection, and tombstone suppression.

R3. Filtering happens before stored-field decode: Stream-only load succeeds despite Measure `p5`'s malformed `_source`; selecting Measure reports `reader.ErrMalformedPropertyDocument`, not the native index-corruption sentinel.

R4. Term count, term length, dictionary enumeration, posting decode, and cancellation are bounded. CRC32 remains ignored and no reader lock or directory mutation is introduced.

- Boundary RED: exercise exact term and two-term OR through the bounded native visitor inherited from NIDX-01C; the compile-time or unsupported-operation failure proves the new slice is RED.
- E2E RED: run `WalkDocs`/schema loaders for Stream, Stream+Group, and malformed Measure against the checked-in legacy corpus and assert the declared results and typed failure. The command must fail before the production caller is switched.

## Acceptance criteria

- `go test ./pkg/index/inverted/... ./banyand/metadata/schema/reader -run 'Test(NativeExactTerms|E2ESchemaWalkNative)' -count=1` passes.
- A call-graph/dependency assertion proves schema walks no longer build or execute Bluge queries.
- Exact, OR, deletion, corrupt-unselected-field, cancellation, and resource-limit cases pass under `-race`.
- `make test-ci PKG='./pkg/index/inverted/... ./banyand/metadata/schema/reader'`, `make build`, and `make pre-push` pass.

## Scope

Packages: `pkg/index/inverted`, its private ICE dictionary/posting reader, and `banyand/metadata/schema/reader`.

Out of scope: numeric/date range, prefix, wildcard, MATCH, NOT, sorting, doc values, Property repair, and every writer.

## Compatibility and rollback

This slice reads historical bytes and emits none. Roll back schema readers to the retained Bluge path without changing any directory.

## Lexical non-regression

## Design

hanahmily:
# Cut over exact-term schema walks (NIDX-01D)

## What this delivers

The offline schema catalog reader — the code a migration or a recovery tool
uses to load stream, measure, group, index-rule and binding schemas straight
out of a `_schema` directory, with no schema server running — stops building
and executing search queries against the retained legacy reader. It now reads
the same historical bytes through BanyanDB's own bounded read-only index
reader.

Doing that needed exactly one new capability from that reader, and this change
adds only that one: given a field name and a set of literal terms, stream the
live documents whose field records any of those terms. The catalog walk uses it
to resolve the schema kinds a caller asked for.

```go
type TermSelection struct {
	Field string
	Terms [][]byte
}

func ReadOnlySelectDocuments(ctx context.Context, path string, selection TermSelection,
	visit func(doc StoredDocument) error) error
```

The terms are matched as raw bytes: no analysis, no normalization, no ranges,
no prefix or wildcard expansion. Their postings are unioned, the pinned
generation's deletion masks are removed from that union, and each surviving
document is handed to the caller exactly once however many terms selected it.
The term dictionary, the posting records and the union all stay private to the
reader's internal package; nothing resembling a query language, a posting
iterator or an analyzer is exposed.

## Why it matters

Three properties of the previous arrangement were the reason to move.

The catalog reader linked a whole search engine to answer "give me the
documents whose kind is one of these". Reading the catalog is a
read-only operation over a handful of stored payloads, and it now costs a
bounded read of one committed generation instead.

Failures were not classifiable. A malformed stored schema payload surfaced as
whatever the search layer happened to produce, so a migration could not tell
"this catalog was never flushed" from "this schema payload is malformed" from
"this schema kind is simply absent". The three are now distinct outcomes: an
absent committed generation, `reader.ErrMalformedPropertyDocument`, and an empty
result that is not an error at all. Malformed application JSON is not reported
as ICE container corruption.

Filtering now precedes decoding. A walk for stream schemas never touches the
stored bytes of a measure revision, so one malformed payload no longer fails a load that never needed it — while a
walk that *does* ask for that document still fails loudly rather than returning
a partly-read catalog.

## How the read flows

```mermaid
flowchart TD
    A["migration / recovery caller"] --> B["reader.WalkDocs<br/>(schemaRoot, visit, kinds...)"]
    B --> C["reader.WalkShard<br/>per shard-* directory"]
    C --> D{"kinds requested?"}
    D -- "no kinds" --> E["ReadOnlyWalkDocuments<br/>match-all"]
    D -- "one or more kinds" --> F["ReadOnlySelectDocuments<br/>TermSelection{Field: kind field, Terms: kinds}"]
    E --> G["pin newest structurally<br/>complete committed generation"]
    F --> G
    G --> H["resolve exact terms in<br/>the field's term dictionary"]
    H --> I["union the selected postings"]
    I --> J["subtract the generation's<br/>deletion masks"]
    J --> K["decode stored fields<br/>of surviving documents only"]
    K --> L["Doc{PropID, KindName, Group,<br/>SourceJSON, ModRev, Deleted}"]
    L --> M["latest revision wins;<br/>tombstoned IDs suppressed"]
    M --> A

    N["malformed stored source<br/>of an unselected document"] -.-> J
    N -.->|"never decoded"| K
```

Callers see no signature change. `WalkShard` and `WalkDocs` keep taking a set
of schema kinds and keep handing back `Doc` values, so the loaders above them —
and the migration paths above those — are untouched.

## Scope

Added: one bounded term selection on the read-only reader, and the schema
catalog cutover onto it.

Not added: numeric and date ranges, prefix and wildcard expansion, analyzers
and MATCH, negation, HAVING, IN, scoring, doc values, sorting and search-after,
property repair, and every writer. No new on-disk version and no migration at
open time.

## Compatibility and rollback

This change reads historical bytes and emits none. Reads pin open handles for every segment in the newest structurally complete
committed generation, acquire no exclusive writer lock,
and leave file contents, modification times and directory entries untouched —
verified by inventorying the corpus before and after. A directory a live writer
owns can be read while it is being written. The reserved CRC32 fields stay
carried bytes that are never calculated, validated or used to classify
corruption.

Rolling back is choosing the retained legacy reader in the schema walks. No
drain, replay, conversion or file change is involved, because nothing was
written.

## Evidence

A checked-in catalog corpus holds seven physical revisions of five logical
properties in one shard: a stream property at two revisions, a measure
property, a stream property whose latest revision is a tombstone, a group
property, and a measure property whose stored `_source` is deliberately malformed. The
ICE container itself is untouched compatibility-writer output. The fixture is
pinned by per-file content hashes and a provenance manifest that records the
exact malformed application bytes,
and it is never regenerated by a normal test run.

Against it, a `{Stream}` walk yields one document, a `{Stream, Group}` walk
yields two, the tombstoned property is absent, the stream walk succeeds despite
the malformed measure revision, and a measure walk reports
`reader.ErrMalformedPropertyDocument` without misclassifying the ICE segment as
corrupt. Beside that, the corpora checked in by the preceding read-only slices
still resolve their identities by exact term, and a generation left
structurally incomplete by a crash is skipped rather than half-read.

Two structural assertions keep the cutover honest rather than merely correct:
the catalog reader's third-party imports and every name it takes from the index
package are on explicit allowlists, so it cannot go back to building queries;
and the read-only reader's own exported surface is compared against a declared
list, so this slice can add the exact-term selection and nothing else.

Fixes apache/skywalking#14011


hanahmily:
Addressed the review findings in `53ab026b`:

- pin open segment handles for the reader lifetime, including cleanup on partial-open failure and idempotent close;
- make posting decode and bitmap union cancellation-aware;
- consolidate read-only open/close handling and add operation context to open failures;
- carry field/terms internally as one selection request;
- regenerate NIDX-01D with an actually malformed `_source` in an otherwise valid ICE segment, classified as `reader.ErrMalformedPropertyDocument` rather than index corruption;
- add deterministic regression coverage for unlink-after-open, cancellation during decode/union, handle cleanup, and error classification.

Verified locally with the milestone acceptance test under `-race`, targeted golangci-lint, and `make build`. The CI Prepare/Generate and vulnerability checks are green; the remaining CI matrix is running.

mrproliu (approved):

PR #1326, opened 2026-09-03
chore: remove the row-based query path from measure, stream and trace

SammyUrfen:
# chore: remove the row-based query path from measure, stream and trace

Related to apache/skywalking#13998.

Vectorized query has been the default for measure, stream and trace since 0.11. This removes the row path it replaced: dispatch branches, plan nodes, storage-layer egress, and the tests that existed only to compare the two.

**-6,917 lines** (10,608 deleted, 3,691 added). Non-test Go: -4,849.

| Area | Net | What went |
| --- | ---: | --- |
| measure | -2,535 | row query planner, row distributed plan, row fallback in `dquery` |
| stream | -1,187 | row execution in dispatch, storage-layer egress (`query_by_ts.go`, `query_by_idx.go`, `tsResult` / `idxResult`) |
| trace | -800 | row query dispatch, SIDX row streaming pipeline |
| shared | -257 | `model.StreamResult` / `StreamQueryResult` / `MergeStreamResults` |
| other | -70 | codec, querybench harness |

`(*stream).Query` and `logicalstream.BuildElementsFromStreamResult` have no callers outside tests.

## Scope

This covers the row dispatch branches, the row logical plans, and the stream storage-layer egress. It does not make the vectorized engine batch-native end to end, so apache/skywalking#13998 stays open after this merges. Four seams still carry row currency inside the vectorized path:

- criteria and hidden-order-tag stream queries materialize Elements before filtering
- multi-group measure merges through `sortedMIterator` rather than a batch-native merger
- schema-evolved multi-group aggregates are rejected by the partial-batch reducer
- the stream distributed coordinator decodes frames back into Elements to merge and dedup

None of their behaviour changes here. `banyand/measure/query_batch.go`, `pkg/query/logical/measure/cross_group_merge.go`, `pkg/query/vectorized/measure/reduce.go`, `pkg/query/vectorized/measure/plan/distributed.go` and `pkg/query/logical/stream/stream_plan_distributed.go` are byte-identical to `v0.11.0`. Two things do move: the stream vec plan node takes over the Element materialization from the deleted `tagFilterPlan.Execute`, and `sortedMIterator` moves verbatim out of the deleted `measure_plan_distributed.go`.

## Behaviour changes

The first three are in `docs/operation/upgrade.md` under 0.12.

**Stream timestamp-order queries with criteria, spanning multiple segments, can come back under-filled.** The row scan resumed segment by segment, pulling from later segments until the limit filled. The vectorized scan does not see segment boundaries, so it caps at `maxElementSize` and stops. Results are correct and correctly ordered, just short of the limit. Index-order queries keep their uncapped merge. Clients can narrow the range or page forward. Since the flags no longer select an engine, this is not avoidable by configuration; rollback is a binary downgrade.

**Index-order queries that do not project their ordering tag** now egress via proto instead of a frame. The scan projects that tag internally for its sort key and strips it before egress, while the frame egress would rebuild its projection from the batch schema and leak it. Same results, no raw-wire fast path for that shape.

**A stream query the vectorized engine cannot plan now fails** instead of falling through to row. One shape reaches this: an `orderBy` naming a tag that no longer resolves against the schema. The error names the reason.

**Stream slow-query logging is gone.** It lived only on the row path after `se.Execute`, and vec dispatch returned before reaching it, so at the shipped default it was already not logging for stream. Not a regression. Both measure processors and `logSlowTraceQuery` still log.

## Flags

`--measure-vectorized-enabled`, `--stream-vectorized-enabled` and `--trace-vectorized-enabled` stay registered as no-ops. `=false` fails fast at startup on standalone, data and liaison nodes with an error naming the issue; `=true` and omission are accepted. Removing the `BoolVar`s outright breaks node startup wherever the flag is still passed: CI, e2e, soak, and the surviving vec integration tests.

## Tests

Test Go is down 3,955 lines, about 3,300 of it code whose subject is gone: `model_test.go`, both `stream/query_by_*_test.go`, and trims of `trace/streaming_pipeline_test.go`, `stream/benchmark_test.go` and the row measure plan tests. The eight stream parity tests split: four became golden fixtures captured from the row path before deletion (1,921 lines of YAML under `banyand/stream/testdata/`), and four went without direct replacements because the 16 `gen_feat_*` goldens in `test/cases/stream/data/want/` cover the same shapes end-to-end. Counting the fixtures, effective coverage delta is about -2,034.

None of the four fixtures is a multi-segment timestamp-order query, so there is no golden for the one case where vec deliberately differs.

`banyand/measure/query_batch_top_test.go` is new. It pins the gate that keeps an ordinary `Top` query on the batch path and drives `PullBatch` over a 100-block, 100k-row result. Forcing the row fallback fails it, so it pins the path rather than the answer.

`test/integration/distributed/querybench/` was a row-vs-vec harness. With no engine accepting `mode=row` the mode dimension collapses to vec-only and `DQB_MODE=row` fails validation. Latency and QPS measurement is unchanged; the ratio table is gone.

## Test plan

`make generate`, `build`, `lint` and `check` pass. `make test-ci` passes across `./banyand/...` (42 suites), `./bydbctl/...` (14), `./pkg/...` (58), `./fodc/...` (22), `./test/integration/standalone/...` (15) and `./test/integration/distributed/...` (13).

Built back-to-back with `make build` and a cold cache, binaries total 501,696,590 to 500,990,918, **-705,672 bytes**. Almost all of it is `banyand-server` (-0.84%); the other seven barely link the query path.

Two pre-existing failures, both reproduced on a clean `origin/main` worktree:

- `make test-race` fails `banyand/trace` *Schema Change → querying data should return null for type-mismatched tags* on a 10 s Ginkgo timeout. Same failure on main at 12.5 s; passes under `make test-ci` on both.
- `make lint-rawgo` fails with 140 violations here against 143 on main. One baseline entry pointed at `banyand/stream/query_by_idx.go:139`, which this PR deletes. That line was removed by hand rather than regenerating the file, so the baseline only shrinks (263 → 262).

---

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Related to apache/skywalking#13998.


banyand/query/processor.go
```diff
@@ -152,52 +152,32 @@ func (p *streamQueryProcessor) Rev(ctx context.Context, message bus.Message) (re
...
-	}
-
-	se := plan.(executor.StreamExecutable)
-	defer se.Close()
-	entities, err := se.Execute(ctx)
-	if err != nil {
-		p.log.Error().Err(err).RawJSON("req", logger.Proto(queryCriteria)).Msg("fail to execute the query plan")
-		resp = bus.NewMessage(bus.MessageID(now), common.NewError("execute the query plan for stream %s: %v", queryCriteria.GetName(), err))
+	// Vec dispatch is the only stream execution path; the row path was removed in
+	// 0.12.0. Every analyzed plan shape must therefore be vec-eligible, so a decline
```
hanahmily, line 156: **[P1] The row path is not fully removed yet.** `banyand/measure/query_batch.go:88-115` still routes ordinary Top queries spanning multiple blocks through `qr.merge` (the row `MeasureResult`/postprocessor path) and then converts the result back into a batch. Please remove that runtime fallback before making this unconditional claim, and add a multi-block Top regression test that proves `PullBatch` stays batch-native.

banyand/query/processor.go
```diff
@@ -233,7 +213,7 @@ func (p *streamQueryProcessor) tryStreamVecDispatch(ctx context.Context, plan lo
 	// so the send path passes it through and the liaison decodes it. The liaison
 	// applies the global offset/limit slice, so the data node emits the whole
 	// (already per-node-capped) batch set — no slice here.
-	if !hasFilter && streamVecEmitAsFrame(p.distributed, data.StreamWireModeRaw(), traced) {
+	if !hasFilter && !vecExec.HidesOrderTag() && streamVecEmitAsFrame(p.distributed, data.StreamWireModeRaw(), traced) {
```
hanahmily, line 226: **[P1] This gate leaves criteria and hidden-order-tag queries on an element/row path.** Below this branch, batches are materialized with `BuildElementsFromBatches`, criteria are evaluated one `Element` at a time, and hidden tags are stripped per element. Please implement filtering/projection on `RecordBatch` columns and materialize protobuf only at the external API boundary.

banyand/query/processor.go
```diff
@@ -452,107 +425,65 @@ func executeMeasurePlan(
...
-// not vec-capable, or any group's dispatch declines to handle the
-// request, the whole multi-measure request falls through to the row
-// path — mixing vec for some groups with row for others would produce a
-// merged result the row path cannot validate.
+// handled per-group: each group is dispatched individually and the
+// per-group iterators are merged through the shared cross-group ordering
+// stack in logical_measure.MergeGroupMIterators (G9f.1, reusing
+// sortableDataPoints + sortedMIterator so the cross-group order and
+// version dedup are reproduced by construction, not by parallel
+// reimplementation).
```
hanahmily, line 458: **[P1] Multi-group Measure still reconstructs the legacy row merger.** `MergeGroupMIterators` wraps the per-group outputs in `sortableDataPoints -> sortedMIterator`, operating on individual `InternalDataPoint` values; raw-frame egress later drains `Next/Current` and serializes those rows back into a frame. Please replace this with batch-native cross-group ordering/deduplication and remove the retained row stack.

banyand/query/processor.go
```diff
@@ -452,107 +425,65 @@ func executeMeasurePlan(
...
-	// Dispatch would reject the schema-evolution case where one group
-	// added a tag/field the others lack (multi_group_new_tag_field
-	// integration fixture). We pre-validate here against the union once,
-	// then skip the per-group validation inside Dispatch.
+	// resolves in ANY group's schema (mirrors the mergeSchema(ss) union).
+	// Running the per-group validation inside Dispatch would reject the
+	// schema-evolution case where one group added a tag/field the others
+	// lack (multi_group_new_tag_field integration fixture). We pre-validate
+	// here against the union once, then skip the per-group validation
+	// inside Dispatch.
```
hanahmily, line 493: **[P1] Union validation does not make schema-evolved multi-group aggregates executable.** The distributed path intersects projections per group, producing differently shaped partial frames, but `ReduceRawFrames/ReducePartialBatches` takes the first frame schema as canonical and rejects later structural mismatches. A projected tag present only in one group can therefore fail with `partial N schema mismatch`. Please normalize/null-fill to the merged schema before reduction and add an old-schema/new-schema GroupBy+SUM test.

banyand/query/processor.go
```diff
@@ -452,107 +425,65 @@ func executeMeasurePlan(
...
 // data-node side handles distributed Top-over-Agg too — the vec plan emits
 // Scan → GroupByAgg(Map) → Top → Limit per node, BatchTop sorts on the
-// partial value column, and the row-path liaison's distributedPlan dedupes
-// + applies the global Top across the per-node partial top-Ns (mirrors the
-// row path's two-pass distributed Top approach; pushDownAgg is set whenever
-// Agg != nil, see measure_analyzer.go:179, and Top is applied AFTER
-// aggregation at :205-207).
-func tryVecDispatch(
+// partial value column, and the liaison's distributed plan dedupes and
+// applies the global Top across the per-node partial top-Ns.
```
hanahmily, line 466: **[P1] Preserve GroupBy+Top operator semantics.** The new distributed reducer in `pkg/query/vectorized/measure/plan/distributed.go:725-760` (and multi-group `:818-835`) applies GroupBy-first and then Top, while the removed planner applied Top before GroupBy. When the ranking field varies within a group, this can select a different winning group. Please restore the prior ordering or explicitly specify the behavior change, and add single-/multi-group fixtures where the Top field varies within a group.

hanahmily (changes requested):
The legacy dispatch branches are removed, but several production paths still execute row-by-row or have correctness gaps. Please address the inline blockers before merging. In particular, the stream distributed coordinator still decodes native frames back into Elements and performs comparableElement merge/dedup in pkg/query/logical/stream/stream_plan_distributed.go:164-229; that must also become batch-native for this PR to claim end-to-end row-path removal.

hanahmily:
I believe the practical way is to first delete all row-based implementations in the query package(pkg/query), then fix all the compile failures. This will ensure that row-based implementations are totally removed.

SammyUrfen:
Two things the PR body should have said:

- Removing the flags removes the rollback lever, so the multi-segment under-fill is no longer avoidable by config. `upgrade.md` has the mechanism, the body doesn't.
- A vec-declined stream plan is now a hard error instead of falling through to row. One shape still reaches it: an order-by on a tag that no longer resolves. I'll fix the error message and document it.

Two corrections:

- `query_batch.go:92` is gated on `topNQueryOptions`, set only when the measure is `_top_n_result` (`query.go:299`). `req.Top` goes `analyzer.go:145` -> `plan/top.go` -> `NewBatchTop` and never reaches `qr.merge`. I'll add the multi-block Top test. `qr.merge` stays a row call site either way, since `Pull()` is held by `topn_plan_localscan.go:201`.
- `distributed.go:731`: `measure_analyzer.go:130` wraps GroupBy and `:146` wraps Top above it, so row was also GroupBy-then-Top. The divergence is that row GroupBy returned the whole group while `NewBatchGroupByFirst` keeps one. `query.proto:108-136` specifies no order between them.

The other four are real. Five of the six sites are byte-identical to `v0.11.0` — this PR removes the row fallback under them rather than changing them. The exception is the criteria/hidden-order-tag gate, where `HidesOrderTag` is mine.

Three ways to go, your call:

**A.** Land this narrowed to the data-node engines, four seams as issues, plus a shrink-only ratchet in `make lint` pinning the four `.Pull()` sites — same shape as `lint-rawgo`. ~250 lines.

**B.** All of `pkg/query` in this PR. 12-18k lines across four subsystems. `MIterator` still survives, since traced queries return proto.

**C.** Columnar filtering/projection and the batch-native cross-group merge here, the schema-evolved reduce and the distributed coordinator as issues. ~1,200 lines, needs #1331 first.

I'd go with A. Either way #1331 adds the columnar pre-merge tag filter for stream and should land first.

hanahmily:
Option A sounds good to me: keep this PR scoped to the data-node engines, track the remaining work in follow-up issues, and add the shrink-only lint guard.

Please also replace both occurrences of `Resolves apache/skywalking#13998` in the PR description (the opening line and the checklist) with a non-closing reference such as `Related to apache/skywalking#13998`. With the narrower scope, this PR only completes part of the original issue, so merging it should not automatically close that issue. Please update the description to make the reduced scope and remaining work explicit.

SammyUrfen:
Updated the description: both references are `Related to` now, and it says what this covers and what it does not. The lint guard is in `scripts/lint/rowcallsites/` and runs as part of `make lint`. It counts sites per file instead of by line number so unrelated edits do not break the build, and the four that remain each carry a note on why. The multi-block Top test is `banyand/measure/query_batch_top_test.go`; if the row fallback is forced the test fails, so it checks the path and not only the result. A stream query the vec engine cannot plan now reports the reason, and I fixed a case where it reported the wrong one for multi-group plans.

I also corrected two things in the description. The binary size numbers were wrong, so I rebuilt both branches from a clean cache and used the measured values. The slow query line references were wrong too: both of them are measure, not measure and trace. The four seams are unchanged.

hanahmily:
We don't need the lint guard in this PR; human review is sufficient to prevent reintroducing the row path. Please remove the guard and its mentions from the PR description. Also, please resolve the conflicts with main after #1331 and rerun CI.

SammyUrfen:
The lint guard and its mentions in the PR description have been fully removed. The conflicts resulting from the #1331 merge have also been reconciled, and the CI is green again locally.

hanahmily:
@SammyUrfen could you pls fix these conflicts?

SammyUrfen:
Conflicts resolved against main, in two commits so you can take them separately.

The merge commit is mechanical, with one exception. `#1334` bounded the row scan allocations and this PR deletes both files it touched. The vectorized path already calls `query.Charge`, so the block budget stays. `banyand/stream/query_capacity.go` goes too, because `unused` fails `make lint` otherwise.

The second commit drops the `engine` matrix axis from the e2e workflow. With one engine left both arms run the same code.

Also stripped `--*-vectorized-enabled=false` from the new lifecycle and rover files.

hanahmily (approved):

PR #1331, opened 2026-09-06
perf(stream): bound the filtered index-order vectorized Top-N merge

### LINKED ISSUE #14056: [Performance] Bound filtered Stream vec Top-N (columnar tag-filter pushdown)
## Summary

Follow-up to [#14050](https://github.com/apache/skywalking/issues/14050) **R3**.

[PR #1322](https://github.com/apache/skywalking-banyandb/pull/1322) bounded `SortedMerge` for **criteria-less** Stream vec queries (R1/R2) and documented why **filtered** index-order queries still set `mergeCap = 0` (`deferLimitToEgress`). That documentation half of R3 is done. This issue tracks the remaining work: make filtered Stream vec queries memory-bounded **without** starving the tag filter or breaking row-path parity.

## Why a naive merge cap is wrong

For `*limit → *tagFilterPlan → *localIndexScan` (index-order), the vec path today is:

```
scan → SortedMerge (uncapped) → Distinct → materialize Elements → applyStreamTagFilter → LIMIT
```

The filter runs **after** merge. Cap at `limit+offset` before the filter can under-fill: the top-N pre-filter rows may contain few or zero survivors, while the row path keeps pulling until it has enough matches (`tagFilterPlan.Execute` nested under `*limit.Execute`).

Bounding this path requires the filter (or an equivalent predicate) to run **before or inside** the merge so the merge can safely apply `mergeCap = maxElementSize` the way criteria-less queries do after #1322.

## Boundary

Deliver a Stream vec path for filtered queries where:

1. Tag-filter predicates are evaluated on columnar batches (or another pre-merge stage), and
2. `SortedMerge` / `BuildStreamMergePipeline` uses a non-zero `mergeCap` for the filtered index-order case without changing the element set or order relative to the row path.

Callers that must keep parity:

- `pkg/query/logical/stream.(*localIndexScan).ExecuteVectorized` (`deferLimitToEgress` / `mergeCap`)
- `banyand/query/processor.go` (`VecTagFilter` / `applyStreamTagFilter` egress)
- Standalone and distributed stream vec integration tests

## Requirements

R1. For a filtered **index-order** Stream query with `limit = L` and `offset = O`, vec results must match the row path (same elements and order for a fixed fixture).

R2. That path must not leave `SortedMerge` uncapped solely because a tag filter exists: merge auxiliary state must be bounded by `L + O` (distinct-`ElementID` aware), subject to the same residual batch-pinning caveats as #1322, or a stricter owned-buffer design.

R3. Time-order filtered queries that already set `deferLimitToEgress = false` (see `scanResumesAcrossPulls`) must remain row-equivalent; do not regress the intentional under-fill parity for single-segment ts-order scans.

R4. Existing stream vec parity / integration tests stay green; add focused RED tests that fail today because filtered index-order still uses `mergeCap = 0` (e.g. assert mid-consume merge buffer growth or planner `mergeCap` for a filter plan).

## Acceptance criteria

- [ ] At least one production caller (`ExecuteVectorized` or processor egress) uses the pre-merge / columnar filter path for filtered index-order queries.
- [ ] `go test` / integration: `pkg/query/vectorized/stream/...`, `pkg/query/logical/stream/...`, `test/integration/standalone/query/vectorized_stream_test.go`, `test/integration/distributed/query/vectorized_stream_test.go`.
- [ ] A new test proves bounded merge state (or non-zero `mergeCap`) for a filtered index-order plan; it fails on current `main` after #1322.
- [ ] No on-disk format or wire protocol change unless separately labeled `bydb file compatible change`.

## Scope

**Packages:** `pkg/query/vectorized/stream/`, `pkg/query/logical/stream/` (`stream_plan_indexscan_local_vectorized.go`, `stream_plan_tag_filter.go`), `banyand/query/processor.go`.

**Out of scope:** criteria-less merge prune (done in #1322); Measure/Trace vec; removing the row path (#13998); Canopy UI.

## Dependencies

- Blocked by: [#14050](https://github.com/apache/skywalking/issues/14050) R1/R2 via [PR #1322](https://github.com/apache/skywalking-banyandb/pull/1322) (**merged**).
- Parent: #14050.

## Size audit

**Classification:** tracking parent (R3 workstream).

**Why not one leaf:** needs a new filter seam (columnar / pre-merge), planner `mergeCap` cutover, possible egress simplification, and parity coverage — multiple activation points.

**Suggested first leaf (file when starting work):** columnar (or batch-level) tag-filter fusible + enable `mergeCap = maxElementSize` for filtered index-order `ExecuteVectorized`, with one parity fixture and one bound/RED test. Decompose further only if that leaf still exceeds one mergeable PR.

**RED test sketch:** filtered index-order plan today has `mergeCap == 0` / unbounded merge buffer under load; after the leaf, `mergeCap > 0` and retained merge state stays O(cap).

SammyUrfen:
A filtered index-order stream query left its vectorized merge unbounded. Follow-up to
#1322, which bounded the criteria-less case and left this one open.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#14056.

## What changed

The criteria tag filter now runs before `SortedMerge`, as a fusible operator that narrows
`batch.Selection`. The merge orders only rows that already match, so its `maxElementSize`
cap holds the top-N of the filtered set. Gated on index order — filtered timestamp-order
queries still filter at the egress, unchanged. The scan-level cap and the client
`offset:offset+limit` slice are unchanged.

The operator is one new file, `pkg/query/vectorized/stream/tag_filter.go`, shaped like
`Distinct` and `Limit` in the same package. It reuses `logical.TagFilter`, so the pre-merge
and egress filters evaluate a criteria expression identically.

The egress filter stays and `hasFilter` stays true. That bool also gates the columnar frame
path off, and disabling it would drop the only hidden-tag strip on the vec path and leak
filter-only tags to the client.

The merge-cap comment this deletes argued a cap is unsound because a top-(limit+offset)
window taken before the filter can hold arbitrarily few survivors — true only while the
filter sits at the egress.

New import edge: `pkg/query/vectorized/stream` now imports `pkg/query/logical`. It is
cycle-free, and `check-import-boundaries` does not cover it.

## Duplicate `ElementID` semantics

For a filtered index-order query the criteria is evaluated first, and among the matching
rows sharing an `ElementID` the element is represented by the first in the requested sort
order. One matching version makes the element eligible even when another fails. Rows of one
element often carry the same ordered-tag value and so tie on the sort key; among equal keys
the first row the scan yields wins, in both directions.

Filtered timestamp-order queries take no pushdown and keep the older rule, as does
`--stream-vectorized-enabled=false`.

This differs from the pre-change vectorized pipeline, which chose an element's first row in
sort order before the criteria ran and so dropped an element whose first row failed even
when a later row matched. Preserving that rule is not compatible with the bound: it needs a
running set of every `ElementID` seen before the criteria runs, trading an unbounded merge
buffer for an unbounded dedup set. This amends R1 of #14056 — "same elements and order" now
scopes to elements with distinct `ElementID`s.

Duplicates need no client misuse: the write path appends with no upsert check
(`write_standalone.go:181,185`), a part merge never dedups by element id
(`merger.go:425-463`), and `ElementID` is a 64-bit xxhash of `group|name|<client id>`
(`pkg/convert/hash.go:28-30`), so distinct client ids collide at about 2^32 elements in one
stream.

Pinned by `TestTagFilterDuplicateWinnerSemantics` — exact ordered ids plus the selected
version's values and sort key, both directions, single and split batches, with and without
an offset, cap on and off — by `TestTagFilterDuplicateTieOnSortKey` for the equal-key case,
and by `TestPushdown_FilteredIndexOrder_BoundsMerge`, which asserts exact ids out of
`ExecuteVectorized` over a corpus whose lowest-sorting rows fail, so removing the pushdown
fails it.

## Verification

`make lint` 0 findings, `make check` clean, `banyand/stream` 17 of 17 Ginkgo specs,
standalone integration 16 suites / 1840 specs, distributed integration 14 suites / 885
specs. No golden fixture was regenerated.

Benchmark: 100k rows, 1024-row batches, cap 20, fixed seed, medians of `-count=5`.
"Before" is merge cap 0 with no pre-merge filter.

| criteria keeps | time | bytes |
|---|---|---|
| 1% | 91.86 ms → 4.13 ms | 13.09 MB → 1.88 MB |
| 95% | 93.22 ms → 35.91 ms | 13.09 MB → 1.98 MB |

At a 95% pass rate the filter costs +1.5-3.7 ms per 100k rows, about one allocation per
scanned row, from the `parseExpr` call `logical.tagExpr` makes on every `Match`. The win is
a lower bound: the "before" arm passes limit 20, while the real pre-change shape passed
`math.MaxUint32`. It does not bound scan I/O, batch pinning (`SortedMerge` still retains
batches), or the liaison-side distributed merge.

Server cases: `sort_duration_no_index_limit` remains the pushdown guard, because its
criteria tag carries no index rule. The new `sort_duration_filter_range_limit_offset` pins
the ordered sequence under a non-zero offset — where the cap is `limit+offset` while the
egress slice is `offset:offset+limit` — but `duration` is indexed, so it is not itself a
pushdown guard. No golden goes red: the pre-change code returned the same ordered set for
every committed fixture and only failed to bound the merge, and no fixture holds a
duplicate `ElementID` whose versions disagree.

## Also found, not fixed here

1. **Null ordered tags sort first ascending.** Pre-existing. The vec path marshals a null
   ordered-tag cell to empty bytes (`query_vectorized.go:312`), so an element missing the
   sort field sorts ahead of every real key, and an ASC query with a limit below the number
   of null-ordered elements returns only nulls. Whether that or "nulls last in both
   directions" is intended is not settled anywhere. It reproduces today on the
   criteria-less capped path.
2. **The `!hasFilter` frame-egress gate.** Lifting it would let criteria queries use the
   columnar wire, but the frame path needs a hidden-tag strip of its own first.
3. **Early scan termination.** The scan still emits every matching row. Bounding that needs
   the filter inside `banyand/stream/query_vectorized.go` plus a sort-key watermark from
   the merge.

Happy to file those as separate issues or fold any into this PR — whichever you prefer.


pkg/query/logical/stream/stream_plan_indexscan_local_vectorized.go
```diff
@@ -310,32 +316,24 @@ func (i *localIndexScan) ExecuteVectorized(ctx context.Context) ([]*vectorized.R
...
-	// pushdown (out of scope here), not a bigger cap.
+	// A criteria query runs its tag filter as a PRE-MERGE fusible, so the merge sees
+	// only surviving rows and the cap keeps the top-N of the FILTERED set — not a
+	// top-N taken before the filter, which the filter's unknown selectivity could
+	// leave empty. The egress still applies the same filter; on rows that already
+	// passed here that re-check is a no-op, and it remains the hidden-tag strip.
+	var preMerge []vectorized.FusibleOperator
+	if i.preMergeFilter != nil {
+		preMerge = append(preMerge,
+			vstream.NewTagFilter(schema, i.projectionTags, i.preMergeFilter, i.filterRegistry))
```
hanahmily, line 330: [P1] Preserve duplicate-winner semantics before pushing down the predicate. This filter runs before SortedMerge's distinct-ID pruning and Distinct. If the first ordered version of ID A fails the predicate but a later version passes, the first version is removed and A is now emitted. A production-operator reproduction with `(A, closed), (A, open), (B, open)`, `state=open`, limit 2 returns `[B]` on the previous pipeline and `[A,B]` here. Rechecking at egress cannot undo this. Existing bounds/selectivity tests do not exercise conflicting predicate values for the same ID; please add exact-result duplicate parity coverage and resolve the semantic regression.

hanahmily (changes requested):
Requesting changes for one correctness regression: pushing the predicate ahead of duplicate selection changes result membership for duplicate ElementIDs with conflicting filter values. This contradicts apache/skywalking#14056 R1 (same elements and order), even though the PR description acknowledges the behavior change.

I reproduced this against this head using the production vectorized operators in an isolated checkout with this PR's generated CI artifacts. For ascending rows `(A, closed), (A, open), (B, open)`, criterion `state = open`, and limit 2:
- Previous merge -> distinct -> filter -> limit: `[B]`.
- New filter -> capped merge -> distinct -> limit: `[A, B]`.
The equality regression fails (`before=[2] after=[1 2]`). This is not dependent on a hash collision: the write path accepts repeated client element IDs. The second egress filter cannot recover the discarded first version.

### Why the current tests miss this
1. `TestPushdown_FilteredIndexOrder_BoundsMerge` uses distinct IDs with every row matching, then checks only that the output count is positive and <= the cap. It does not assert duplicate-winner preservation or exact result membership.
2. `TestTagFilterBoundsRetainedMergeState` uses unique IDs and manually constructs Selection before feeding the merge. It proves a memory bound, not equivalence of filter-before-dedup and dedup-before-filter.
3. `TestTagFilter_Process` checks predicate/selection behavior in isolation, not its ordering relative to Distinct. The focused filtered Top-N fixture checks ordering/selectivity, but does not contain repeated IDs whose versions disagree on the predicate.
Consequently, these tests can all pass while query membership changes. I also ran the existing `pkg/query/vectorized/stream` and `pkg/query/logical/stream` package suites successfully before running the failing targeted reproduction.

Please preserve duplicate-winner semantics before evaluating the predicate, or obtain an explicit change to the parity/duplicate semantics contract rather than closing this issue as parity-preserving. Add a regression with conflicting versions of the same ID (including across batches) and assert exact ordered IDs and offset/limit results against the intended oracle, not just count or bounded state.

SammyUrfen:
Thanks, that reproduces — and you're right that the egress re-check can't recover it.

I've added a regression for it: `TestTagFilterDuplicateWinnerSemantics` in `pkg/query/vectorized/stream`. Six rows, two elements carrying conflicting versions, asserting exact ordered ids in a single batch and with each duplicate pair straddling a batch boundary, for limit and for offset+limit. It also runs the pushdown with the merge cap disabled and gets the same ids back, so the cap isn't what changes membership, and no cap can undo it.

One thing I hit while writing it that bears on the rule you pick: the row path isn't the old vec pipeline either. For a filtered index-order query it admits an element if any of its index documents matches (`inverted.go:232` inserts rather than updates), takes the sort position from the first doc in sort order (`query_by_idx.go:228-240`), then re-tests the criteria on the first row in storage order among the rows actually scanned (`mergeByTagValue`, `query_by_idx.go:301-311`). For your example that comes out as `[B]` or `[A,B]` depending on which of A's rows was stored first, and on compaction state. The old vec pipeline keys on the sort-first row, so the two already disagree on a corpus like this.

Preserving the sort-first rule under the pushdown also looks incompatible with the bound. The representative is picked before the predicate runs, and two rows of one element can sit arbitrarily far apart in sort order, so the state needed grows with every distinct id in the scan whenever the criteria is selective. I tried four designs and broke all four — happy to write that up if it's useful.

Could you pin the rule for me: first-in-storage, first-in-sort, or any-row-matches? I'll make the test assert whichever you choose, and if it isn't any-row-matches I'll pull the pushdown out of this PR and open a separate design issue for it.

hanahmily:
Thanks for the investigation and the duplicate coverage. Filter-first is the intended behavior: evaluate the criteria first, then among matching rows with the same ElementID, keep the first in the requested sort order. A matching version should make the element eligible even if another version fails the criteria.

Please keep the pushdown. Do not use the row path as the oracle for this decision; it will be removed. The behavior difference from the old vectorized pipeline is accepted under this rule.

Please finalize the regression tests against this contract, retaining the cross-batch and offset/limit coverage and asserting the selected version’s values/sort key as well as the ordered IDs. Update the PR description, release note, and relevant comments to mark the duplicate semantics as settled rather than an open question, and remove claims of unconditional row-path parity.

The existing benchmark is sufficient for this change. Please also rerun the failed Zipkin Vec check; its log shows a swctl archive installation failure rather than a query assertion failure.

SammyUrfen:
Pushed, and the description is updated — the duplicate rule reads as settled there, in the release note and in the comments, and the row-path parity claims are gone.

`TestTagFilterDuplicateWinnerSemantics` asserts your rule with the selected version's tag values and sort key alongside the ordered ids, ascending and descending, across a batch boundary, and with an offset. `TestTagFilterDuplicateTieOnSortKey` covers the equal-sort-key case, where the tie falls to arrival order in both directions.

I also tightened `TestPushdown_FilteredIndexOrder_BoundsMerge`, item 1 from your first review. It was count-only over a corpus where every row matched, so the pushdown could be deleted without failing anything. It now asserts exact ordered ids.

Zipkin Vec is green.

hanahmily (approved):
Re-reviewed the update against the agreed filter-first contract. The duplicate-winner tests now assert selected values and sort keys, ASC/DESC, cross-batch duplicates, offset/limit, and equal-key ties. Both affected package suites pass locally, including race checks; the focused storage-level filtered Top-N test passes, and disconnecting production pushdown makes the strengthened regression fail as expected. No remaining correctness blocker found.

Nonblocking documentation nit: please qualify the already-filtered input comment in pkg/query/vectorized/stream/distinct.go as filtered index-order queries; timestamp-order filtering still happens at egress.

PR #1335, opened 2026-09-08
feat(index): page repair tuples natively and cut Property repair over to them

### LINKED ISSUE #14012: [NIDX-01E] Cut over Property repair search-after
# [NIDX-01E] Cut over Property repair search-after

Parent workstream: #14002
Blocked by: #14011

## Summary

Add the bounded doc-value tuple page required by Property repair and move repair tree construction off Bluge. This final NIDX-01 leaf implements only ascending `(group, name, entityID, timestamp)` ordering, strict search-after, stored SHA projection, and pinned-generation behavior.

## Blocked by

NIDX-01D (#14011) must be merged before this work starts. Before applying `Backlog`, update inherited visitor identifier names to match main; do not widen the behavior.

## Pre-agreed test seam

The product seam is one complete Property repair build: provide a committed shard snapshot to the existing repair build operation and observe its persisted repair tree/state output. Tests may exercise the bounded native tuple-page operation through the BanyanDB-owned read-only visitor merged by earlier leaves, but must not target doc-value chunk decoders, heaps, or ICE types.

No generic collector, arbitrary sort expression, offset pagination, or relevance order is part of the seam.

## Independent fixture

The checked-in legacy Property shard contains these visible repair rows in declared ascending order:

1. `(g-a, n-a, e-1, 10, sha-a)`
2. `(g-a, n-a, e-1, 20, sha-b)`
3. `(g-a, n-b, e-2, 5, sha-c)`
4. `(g-b, n-a, e-3, 7, sha-d)`

With page size `2`, page one is rows `1–2`; the strict cursor is `(g-a,n-a,e-1,20)`; page two is rows `3–4`. The completed repair tree maps `g-a/n-a/e-1 → sha-b`, `g-a/n-b/e-2 → sha-c`, and `g-b/n-a/e-3 → sha-d`.

## Requirements

R1. The native page orders encoded doc values by all four ascending components and resumes strictly after the complete prior tuple on the same pinned generation. It returns no duplicate or skipped row.

R2. Each page holds at most the configured page size plus bounded comparison state; it never performs a growing offset scan or materializes the shard.

R3. The existing repair build uses native pages and produces exactly the declared three entity/SHA leaves and its normal generation state.

R4. A generation published after page one does not enter page two. Missing sort values, malformed doc-value chunks, invalid cursors, cancellation, and limit overflow return typed bounded errors and publish no partial repair state. CRC32 remains ignored.

- Boundary RED: request the two declared pages through the native tuple-page seam; earlier leaves do not support doc values or search-after.
- E2E RED: run a complete Property repair build at page size `2`, publish another legacy generation between pages, and verify the tree contains only the pinned fixture's three declared leaves and SHA values.

## Acceptance criteria

- `go test ./pkg/index/inverted/... ./banyand/property/db -run 'Test(NativeRepairTuplePage|E2EPropertyRepairNative)' -count=1` passes, including under `-race`.
- A call-graph/dependency assertion proves Property repair no longer opens a Bluge reader or constructs a Bluge `TopNSearch`.
- Pagination, concurrent publication, corrupt doc values, invalid cursor, cancellation, and bounded-memory cases pass.
- `make test-ci PKG='./pkg/index/inverted/... ./banyand/property/db'`, `make build`, and `make pre-push` pass.

## Scope

Packages: `pkg/index/inverted`, its private doc-value reader, and `banyand/property/db` repair.

Out of scope: arbitrary sort, descending repair order, range/prefix/wildcard/MATCH queries, online Property writer/query cutover, snapshots written by native code, merge, expiry, and GC.

## Compatibility and rollback

This slice emits only the existing repair state/tree artifacts, not index bytes. Roll back repair reads to the retained Bluge path; the source ICE/snapshot directory remains unchanged.

## Lexical non-regression

## Design

hanahmily:
Property repair builds a Merkle tree over every visible property in a shard,
and its peers compare that tree over gossip to decide what to re-send. To build
it, the repair scan needs the shard's rows in one specific order — group, then
name, then entity, then revision — with the stored SHA of each row, paged so a
large shard never has to be held in memory at once.

Until now that ordering came from a third-party ranked searcher: the scan opened
a reader over the snapshot directory, built a top-N search sorted on four
fields, and re-issued it with a search-after cursor for each page. This change
replaces it with a bounded tuple page served by BanyanDB's own read-only index
reader, completing the read-only slice of the native inverted index work.

## What the reader now offers

The reader gains a pinned view of one committed generation. The generation is
chosen when the view opens and fixed for its lifetime, so a caller can page
through an order across many calls and see one consistent set of documents —
something the existing single-shot count, walk and select operations could not
promise. On top of that view sits one operation: a page of rows ordered ascending
by exactly four components, resuming strictly after a complete prior tuple, each
row carrying one projected stored value.

That is deliberately not a query language. The sort arity is part of the request
type rather than a runtime check, there is no filter, no descending direction,
no offset, no relevance order, and the encoded doc values travel back to the
caller unchanged so no third-party codec type escapes the reader's internal
boundary. The doc-value decoding that makes it work stays private and brings no
new dependency: the historical chunks use the same block codec the reader
already reads stored fields with.

```mermaid
flowchart TD
    subgraph repair["banyand/property/db — repair build"]
        B[buildStatus] --> O["openGeneration(shardPath)"]
        O --> L[page loop]
        L -->|"cursor = last row's sort values"| L
        L --> C[tree composer]
        C --> T[(repair tree)]
        B --> S[(generation state)]
    end
    subgraph reader["pkg/index/inverted — read-only reader"]
        O -.-> G[pinned generation]
        G --> ID[SnapshotID]
        L -.-> P[RepairTuplePage]
        P --> DV[private doc-value reader]
        DV --> M[deletion masks applied]
        M --> R["ordered rows + projected SHA"]
    end
    G -->|"pins one generation"| F[(ICE v3 / snapshot v3 shard)]
    W[Property writer] -->|"publishes a newer generation"| F
    F -.->|"invisible to the pinned view"| G
```

## Why pinning matters here

A repair build runs for as long as a shard is large, and the Property writer does
not stop while it does. Under the old scan each page re-entered the directory,
so a build could see part of one generation and part of the next and hash a tree
that never existed on disk — and its peers would then repair against it. With a
pinned view, a generation published between two pages simply is not there, and
the generation marker the build records afterwards names the generation it
actually read rather than whatever the directory has moved on to. That closes a
window in which the next build could skip a generation this one never saw.

## Behaviour worth calling out

Rows are ordered on the encoded doc values as raw bytes, which is the ordering
the on-disk encoding was designed to preserve, and a row that records no value
for a component sorts last among its siblings rather than failing the build.
That case is real rather than defensive: a property written without a
modification revision records no revision timestamp at all, and such shards read
correctly today.

Failures are bounded and typed. A page size or cursor outside the reader's
limits is rejected before any byte is read and stays distinguishable from
damaged bytes; a damaged doc-value section and a shard that was simply never
flushed report different sentinels; cancellation reports cancellation. Every
failure yields no rows, and a build that fails part way through publishes
neither a tree nor a generation marker, so the next build redoes the work
instead of trusting a tree that covers half a shard.

Nothing is written. The slice emits only the repair state and tree artifacts
that already existed; the shard's bytes, modes and modification times are
untouched, no exclusive directory lock is taken, and the reserved checksum
fields are carried without being calculated or validated. Rolling back is
therefore redeploying the previous release on unchanged files.

## Compatibility

Existing shards are the input, and one is checked in as a fixture: a directory
produced by the compatibility writer, with a manifest recording its content
hashes and every row it declares, including a deleted row placed so that a page
built without the deletion masks is caught in the first page rather than at the
end. Because the tree's slot and root hashes derive from the order in which
leaves are appended, a node running this change and a node running the previous
release must build byte-identical trees for identical shard content; the
existing Property repair and gossip suites pass unmodified, which is what
demonstrates that.

Fixes apache/skywalking#14012


hanahmily:
## Real-data compatibility and robustness evidence (264a7120)

Executed read-only from clean `fix-pr1335-review` at `264a7120bf25af687027b8838e40780a74ee0519`. Original fixtures and source data were not modified; SHA-256 inventories before/after remained identical. No raw values or private fixture contents are included here.

### Differential results

- Small retained fixtures: native and legacy canonical stored-field walks are exactly equal: Property `sw_property` 633 (SHA-256 `bb0c37aefa3622ab2ddcead19c11970aeb7400d398431d4151e80f5d50d92a06`), deletion task 48 (`a06b9d7a5033a3834ff2c66bdc754b01cb54ac376f672fa61cec40c142be4fb4`), schema 8774 (`9c71fb27b924d11f1694f23a85127335ed235c33ada1bf46da63adec4ec16327`). Corrected repair TopNSearch cursor differential also passes: 633 and 8774 rows, digests `4a5a9ee7664c4c008aea3c3e284bd1af1bae91f279253c379338c1bc3dcdc01` and `62c77548ebd1a5377ccd5c2ffb0d6233607fd5b048e6684afbae86a88e1f5c9c`.
- Broader live read-only snapshot representatives: Stream recordsLog 64739, browserErrorLog 105825, Measure metricsHour 9587, Trace 10; native count, legacy count, and MatchAll all agree. Canonical stored-field digests also agree (respectively `08902604024d1910c29d94f14b8035fa05c60143b64f530221259b5f03f63a4b`, `7aface19af78b778c410da31e22a2e6e52998e7986e58cc289b1812021968cf2`, `f8717591693626a31a3f65be5b59b466597284853c8f78e49c743ffa42ebff58`, `d5e4e2b23ed7032325e62c62cf442187987fce5aab327d881e59972782bed2a8`). A Stream sidx representative (132) also passes (`c9bd0c88b6da39926191d80237b4a60aae129bc58191ab9b5255b2553260fae0`).
- Exact and Boolean-OR selection against private dictionary terms: 37 cases across those representatives, all native/legacy counts and canonical digests equal.

### Robustness/publication

Commands:
```
go test ./pkg/index/inverted -run 'TestE2EReadOnlyCountDuringPublication|TestReadOnlyGeneration' -count=1
go test -race ./pkg/index/inverted -run TestE2EReadOnlyCountDuringPublication -count=1
```
Both pass. Cancellation, invalid-page, corruption, pinned-generation, and bounded-page tests pass; finite memory sampling is reported without a global bound claim. Real-data disposable-copy publication: pinned old view stayed 633 while retained writer appended one record; fresh view became 634. Reserved trailing CRC slots rewritten to arbitrary bytes remained readable (CRC is ignored by design); corrupt doc-value locations return typed `ErrCorruptIndex`.

### Limits

This revision contains no native ICE writer: mixed-data evidence is legacy-writer -> native-reader interoperability only (small fixture append: 633 -> 635, repair 635 rows, exact digest equality). Native-writer -> legacy-reader, native append/update/delete/merge, and native snapshot publication remain untested and are outside this PR. Compatibility caveat: old revision-0 records lacking timestamp fail repair-sort validation and are rejected; this is intentionally not hidden. Synthetic publication/race evidence is not a native-writer proof.

Detailed local reports (not reviewer links): `/tmp/showcase-native-small-audit.md`, `/tmp/compat-read-20260910/README.md`, `/tmp/read-robustness-20260910.md`, `/tmp/showcase-native-mixed-legacy-20260910T2862382.md`.

mrproliu (approved):

PR #1355, opened 2026-09-14
docs: align cluster and node discovery ports and remove etcd leftovers

### LINKED ISSUE #14081: [Docs] Align BanyanDB cluster/discovery docs, drop etcd leftovers, and publish docs from a docs branch
## Background

Follow-up from https://github.com/apache/skywalking/discussions/14077#discussioncomment-18431254: community hit contradictory port examples between the cluster install guide and the file-based discovery guide, and asked which gRPC port liaison should use.

Default `--grpc-port` for both `banyand data` and `banyand liaison` is **17912**. `18912` is only a common override when data and liaison share the same host/IP. Etcd-based discovery has been removed; docs still mention it in places.

### 1. Fix cluster / file-discovery port examples

- Align [installation/cluster](https://skywalking.apache.org/docs/skywalking-banyandb/latest/installation/cluster/) and [operation/node-discovery](https://skywalking.apache.org/docs/skywalking-banyandb/latest/operation/node-discovery/) so `nodes.yaml` / liaison examples are consistent.
- State clearly: default gRPC port is `17912` for data and liaison; when co-located on one IP, use different ports (e.g. data `17912`, liaison `18912`) and list the same ports in `nodes.yaml`.

### 2. Remove redundant etcd references from documentation

- Purge leftover etcd discovery / configuration / upgrade wording from BanyanDB docs (e.g. `docs/operation/node-discovery.md`, `docs/operation/upgrade.md`, and other hits under `docs/`).
- Keep only current discovery modes: `none`, `dns`, `file`.

### 3. Docs branch for continuous doc updates without a release

- Create and maintain a dedicated docs branch in https://github.com/apache/skywalking-banyandb (e.g. `docs` / `asf-site` workflow as appropriate) so documentation can be updated and published without waiting for an official BanyanDB release.
- Document how doc PRs should target that branch and how they reach skywalking.apache.org.

## Acceptance

- [ ] Port guidance is consistent across cluster + node-discovery docs
- [ ] No obsolete etcd discovery instructions remain in the published docs
- [ ] Docs branch exists and is the agreed path for mid-release doc fixes

HesandaLiyanage:
### Align BanyanDB cluster and node discovery port documentation and remove etcd leftovers

- Align port guidance across `docs/installation/cluster.md` and `docs/operation/node-discovery.md`:
  - Clarify that the default `--grpc-port` for both `banyand data` and `banyand liaison` is `17912`.
  - When co-located on the same host/IP, different ports must be assigned (e.g., data `17912`, liaison `18912` via `--grpc-port=18912`) and listed in `nodes.yaml`.
  - Provide matching `nodes.yaml` co-location examples and liaison startup commands.
- Remove obsolete etcd-based discovery references and sections from BanyanDB documentation (`docs/operation/node-discovery.md`, `docs/operation/upgrade.md`, `docs/operation/observability/providers.md`, `docs/concept/storage-and-format.md`).
- Consolidate the `None Mode` section in `docs/operation/node-discovery.md` and remove duplicate text.
- Retain only active discovery modes: `none`, `dns`, and `file`.

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue number. Fixes apache/skywalking#14081.


docs/installation/cluster.md
```diff
@@ -105,7 +118,14 @@ Liaison nodes are stateless and use the same discovery source to locate data nod
...
 
 ```shell
+# Dedicated host (default port 17912)
+banyand liaison \
+  --node-discovery-mode=file \
+  --node-discovery-file-path=/etc/banyandb/nodes.yaml
+
+# Co-located on the same host/IP as a data node (override port, e.g. 18912)
 banyand liaison \
+  --grpc-port=18912 \
```
hanahmily: [P1] This public-port override collides with liaison's own internal gRPC server: pkg/cmdsetup/liaison.go:68 creates internalPipeline with ports 18912/18913. Consequently, --grpc-port=18912 makes two servers in the same process bind the same endpoint. Co-location also leaves data and liaison competing for HTTP 17913, observability :2121, and pprof :6060.

Please provide a complete non-conflicting configuration, either choosing another public gRPC port or explicitly moving the internal listener, and separating the other shared listeners. Update nodes.yaml and the duplicate example in docs/operation/node-discovery.md to match. Note that 18913 is also already used internally, so it is not a free HTTP override by default.

docs/operation/node-discovery.md
```diff
@@ -363,16 +270,44 @@ nodes:
...
+
+```shell
+# Dedicated host (default port 17912)
+banyand liaison \
+  --node-discovery-mode=file \
+  --node-discovery-file-path=/etc/banyandb/nodes.yaml
+
+# Co-located on the same host/IP as a data node (override port, e.g. 18912)
+banyand liaison \
+  --grpc-port=18912 \
```
hanahmily: [P2] Changing --grpc-port does not change the HTTP gateway's upstream. banyand/liaison/http/server.go:106 independently defaults --http-grpc-addr to localhost:17912, which in this co-located setup is the data node rather than liaison. Even after resolving the listener collisions, HTTP API calls would still target the wrong server.

Please explicitly set --http-grpc-addr=localhost:<chosen-liaison-public-grpc-port> in both guides' co-location commands.

docs/operation/node-discovery.md
```diff
@@ -32,120 +40,19 @@ None mode is the default discovery mode since 0.10.0. In this mode, nodes do not
...
+- Smoke tests and unit tests where a cluster is not needed and there are no external infrastructure dependencies.
 
-- `banyand data`
-- `banyand liaison`
-- `banyand standalone` (only `none` is meaningful here)
-- `lifecycle` (the lifecycle agent — see the [lifecycle documentation](lifecycle.md) for how it uses the discovery registry to pick migration targets)
+### Caveats
 
-This document provides guidance on configuring and operating all discovery modes.
+- **Not valid for clusters.** Running any of `banyand data`, `banyand liaison`, or `lifecycle` with `--node-discovery-mode=none` will leave the schema client with no server to talk to, liaison routing tables empty, and lifecycle migrations as no-ops. Use `dns` or `file` for every clustered deployment.
```
hanahmily: [P2] The cluster recommendation is correct, but the retained explanation overstates what none mode does. A data node can bootstrap its own schema endpoint directly: banyand/metadata/schema/property/client.go:221–223 seeds the connection manager from CurNode when it hosts the property schema. Also, discovery/none/none.go:39–52 caches the local node rather than providing an entirely empty registry.

Please describe none as disabling discovery of remote peers, and distinguish data-node local schema bootstrap from liaison/lifecycle discovery requirements. The blanket statement that all three commands have no schema server is inaccurate; align the overview at line 34 as well.

hanahmily (changes requested):

hanahmily:
### Review summary — Request changes

Remaining documentation cleanup:
- The source comments in api/proto/banyandb/database/v1/rpc.proto still describe schema mod_revision fields as “etcd revision.” Please correct those source comments; do not edit the generated docs/api-reference.md in this PR.
- docs/operation/node-discovery.md:225–226 and 341 still describe failed nodes as retried on each file reload. After startup, unchanged failed entries are handled by the separate backoff retry mechanism.
- The same guide calls the file-entry name required (line 268), and the cluster guide requires unique names, although the parser does not enforce that requirement. Please distinguish a recommended diagnostic label from a required field.

HesandaLiyanage:
Hi @hanahmily , thanks for the detailed feedback! I have addressed all the requested changes

*Note on CI:* All doc checks (`CheckDeadLinks`), integration, and E2E suites passed. The single failing check in `Test Bydbctl` appears to be an unrelated timing flake in `measure_test.go:208` (asynchronous `_topn_result` creation race). Could you please re-run that job when you have a moment?

HesandaLiyanage:
Hi @hanahmily , I noticed the `Vulnerability Scan` CI check is currently failing across PRs due to new advisories published against `google.golang.org/grpc@v1.83.0` (fixed in `v1.83.2`).

Would you like me to open a separate PR to bump this dependency and get the scan green, or is that already being handled?

hanahmily (approved):

PR #1363, opened 2026-09-18
docs(design): correct tag-aggregation design doc after #14089's review cycle

hanahmily:
## Summary

Implementing and reviewing time-bucket grouping (apache/skywalking#14089, #1362) surfaced three places where the merged [tag aggregation and time bucketing design](docs/design/0.12.0/tag-aggregation/README.md) (#1360) no longer matches what actually shipped:

- **§7.5** claimed liaison-merge and index-mode ordering were bucket-ascending "by construction," same as the node-local streaming case. Both actually need an explicit sort (`BatchAggregation.SortInsertionByBucket`), added in #1362 after two review rounds found the gap. Node-local ordering is still free; the doc now says which case is which, and flags that COUNT_DISTINCT's distributed reduce (§7.4, apache/skywalking#14090) shares this same reduce path and must not re-introduce the false assumption.
- **§7.2**'s "relaxes to a pipelined operator" undersold the actual change: a `BreakerOperator` cannot be pipelined by relaxation, since `breakerStage`'s contract fully drains upstream before serving output. The doc now names the real fix (`BatchTimeBucket` as a genuine `PullOperator`, spliced in via a new `PipelineBuilder.Transform` primitive) and the follow-up fix for map-mode's terminal aggregator (`drainMapModeAggregator`), both discovered during review.
- **§3** never stated whether `group_by.time_bucket` requires `agg`. It does, for a concrete reason found during implementation (the no-`agg` shape doesn't preserve the full output schema) — recorded as a current limitation, not a permanent decision.

No behavior change — the shipped code already reflects all three corrections. This is a documentation-only PR to keep the design doc accurate for whoever picks up apache/skywalking#14090 (COUNT_DISTINCT) and apache/skywalking#14091 (BydbQL) next.

## Test plan

- Docs-only change; no code affected.
- Verified the patch applies cleanly on top of `main` (post-#1362) and section numbering/headers remain consistent.


hanahmily:
Folding this into the #14090 implementation branch instead of shipping it standalone.

PR #1371, opened 2026-09-21
Add data export import design docs

mrproliu:


docs/design/0.12.0/data-export-import/README.md
```diff
@@ -0,0 +1,1381 @@
...
+
+Native import records progress per part, tracking successful and missing target nodes separately. Native indexes and property record the whole blob, with no part level. CSV records a cursor per split file, where `EOF` means the file is fully processed.
+
+| Time | Action | Persist |
+|---|---|---|
+| Native: before opening a stream | Set `inflight[worker] = {unit, kind, partID, nodes, isRetry}` | Write |
+| Native: after the response | Clear `inflight[worker]`; add successful nodes to `done[].nodes`, failed ones to `missing` | Write |
+| CSV: before a checkpoint range | Set `inflight[worker] = {unit, slice, fromRecord, records}` | Write |
+| CSV: after each successful batch in a range | Update the in-memory cursor only | No forced disk write |
+| CSV: when a checkpoint range completes | Clear `inflight[worker]`, write the cursor to `done[].slices[slice]` | Write |
```
hanahmily: [P1] Make import replay safe across checkpoint failures

If a receiver commits a trace part and the client crashes before persisting success (or loses the response), resume replays that part. The receiver allocates another local part ID and exposes duplicate spans. CSV also leaves successful batches uncheckpointed until the range completes. More frequent client checkpoints cannot eliminate the server-commit/client-checkpoint gap.

Required change: specify durable receiver-side operation identities and deduplication, or stop ambiguous retries for explicit reconciliation rather than claiming safe automatic resume. Cover both native and CSV crash/response-loss cases.

docs/design/0.12.0/data-export-import/README.md
```diff
@@ -0,0 +1,1381 @@
...
+
+- **trace is the only catalog whose query result is wrong**, and it does not self-heal. That is the whole reason §6.3 has to be accurate; for every other catalog a misjudgement is absorbed by the read or write path.
+- **The element index cannot be reduced to a covering set.** Each source's `idx/` may be complementary rather than duplicated, and a dropped index entry never reappears. It was measured at exactly ×3.000 for three sources (4,000 → 12,000 documents, 765 KB → 2,295 KB), and Bluge's merge does not reclaim duplicates because the element index does not enable deduplication. Extrapolated to a 167-segment group with a 200 MB per-shard baseline, that is 65 GB at `replicas=1` and 98 GB at `replicas=2`. Provision `idx/` at `(replicas + 1)` times.
+- **The segment-level series index is the opposite**: only the minimum covering set is pushed, because the receiver deduplicates and over-pushing costs no disk. But the winner is the first arrival, and the discarded copy's index tag values become unqueryable — `Lookup(entity)` still matches, because the Bluge document `_id` is the entity value itself, while a filter on a non-entity indexed tag returns nothing for the losing value.
```
hanahmily: [P1] Preserve series indexes from every selected data source

Shard coverage does not establish series-document coverage. After topology drift, two sources can both advertise shard 0 while containing different series. Both sources' disjoint-time core parts can qualify for import, but the minimum shard-covering index set can select only one source. The other source's unique series then have data without corresponding series-index entries and become undiscoverable.

Required change: include series indexes from every source whose data is selected, unless containment of the actual series documents—not just shard membership—is proven. Define the complementary-series case explicitly.

docs/design/0.12.0/data-export-import/README.md
```diff
@@ -0,0 +1,1381 @@
...
+| Type | Encoding |
+|---|---|
+| STRING | Written via `encoding/csv`; only values containing `\r\n` need extra encoding as `\B` + base64 |
+| INT | Decimal |
+| FLOAT (measure field) | `strconv.FormatFloat(v, 'f', -1, 64)`. Precision **must** be `-1`, or the round trip loses precision |
+| DATA_BINARY / span payload | `base64.StdEncoding`, standard alphabet, with padding |
+| STRING_ARRAY | JSON array string |
+| INT_ARRAY | Semicolon-separated decimal int64 values; an empty array is an empty string |
+| All timestamp columns | Decimal int64 UnixNano; trace-only for TIMESTAMP tags |
+| null | `\N`; a literal `\N` is escaped as `\\N` |
```
hanahmily: [P1] Define reversible CSV string escaping

Under these rules, a CRLF string and the literal string `\BDQo=` both encode as `\BDQo=`. Likewise, literal `\N` and literal `\\N` both encode as `\\N`. The importer cannot distinguish the original values, so the round trip corrupts one of them.

Required change: define escaping for literal escape prefixes and an explicit decoding order. Include round-trip cases for null, literal `\N`, literal `\\N`, literal `\B`-prefixed strings, and CRLF-containing strings.

docs/design/0.12.0/data-export-import/README.md
```diff
@@ -0,0 +1,1381 @@
...
+
+| # | bydbctl → liaison (import/v1) | liaison → data-* (cluster.v1) | Liaison behavior |
+|---|---|---|---|
+| ⓪ | Read the manifest and target schema/topology | — | — |
+| ① | Read the footer of the selected `.bnc` files and split into streams by part | — | — |
+| ② | `Init{catalog, group, shard_id, stage, kind, segment_version, parts{…}}` | — | Validate, resolve the target stage, call `LocateAll(group, shard, copies)`, open `SyncPart` streams |
+| ③ | `Chunk{chunk_index=0}` | `SyncPartRequest{chunk_index=0, metadata=SyncMetadata{…}, …}` | Build `SyncMetadata`, fill `session_id` and `version_info` |
+| ④ | `Chunk` frames | `SyncPartRequest` frames × N | Forward to all N target nodes |
+| ⑤ | `Completion` | `SyncPartRequest{completion=SyncCompletion{…}}` | Forward |
+| ⑥ | `ImportPartsResponse{requested_copies, targets[]}` | Final responses from all streams | Wait for all N results before returning |
```
hanahmily, line 1019: [P2] Specify how partial-success retries avoid successful replicas

The checkpoint records successful and missing targets separately, but ImportParts.Init has no target subset and every call fans out through LocateAll. If target A succeeds and B fails, retrying B also imports into A again. For trace this exposes duplicate spans even without a client crash.

Required change: add a validated retry-target subset to the protocol, or durable per-target completion handling that suppresses already-successful deliveries. Specify the A-success/B-failure retry flow.

docs/design/0.12.0/data-export-import/README.md
```diff
@@ -0,0 +1,1381 @@
...
+
+The liaison stores nothing: it converts one frame at a time and forwards it, which is why it stays memory-flat at line rate. The memory protector is not a backstop here — it only observes memory, and a relay saturating the NIC has a flat memory curve.
+
+### 5.3 CSV data rewrite
+
+For CSV, `bydbctl` reads the files directly, rebuilds the corresponding requests, and redistributes them to the target data nodes.
+
+```protobuf
+service ImportService {
+  rpc ImportRows(stream ImportRowsRequest) returns (ImportRowsResponse);
```
hanahmily: [P2] Align ImportRows responses with batch acknowledgments

ImportRows declares a client-streaming RPC with one response, but the workflow at line 1116 and checkpoint rules at lines 1375–1377 require individual successful-batch acknowledgments. The declared RPC cannot deliver those responses during a multi-batch stream.

Required change: specify bidirectional streaming with correlated batch acknowledgments, or explicitly use one RPC per batch and align the workflow/checkpoint rules with that choice.

hanahmily (changes requested):
Requesting changes to the design before implementation. The five inline comments identify required changes for crash-safe resume, series-index completeness, reversible CSV encoding, partial-replica retries, and CSV batch acknowledgments.

Reviewed c67b79ee against the main merge-base 89403f92. No actionable project-standards findings. Verification included focused code inspection, git diff --check, and reproduction of the CSV encoding collisions; no implementation tests were run.

hanahmily (approved):

